PR #5, opened 2020-02-04
VSCODE-12: Connection model and data service

Anemy:
https://jira.mongodb.org/browse/VSCODE-12

This PR introduces the backend wiring for connecting and disconnecting from mongoDB instances from the VSCode extension.

This PR also does some re-architecting: I moved the extension away from having the top level singleton to be more controller driven, so that hopefully actions flow down from a centralized place. The intention here is to isolate parts of the extension so in time while we introduce more complexity we can have tighter abstractions. I think it will in the long run it will help push us towards making related code live together, and avoid letting things get coupled/tangled. I also think it helps make testing easier as we can target behaviors in certain domains. I might be totally wrong here - I haven't written an extension like this so let's discuss here or in person. In the end it's what we like and what we want to work with. 
Main changes here are adding `mdbExtensionController.ts`, the entry point for commands, and `connectionManager.ts`, which controls our connections.

We're using `mongodb-connection-model` as the model for connections, and `mongodb-data-service` as our data service. We could move to just using the driver later, we're not too ingrained.

Connection configurations (not necessarily live connections) are stored in the `ConnectionManager` in a map called `_connectionConfigs` *Open q* Other possible names: `connectionContracts` or just `connections` or suggestions?

*Screenshots*
**Disconnecting with no current connection**
Screen Shot 2020-02-04 at 5 14 13 PM

**Connecting with uri with command pallete**
Screen Shot 2020-02-04 at 10 36 47 PM

**Disconnecting with multiple connections configured (only 1 can be active at a time)**
Screen Shot 2020-02-04 at 9 38 16 PM

**Status bar item - active when connecting & disconnecting**
Screen Shot 2020-02-04 at 9 39 26 PM


package.json
```diff
@@ -16,7 +16,9 @@
     "onCommand:mdb.connectWithURI",
     "onCommand:mdb.addConnection",
```
alenakhineika: Can we have several active connections at the same time? For me `addConnection`sounds like we are already connected to one server and want to add one more connection to the existing one.

package.json
```diff
@@ -16,7 +16,9 @@
...
     "onCommand:mdb.addConnectionWithURI",
+    "onCommand:mdb.disconnect",
     "onCommand:mdb.removeConnection",
+    "onCommand:mdb.launchShell",
```
alenakhineika: `launchShell` sounds like a shell will be opened as a separate window, won't it? Maybe `openShell` or `showShell` or `mongodbShell` would be more accurate?

src/connectionManager.ts
```diff
@@ -0,0 +1,264 @@
...
+
+    let connectionString;
+    try {
+      connectionString = await vscode.window.showInputBox({
+        value: '',
+        placeHolder: 'e.g. mongodb+srv://username:password@cluster0.mongodb.net/admin',
+        prompt: 'Enter your connection string (SRV or standard)',
+        validateInput: text => {
+          let connectionStringError = null;
+          if (text && text.indexOf('mongodb://') === -1 && text.indexOf('mongodb+srv://') === -1) {
```
alenakhineika: You can use `Connection.isURI(text)` for this check, And I would use `uri` or `connectionString` instead of `text` to make it more readable what is happening here.

src/connectionManager.ts
```diff
@@ -0,0 +1,264 @@
...
+  public getConnections() {
+    return this._connectionConfigs;
+  }
+  public getActiveConnectionInstanceId() {
+    return this._currentConnectionInstanceId;
+  }
+  public getActiveConnection() {
+    return this._currentConnectionInstanceId;
+  }
+  public getConnnecting() {
```
alenakhineika: Maby `isConnecting` or something would be better? I am not sure what to expect from `getConnnecting` :)

package.json
Anemy: Ah - I like `openShell`. good suggestion.

package.json
Anemy: Only 1 active connection at a time, but we can have a few saved connection configurations. I think `connect` and `connectWithURI` is probably best - good point. I'll update them to that.

Anemy:
> Is it any particular reason why you are thinking about using the driver directly instead of the connection model? The connection model is a middle layer between us and the driver. It validates connection options, prepares a valid driver URI and also has lots of tests, that would be nice to reuse instead of writing them again to test vscode connection module.

I think the driver already handles a lot of the things the connection model handles. I think it's fine to use the connection model, and it does help us, just in future it gives us another thing to maintain. 

>What do you mean by "Connections"? Do you mean you can be connected to several servers at the same time?

By "connections" I mean connection models or configurations. We can only have one active connection at a time, however we can have multiple configurations saved that the user can then choose another of and connect to it (which will disconnect from their current connection).

@alenakhineika just pushed some updates to the things you brought up - thanks

package.json
```diff
@@ -14,9 +14,9 @@
   "activationEvents": [
     "onCommand:mdb.connect",
     "onCommand:mdb.connectWithURI",
-    "onCommand:mdb.addConnection",
-    "onCommand:mdb.addConnectionWithURI",
+    "onCommand:mdb.disconnect",
     "onCommand:mdb.removeConnection",
+    "onCommand:mdb.openMongoShell",
```
alenakhineika: I think we don't use mongo without db anymore to not be accidentally offensive to someone. Maybe `openMongoDBShell ` would be better? Here we also call it `MongoDB Shell`: https://github.com/mongodb-js/mongosh

alenakhineika:
Oh and maybe you can add newly available commands to read.me? It would be easy to track what is new using read.me 🤓

PR #6, opened 2020-02-11
VSCODE-17: Explorer tree view and data provider

Anemy:
https://jira.mongodb.org/browse/VSCODE-17

This PR adds an explorer view to the extension. This is the left tree view of your database. Because you can only have one active connection at a time I opted to it so only the active connection can have a revealed dropdown (once a connection is disconnected the dropdown tree closes itself).

The tree consists of a few tree items:
- `ExplorerTreeController`
- `MDBConnectionsTreeItem` (The 'Connections' root)
- `ConnectionTreeItem`
- `DatabaseTreeItem`
- `CollectionTreeItem`
- `DocumentTreeItem`

When a `CollectionTreeItem` loads and more than the max documents are shown it shows a `Show more...` item. When the show more is clicked it fetches more documents.

Sorry, I think this one ended up being a bit bigger than I anticipated - I should have cut it down into a couple more tasks - I think I'm getting used to working again and not sure .

One thing this PR does not cover is when a connection goes inactive/when the user goes off the internet and back on. I think that currently breaks and we can hit it in a future ticket: https://jira.mongodb.org/browse/VSCODE-29
Another thing this PR does not cover is cache the expanded/collapsed state of nested tree items, so when a parent is collapsed their state is lost. Ticket created for that here: https://jira.mongodb.org/browse/VSCODE-30

**Screenshots**
*Empty tree view*
Screen Shot 2020-02-11 at 11 55 40 AM
*Connected tree view, expanded*
Screen Shot 2020-02-11 at 11 59 27 AM
*After clicking show more*
Screen Shot 2020-02-11 at 12 01 33 PM
*Connecting to a new connection*
Screen Shot 2020-02-11 at 12 02 08 PM


src/connectionController.ts
```diff
@@ -23,7 +24,11 @@ function getConnectWebviewContent() {
   </html>`;
 }
 
-export default class ConnectionManager {
+export enum DataServiceEventTypes {
+  connectionsDidChange
```
imlucas: SDAM events were talking about at lunch:
- https://github.com/mongodb/node-mongodb-native/blob/master/docs/guide/management/sdam-monitoring.txt
- which you can listen for on ds https://github.com/mongodb-js/data-service/blob/master/lib/native-client.js

src/connectionController.ts
```diff
@@ -151,13 +171,23 @@ export default class ConnectionManager {
         this._currentConnection = newConnection;
         this._connecting = false;
 
+        this.fireConnectionEvent(DataServiceEventTypes.connectionsDidChange);
```
imlucas: fireEvent ➡️ emit

src/connectionController.ts
```diff
@@ -23,7 +24,11 @@ function getConnectWebviewContent() {
   </html>`;
 }
 
-export default class ConnectionManager {
+export enum DataServiceEventTypes {
+  connectionsDidChange
+}
+
+export default class ConnectionController {
```
imlucas: If you extend from require(‘events’).EventEmitter, you’ll get all of the listener/emit boilerplate for free.

src/connectionController.ts
imlucas: https://github.com/mongodb/node-mongodb-native/blob/master/test/tools/sdam_viz

Anemy: nice - made a ticket for it: https://jira.mongodb.org/browse/VSCODE-29

src/connectionController.ts
Anemy: Good stuff - updated to use it.

PR #10, opened 2020-02-13
VSCODE-30: cache expanded tree item state

Anemy:
https://jira.mongodb.org/browse/VSCODE-30

This PR makes it so that the explorer tree view in the extension save their expanded state when a parent is expanded or collapsed.

*Probably going to wait to merge this one after eslint is merged so that we can just fix the new changes here.*

We also add an icon for collection views vs just collections. And documents in views are given a title `Document INDEX` where INDEX is just a number starting at 1.
Here that is in dark & light mode:
Screen Shot 2020-02-13 at 4 55 32 PM
Screen Shot 2020-02-13 at 4 55 48 PM

Cached tree expansion example:
gif


src/explorer/connectionTreeItem.ts
```diff
@@ -53,58 +60,65 @@ export default class ConnectionTreeItem extends vscode.TreeItem implements TreeI
...
-              return reject(err);
-            }
+    if (!this.isExpanded || this._connectionController.isDisconnecting() || this._connectionController.isConnnecting()) {
+      return Promise.resolve([]);
+    }
+
+    if (this._childrenCacheIsUpToDate) {
+      return Promise.resolve(Object.values(this._childrenCache));
+    } else {
+      return new Promise(async (resolve, reject) => {
```
alenakhineika: I have seen this `new Promise(async (resolve, reject))` syntax in master and eslint didn't like it :) so I switched off `no-misused-promises` rule.

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-misused-promises.md

But maybe we can avoid this syntax in the future and in this particular case you removed `await this._connectionController`, so you maybe don't need `async` keyword at all.

src/explorer/connectionTreeItem.ts
```diff
@@ -53,58 +60,65 @@ export default class ConnectionTreeItem extends vscode.TreeItem implements TreeI
...
   }
 
-  async onDidExpand() {
+  onDidExpand(): Promise<any> {
     this._childrenCacheIsUpToDate = false;
     this.isExpanded = true;
+
+    // If we aren't the active connection, we reconnect.
+    if (this._connectionController.getActiveConnectionInstanceId() !== this._connectionInstanceId) {
+      return new Promise((resolve, reject) => {
```
alenakhineika: You don't need `reject` here if you use only resolve both in case of success and in case of error.

src/explorer/databaseTreeItem.ts
```diff
@@ -32,42 +36,70 @@ export default class DatabaseTreeItem extends vscode.TreeItem implements TreeIte
...
-      }
+    if (!this.isExpanded) {
+      return Promise.resolve([]);
     }
 
-    return Promise.resolve([]);
+    if (this._childrenCacheIsUpToDate) {
+      return Promise.resolve(Object.values(this._childrenCache));
+    } else {
+      return new Promise((resolve, reject) => {
```
alenakhineika: You can get rid of `else` after having `return` above.

src/test/suite/explorer/databaseTreeItem.test.ts
```diff
@@ -26,4 +26,35 @@ suite('DatabaseTreeItem Test Suite', () => {
       assert(collections[1].label === mockDatabases[mockDatabaseNames[1]].collections[1].name, `Expected a tree item child with the label collection name ${mockDatabases[mockDatabaseNames[1]].collections[1].name} found ${collections[1].label}`);
     }).then(() => done(), done);
```
alenakhineika: You can also write it like this:

`.then(done).catch(done);`

It might be easier to read and notice that done should happen both in case of the success of test and failure.

Anemy:
@alenakhineika Rebased on master and cleaned up the styles a bit. Ready for another pass :)

src/test/suite/explorer/databaseTreeItem.test.ts
Anemy: Hmm I agree `.catch(done)` is a lot easier to read and declarative. eslint doesn't like it however since it's not a variable on `Thenable<void>`. Which is what a lot of the vscode interface functions use as a return type. I went and made the tests use `.then(done, done);` to be a bit cleaner. 🤔

src/explorer/collectionTreeItem.ts
```diff
@@ -1,12 +1,13 @@
...
+const path = require('path');
 
 import DocumentTreeItem from './documentTreeItem';
 import TreeItemParent from './treeItemParentInterface';
 
 // We fetch 1 more than this in order to see if there are more to fetch.
 // Each time `show more` is clicked, the collection item increases the amount
 // of documents shown by this amount.
-const defaultMaxDocumentsToShow = 10;
+export const defaultMaxDocumentsToShow = 10;
```
imlucas: nit:
```suggestion
export const MAX_DOCUMENTS_VISIBLE = 10;
```

src/explorer/collectionTreeItem.ts
```diff
@@ -24,103 +25,136 @@ class ShowMoreDocumentsTreeItem extends vscode.TreeItem {
...
+        namespace,
+        { /* No filter */ },
+        {
+          // We fetch 1 more than the max documents to show to see if
+          // there are more documents we aren't showing.
+          limit: 1 + this._maxDocumentsToShow
+        },
+        (err: Error, documents: []) => {
+          if (err) {
+            return reject(`Unable to list documents: ${err}`);
```
imlucas: Reject expects an `Error` instance and not just a string, no?

src/explorer/connectionTreeItem.ts
```diff
@@ -59,63 +66,85 @@ export default class ConnectionTreeItem extends vscode.TreeItem
...
+          err => {
+            this.isExpanded = false;
+            vscode.window.showErrorMessage(err);
+            resolve(false);
+          }
+        );
+      });
+    }
+
+    return Promise.resolve(true);
```
imlucas: nit: switch this to an early return by moving `return Promise.resolve(true);` to top

```javascript
if (this._connectionController.getActiveConnectionInstanceId() === this._connectionInstanceId) {
  return Promise.resolve(true);
}
```

src/explorer/collectionTreeItem.ts
Anemy: Hmmm it looks like it works with both. That being said I think it's supposed to be `Error` instances - interesting that strings do the expected behavior in vscode though. I'll swap out all of our rejecting with strings to errors.

src/test/suite/explorer/databaseTreeItem.test.ts
alenakhineika: Let's do `done, done` then :)

PR #12, opened 2020-02-18
VSCODE-31: escape dollars in comments

alenakhineika:
## Description

Some agg stages' comments contain `$` symbol that is being interpreted as variable name, but the variable is not known. The solution is to escape this symbol before building the stage body.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/scripts/get-snippets.ts
+++ b/src/scripts/get-snippets.ts
@@ -12,12 +12,16 @@ const snippetTemplate = (
   comment?: string
 ): { prefix: string; body: Array<string>; description: string } => {
   const body = snippet.split('\n');
+  const find = /[$]/;
+  const re = new RegExp(find, 'g');
 
   body[0] = `\\${prefix}: ${body[0]}`;
 
   return {
     prefix,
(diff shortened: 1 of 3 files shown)


src/scripts/get-snippets.ts
```diff
@@ -12,12 +12,16 @@ const snippetTemplate = (
...
+  const find = /[$]/;
+  const re = new RegExp(find, 'g');
 
   body[0] = `\\${prefix}: ${body[0]}`;
 
   return {
     prefix,
-    body: comment ? [...comment.split('\n'), ...body] : [...body],
+    body: comment
+      ? [...comment.replace(re, '\\$').split('\n'), ...body]
```
Anemy: I think a good amount is going on here, probably a good place for a comment.

alenakhineika: Good point! Added some comments.

Anemy: :shipit:

PR #14, opened 2020-02-20
VSCODE-34: Save connection models

Anemy:
https://jira.mongodb.org/browse/VSCODE-34

This PR makes it so that after a new connection is added, it can be saved into either the vscode workspace, globally in vscode, or just kept on the current session. There's a setting that can disable the prompt that asks for where to save the connection in settings. There's an additional setting which dictates when the prompt is disabled, where new connections should be saved (`Global`, `Workspace`, or `Session Only`).

Currently we're storing the entire connection's driverUrl in storage. In the future we'll want to not store the password (or maybe give the option to store it). 

Another thing we could do is expose these connections as settings in the extension settings (an array or object of connections that the user can even add to or configure there in tandem with command palette commands). I'm still thinking on if that's the move, I don't think it would require too much work, and may be something that could be useful to users - open to thoughts on that one. @mmarcon 

It might be nice to display a message on the bottom right reflecting the action that the user took when saving. When ya'll try it does it feel like it needs that?

Here's the prompt after a user adds a new connection (Cancelling with escape makes the connection just exist on the session.):
Screen Shot 2020-02-20 at 10 34 17 AM
I wanted to use an information message prompt with a few options for this, but their API is fairly restrictive with showing multiple input options, a `Cancel` option is *always* shown as the 2nd to the right option...
Screen Shot 2020-02-20 at 12 44 29 AM
So I think we're using the best UI we can for this.

Extension connection saving settings:
Screen Shot 2020-02-20 at 10 38 41 AM

Default saving location setting dropdown:
Screen Shot 2020-02-20 at 10 38 58 AM

### DIFF
--- a/src/test/suite/connectionController.test.ts
+++ b/src/test/suite/connectionController.test.ts
@@ -473,6 +473,60 @@ suite('Connection Controller Test Suite', () => {
     });
   });
 
+  test('A saved connection can be loaded and connected to', function (done) {
+    const testExtensionContext = new TestExtensionContext();
+    const testStorageController = new StorageController(testExtensionContext);
+
+    const testConnectionController = new ConnectionController(
+      new StatusView(),
+      testStorageController
+    );
+
+    testConnectionController.activate();
+
+    vscode.workspace.getConfiguration('mdb.connectionSaving').update(
+      'defaultConnectionSavingLocation',
+      DefaultSavingLocations.Workspace
+    ).then(() => {
+      testConnectionController
+        .addNewConnectionAndConnect(testDatabaseURI)
+        .then(() => {
+          const workspaceStoreConnections = testStorageController.get(
+            StorageVariables.WORKSPACE_CONNECTION_MODELS,
+            StorageScope.WORKSPACE
+          );
+          assert(
+            Object.keys(workspaceStoreConnections).length === 1,
+            `Expected workspace store connections to have 1 connection found ${Object.keys(workspaceStoreConnections).length}`
+          );
+
+          testConnectionController.disconnect().then(() => {
+            testConnectionController.clearConnectionConfigs();
+            assert(
+              testConnectionController.getConnectionInstanceIds().length === 0,
+              'Expected no connection configs.'
+            );
+
+            // Activate (which will load the past connection).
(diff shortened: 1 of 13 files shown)


mmarcon:
I like it!

I suggest to show it at stakeholders demo and maybe also in the design sync, people may have useful feedback.

Some thoughts/questions:
* I think the option of saving connections from the command palette and selecting where to save them still in the command palette is a pretty common UI pattern in VS Code, so I think that's the way to go. Another option, not necessarily as an alternative but maybe an additional option, could be an icon next to the connection in the tree to save it – like a favorite ⭐️ or something like that. Not sure if VS Code supports this out of the box. No need to do it now, we can think about this later as a usability improvement.
* Saving the connection string is a good start but, as you said, having the password in clear text isn't great. Can you get the connection model as Object with the password stripped out and stored that as stringified JSON? It's definitely possible to then instantiate the connection model from the object and internally that gets converted back to the string but I don't know if there's a way to get the Object and then we'd have to figure out the flow to ask for the password. I think what you have is a good start and we can keep thinking about other options in the back of our heads.

Anemy:
@mmarcon Yes we can store it with the password stripped, and maybe then prompt the user fo the password when they try and connect. Let's give it a think and follow up in another pr for that.

src/test/suite/connectionController.test.ts
```diff
@@ -5,20 +5,56 @@ import { before, after } from 'mocha';
...
+      true
+    );
+    // Don't save connections on default.
+    await vscode.workspace.getConfiguration('mdb.connectionSaving').update(
+      'defaultConnectionSavingLocation',
+      DefaultSavingLocations['Session Only']
+    );
+  });
+  after(async () => {
+    require('mongodb-runner/mocha/after');
```
imlucas: fyi: you may want to move these runner mocha helpers to `pretest` and `posttest` package.json script hooks. They will not function as expected as you have them currently implemented.

Anemy: ah.... hah. Totally had these in and hadn't looked into what they were up to yet. ty - removed them since we already have them in our `pretest` & `posttest`.

PR #15, opened 2020-02-23
VSCODE-28: View collection in editor

Anemy:
https://jira.mongodb.org/browse/VSCODE-34

This PR adds the right click option `View Documents` to collections in the tree view. It currently shows the first `10` documents resulting from a `find` operation on the collection in a `read-only` `.json` file. In future work we're looking into code lenses to provide an option to show more documents. That work is in here: https://jira.mongodb.org/browse/VSCODE-36 I think once I add something with code lenses I'll also add in a status bar item for while we're performing the request to find the documents.

Feb-23-2020 21-14-07

If there's an error, like internet is disconnected, an error message is displayed in the bottom right and a document like this is opened:
Screen Shot 2020-02-23 at 9 15 16 PM

### DIFF
--- a/src/editors/collectionDocumentsProvider.ts
+++ b/src/editors/collectionDocumentsProvider.ts
@@ -1,0 +1,42 @@
+import * as vscode from 'vscode';
+
+import ConnectionController from '../connectionController';
+
+export const NAMESPACE_URI_IDENTIFIER = 'namespace';
+export const VIEW_COLLECTION_SCHEME = 'VIEW_COLLECTION_SCHEME';
+export const DOCUMENTS_LIMIT = 10;
+
+export default class CollectionViewProvider implements vscode.TextDocumentContentProvider {
+  _connectionController: ConnectionController;
+
+  constructor(connectionController: ConnectionController) {
+    this._connectionController = connectionController;
+  }
+
+  onDidChangeEmitter = new vscode.EventEmitter<vscode.Uri>();
+  onDidChange = this.onDidChangeEmitter.event;
+
+  provideTextDocumentContent(uri: vscode.Uri): Promise<string> {
+    return new Promise(resolve => {
+      const dataservice = this._connectionController.getActiveConnection();
+
+      // Remove the `namespace=` attribute from the uri query.
+      const namespace = uri.query.slice(NAMESPACE_URI_IDENTIFIER.length + 1);
+      dataservice.find(
+        namespace,
+        {}, // No filter.
+        {
+          limit: DOCUMENTS_LIMIT
+        },
+        (err: Error, documents: []) => {
+          if (err) {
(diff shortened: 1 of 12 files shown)


src/editors/collectionDocumentsProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+        {
+          limit: DOCUMENTS_LIMIT
+        },
+        (err: Error, documents: []) => {
+          if (err) {
+            vscode.window.showErrorMessage(`Unable to list documents: ${err}`);
+            return resolve(`Unable to list documents: ${err}`);
+          }
+
+          return resolve(JSON.stringify(documents, null, 2));
```
imlucas:
```suggestion
          import { EJSON } from 'bson';
          return resolve(EJSON.stringify(documents, null, 2));
```

src/editors/editorsController.ts
```diff
@@ -0,0 +1,45 @@
...
+    vscode.workspace.registerTextDocumentContentProvider(
+      VIEW_COLLECTION_SCHEME, collectionViewProvider
+    );
+
+    log.info('activated.');
+  }
+
+  onViewCollectionDocuments(namespace): Thenable<void> {
+    const uriQuery = `?${NAMESPACE_URI_IDENTIFIER}=${namespace}`;
+    // We attach the current time to ensure a new editor window is opened on
```
imlucas: Have you thought about `cursor.id` vs. timestamp?

Anemy: Hmmm... I think for now since we're just showing 10 documents and closing the query using `timestamp` makes the most sense. 

Once we're opening and storing a cursor for pagination using code lenses (hopefully) then using an identifier for the cursor might make sense. I'm not sure we have an `id` on a cursor we can use though? https://mongodb.github.io/node-mongodb-native/3.5/api/Cursor.html#~toArrayResultCallback Maybe I'm not following.

We also could make the timestamp string a readable string, but I found it ended up as a longer than we want to show.

src/editors/collectionDocumentsProvider.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  onDidChangeEmitter = new vscode.EventEmitter<vscode.Uri>();
+  onDidChange = this.onDidChangeEmitter.event;
+
+  provideTextDocumentContent(uri: vscode.Uri): Promise<string> {
+    return new Promise(resolve => {
+      const dataservice = this._connectionController.getActiveConnection();
+
+      // Remove the `namespace=` attribute from the uri query.
+      const namespace = uri.query.slice(NAMESPACE_URI_IDENTIFIER.length + 1);
```
Anemy: I'm working on the `code lens` implementation currently and I've changed this uri query parsing to use `URLSearchParams`: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

PR #16, opened 2020-02-24
VSCODE-23 setup syntax highlighting

alenakhineika:
## Description

**Note:** Some tests are failing, not sure if this PR causes it. Needs to be investigated.

Since the MongoDB language is the JavaScript language with a BSON support, it is beneficial to reuse the existing Typescript grammar and inject it with the MongoDB keywords. In this case:

- We get the JavaScript/TypeScript highlighting for free
- We don't have to maintain the JavaScript/TypeScript syntax itself
- We can focus only on MongoDB keywords support

Screenshot 2020-02-24 at 21 40 03

## ts.tmLanguage

The [ts.tmLanguage](https://github.com/mongodb-js/vscode/blob/VSCODE-23-setup-syntax-highlighting/src/syntaxes/ts.tmLanguage) file contains unchanged TypeScript grammar. The ts.tmLanguage file contains unchanged TypeScript grammar. It can be replaced with a new version when the TypeScript syntax will be changed at some point.

The ts.tmLanguage file is derived from [TypeScriptReact.tmLanguage](https://github.com/Microsoft/TypeScript-TmLanguage/blob/master/TypeScriptReact.tmLanguage) by [yo code](https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#developing-a-new-grammar-extension) generator.

## mongodb-injection.json

The [mongodb-injection.json](https://github.com/mongodb-js/vscode/blob/VSCODE-23-setup-syntax-highlighting/src/syntaxes/mongodb-injection.json) file contains only MongoDB related syntax.

The mongodb-injection.json file is generated by the `npm run getsymbols` script that uses agg symbols defined in [mongodb-ace-mode](https://github.com/mongodb-js/ace-mode/blob/master/index.js#L63-L263) repo. Currently, the array of following symbols is being used:

- AGG_ACCUMULATORS
- AGG_CONVERTERS
- AGG_EXPRESSION_OPERATORS
- AGG_QUERY_OPERATORS
- AGG_STAGE_OPERATORS
- EXT_JSON

**Note:** It would be nice to extract this list into the separate repo and require it everywhere else to avoid maintaining these lists within several projects. We also do not include the `npm run getsymbols` script into the compile script, because we do not expect frequent changes of the TS grammar and can to it manually when needed.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions

- Do we really need to maintain our own MongoDB TextMate grammar? We need anyway to reuse or reimplement all these rules for TS or JS.
- If yes can we just copy rules from ts.tmLanguage? It is the same issue we had with BSON transpiler. We didn't want to write a grammar from scratch because it is error-prone.
- When we reuse all the rules from TS grammar, is it fair to call it MongoDB grammar then? We basically need only injections presented in this PR.
- There is a list of language extensions: https://github.com/Microsoft/vscode/tree/master/extensions/. Maybe we can create a MongoDB language extension that will use TS grammar and our MongoDB injections. I haven't looked in detail yet how these extensions are being used.

## Types of changes

- [ ] *Backport Needed*


src/syntaxes/language-configuration.json
```diff
@@ -0,0 +1,30 @@
...
+        ["[", "]"],
+        ["(", ")"]
+    ],
+    // symbols that are auto closed when typing
+    "autoClosingPairs": [
+        ["{", "}"],
+        ["[", "]"],
+        ["(", ")"],
+        ["\"", "\""],
+        ["'", "'"]
```
Anemy: I think we also want to add ```["`", "`"]``` from https://github.com/microsoft/vscode/blob/master/extensions/javascript/javascript-language-configuration.json and remove the comments so that it's valid json.

alenakhineika:
Summary from my side:

- `Cosmosdb`, `mongodb-runner` and vscose itself use ts grammar to highlight js language
- The current PR does the same to avoid maintaining a new grammar that should support the whole js syntax and just add a list on mongodb keywords to it.
- `Vscose` has a list of extensions with grammars: https://github.com/microsoft/vscode/tree/master/extensions. We can do the same and build mongodb grammar based on ts grammar. Here are two ways:
	1) Create a parser, something similar to: https://github.com/microsoft/vscode/tree/master/extensions/javascript/syntaxes, that will convert ts to mongodb. We can still have our mongodb injection. This conversion will allow us to support highlighting only for .mongodb files. Currently, we allow it both for .ts and .mongodb files
	2) Copy ts grammar and modify it manually. But the next time we want to have a new version of ts grammar we will need to do it again without a parser.
	The question is if we want to allow later different file extensions to support our mongodb highlighting does it worth an effort now changing it. This I why we decided to keep the simple implementation and rethink it later.

PR #17, opened 2020-02-26
VSCODE-36: Add code lenses for viewing more collection documents

Anemy:
https://jira.mongodb.org/browse/VSCODE-36

This PR adds a `codeLens` as an action to view more documents when looking at documents in a collection/view. This is an action a user can take after they've hit `view documents` in the tree view (right click a collection or view).

Some things I found: We have to re-fetch all of the documents every time the user hits show more. This is because we can't really do smart pagination without knowing for sure how the user is using `_id`. We can't want to assume `_id` is chronological and use `skip` and `limit` because `views` don't have `_id` and `_id` can be anything the user wants. Using cursors could offer a solution but will be trickier. Added a ticket we can revisit: https://jira.mongodb.org/browse/VSCODE-44

Another small change on the diff is I added all of our commands and listeners onto our subscriptions disposable to be sure when our extension is deactivated is properly cleans up.

Screen Shot 2020-02-26 at 12 47 52 PM

show more

What happens when the user hits `show more` when the connection is no longer active:
Screen Shot 2020-02-26 at 12 38 02 PM
 
Extension setting for amount of documents to fetch: **Updated variable name to be Default Limit**
Screen Shot 2020-02-26 at 12 29 25 PM

Just added light & dark theme icons for the tree items: 
- `connection` (both inactive and active) (Link and unlink from font awesome)
- `database` (Database from font awesome)
- `collection` (Asset from Fred)

Screen Shot 2020-02-26 at 5 36 27 PM

Possible improvements: We could use a status bar item while the query is being executed (at the bottom of vscode) right now there's little feedback if your connection takes ages or is offline. Added ticket:  https://jira.mongodb.org/browse/VSCODE-43 We also don't free up the past operations ever - I don't think I see this being too much of an issue since it's not a big memory operation and is freed after the session...

I'm also going to be looking into why some of the vscode `configuration` update requests are timing out on `linux` which is making some of our tests flaky - some of the newer commits might be me trying things out.


package.json
```diff
@@ -118,6 +123,11 @@
           "default": true,
           "description": "Show or hide the MongoDB view."
         },
+        "mdb.amountOfDocumentsToFetch": {
```
imlucas:
```suggestion
        "mdb.defaultLimit": {
```

Anemy:
Added a ticket for the status bar item: https://jira.mongodb.org/browse/VSCODE-43
And here's a ticket for adding a cleaner pagination for how we find more documents:  https://jira.mongodb.org/browse/VSCODE-44

Just added light & dark theme icons for the tree items: 
- `connection` (both inactive and active) (Link and unlink from font awesome)
- `database` (Database from font awesome)
- `collection` (Asset from Fred)

Screen Shot 2020-02-26 at 5 36 27 PM

Still looking into why linux tests are flaky - somehow the configuration is timing out...

Anemy:
I think I might have fixed the linux flaky test bug... My theory is that the extension wasn't fully activated when the before `each` was called on the first test sometimes on linux. Maybe startup time is something that is dependent on the linux build. I moved the initial extension activation to before the tests are run which hopefully is the fix. Also found a test that wasn't properly running and fixed 👍 

Going to run the tests one more time, if we merge and the test flakes again somewhere else I'll revisit and maybe try some more hacky things around ensuring that extension configurations are set in all linux settings. 🙃

Anemy:
Hmmm... Still not sure why linux times out sometimes on setting configuration. I moved where we activate the extension and set default settings which should make it not spend as much time setting defaults on initial load. Not sure it'll fix it, I'll try something a bit hackier if it's not gone.

Anemy:
Fixed it by avoiding settings some defaults and mocking storage.

Anemy:
now it's just the unrelated windows bug that runs for an hour 🙃

alenakhineika:
@Anemy i restarted all of them and windows is ok now :)

PR #18, opened 2020-02-26
VSCODE-39: replace typescript grammar with mongodb grammar

alenakhineika:
## Description

- Use the MongoDB grammar instead of the TypeScript-MongoDB-injected grammar.
- Download the derived `mongodb.tmLanguage.json` from [vscode-mongodb-language](https://github.com/mongodb-js/vscode-mongodb-language) during `npm run compile`.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


Anemy:
Do we want to download this every time we compile or can we just reference it from the package json?

alenakhineika:
@Anemy what do you mean by "reference it from the package json"? like, specify a remote url as the location of the grammar? I think we should keep a downloaded grammar inside the project. but maybe we can remove the script from `npm run compile` and run `npm run update-grammar` manually when needed.

Anemy:
@alenakhineika  Ah yeah good call having it in the project is the move. Do you think moving out of compile might be better? It might make it easier to possibly develop offline. I think the risk is that we forget about running it when changes have happened in the grammer.

alenakhineika:
@Anemy yeah this is why i added it to compile script :) but i thought you have concerns. i personally prefer downloading a new version during compilation. the file is small. The script does not delete the old file before downloading, just pipes new data into the current one. so if something wrong with the network we won't end up without the grammar.

Anemy:
Sounds good - thanks for clarifying

PR #26, opened 2020-03-04
VSCODE-43: Show status bar item while fetching documents & creating collections

Anemy:
https://jira.mongodb.org/browse/VSCODE-53

This PR adds showing a status bar item message when we're fetching collection documents (view documents and `show more` code lens) and when we're creating a collection (create database and create collection in the tree view). Easy way to test is to disable your internet when you run one of the commands. Hopefully this will feel like some decent feedback that the operation is running when things are taking some time.

I added context to the status view parameters so when the extension is disposed it cleans up a bit better. Going to read up on how vscode prefers disposal. ah just saw https://code.visualstudio.com/api/references/icons-in-labels might incorporate...

stataus bar

### DIFF
--- a/src/editors/collectionDocumentsProvider.ts
+++ b/src/editors/collectionDocumentsProvider.ts
@@ -4,6 +4,7 @@ import { EJSON } from 'bson';
 
 import CollectionDocumentsOperationsStore from './collectionDocumentsOperationsStore';
 import ConnectionController from '../connectionController';
+import { StatusView } from '../views';
 
 export const NAMESPACE_URI_IDENTIFIER = 'namespace';
 export const OPERATION_ID_URI_IDENTIFIER = 'operationId';
@@ -14,10 +15,16 @@ export const VIEW_COLLECTION_SCHEME = 'VIEW_COLLECTION_SCHEME';
 export default class CollectionViewProvider implements vscode.TextDocumentContentProvider {
   _connectionController: ConnectionController;
   _operationsStore: CollectionDocumentsOperationsStore;
+  _statusView: StatusView;
 
-  constructor(connectionController: ConnectionController, operationsStore: CollectionDocumentsOperationsStore) {
+  constructor(
+    connectionController: ConnectionController,
+    operationsStore: CollectionDocumentsOperationsStore,
+    statusView: StatusView
+  ) {
     this._connectionController = connectionController;
     this._operationsStore = operationsStore;
+    this._statusView = statusView;
   }
 
   onDidChangeEmitter = new vscode.EventEmitter<vscode.Uri>();
@@ -45,6 +52,8 @@ export default class CollectionViewProvider implements vscode.TextDocumentConten
         return reject(new Error(`Unable to list documents: no longer connected to ${connectionInstanceId}`));
       }
 
+      this._statusView.showMessage('Fetching documents...');
+
       const dataservice = this._connectionController.getActiveConnection();
       dataservice.find(
         namespace,
@@ -54,6 +63,7 @@ export default class CollectionViewProvider implements vscode.TextDocumentConten
         },
         (err: Error, documents: []) => {
           operation.isCurrentlyFetchingMoreDocuments = false;
+          this._statusView.hideMessage();
 
           if (err) {
             vscode.window.showErrorMessage(`Unable to list documents: ${err.message}`);
--- a/src/editors/editorsController.ts
+++ b/src/editors/editorsController.ts
@@ -10,6 +10,7 @@ import CollectionDocumentsProvider, {
 } from './collectionDocumentsProvider';
 import ConnectionController from '../connectionController';
 import { createLogger } from '../logging';
+import { StatusView } from '../views';
 
 const log = createLogger('editors controller');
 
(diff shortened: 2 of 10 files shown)


src/editors/collectionDocumentsProvider.ts
```diff
@@ -14,10 +15,16 @@ export const VIEW_COLLECTION_SCHEME = 'VIEW_COLLECTION_SCHEME';
 export default class CollectionViewProvider implements vscode.TextDocumentContentProvider {
   _connectionController: ConnectionController;
   _operationsStore: CollectionDocumentsOperationsStore;
+  _statusView: StatusView;
 
-  constructor(connectionController: ConnectionController, operationsStore: CollectionDocumentsOperationsStore) {
+  constructor(
```
alenakhineika: When do we use constructor() and when we use activate()? What the purpose of activation?

src/editors/editorsController.ts
```diff
@@ -27,7 +28,8 @@ export default class EditorsController {
     log.info('activating...');
```
alenakhineika: Where does logger write? into file? Or debug window?

src/test/suite/editors/collectionDocumentsProvider.test.ts
```diff
@@ -20,6 +22,10 @@ const mockDocumentsAsJsonString = `[
 ]`;
 
 suite('Collection Documents Provider Test Suite', () => {
+  afterEach(function () {
```
alenakhineika: We need to think about which tests ui we want to use. Mocha differentiates keywords for bdd and tdd approaches: https://mochajs.org/#bdd

According to our settings we use tdd:
https://github.com/mongodb-js/vscode/blob/VSCODE-43/Show-status-bar-item-while-running-actions/src/test/suite/index.ts#L17-L21

And in tests we use `suite()` what matches the concept. But if we want to use `afterEach()` or `before()` we start mixing everything. I just was writing tests for `mongosh` integration and stumbled upon the same problem :) I wanted to have `before(), after() and context()`. The `context()` is an even more interesting case. Tests runner complains that `context is not a function`, what is weird since it should be a callback.

src/test/suite/mdbExtensionController.test.ts
```diff
@@ -441,6 +441,64 @@ suite('MDBExtensionController Test Suite', () => {
...
+  test('mdb.addDatabase shows a status bar item while it is creating the collection then hide it', (done) => {
+    const stubShowMessage = sinon.fake();
+    const stubHideMessage = sinon.fake();
+
+    const testStatusViewObject = {
+      show: stubShowMessage,
+      hide: stubHideMessage,
+      text: ''
+    };
+    const mockSatusBarItem = sinon.fake.returns(testStatusViewObject);
```
alenakhineika: Really like how you mock and fake entities in tests! very smooth! ✨

src/editors/collectionDocumentsProvider.ts
Anemy: Ah - initially I was thinking we would create all the controllers before activation of the extension, but I think we just want to create everything in the `activate()` extension method. I'll update this in the PR: https://github.com/mongodb-js/vscode/pull/28. Just since the diff there is already big I want to avoid a large merge conflict with this PR if possible 😅

src/editors/editorsController.ts
Anemy: Currently these are going into an output channel called `mongodb`. In the extension workspace  the channel is a bit of a secret in the output panel with a dropdown:
Screen Shot 2020-03-08 at 8 06 34 PM

src/test/suite/editors/collectionDocumentsProvider.test.ts
Anemy: hmmm agreed... let's give it a think... we can definitely do some things to make the tests a bit more consistent and pattern driven.

src/editors/collectionDocumentsProvider.ts
alenakhineika: Yeah agreed, the separate PR for it will be better! Nobody likes merge conflicts 😀

Anemy: Hmm - working on it now and I think actually separating activation of the main extension controller from the constructor might have a benefit in testing. We can have our workspace extension which has the commands registered and then create new instances of the extension which we can individually test.
I have gone and renamed the `activate` method in `connectionController` to `loadConnections` and the `activate` method in the `explorerController` to `createTreeView` to be more descriptive about what actually happens. (In https://github.com/mongodb-js/vscode/pull/28 )

src/editors/editorsController.ts
alenakhineika: Ah here it is! Really hidden :)

src/editors/collectionDocumentsProvider.ts
alenakhineika: To be honest I like `activate` more...) We use activate method for compass plugins and it feels like a right naming for it. it is easier to understand what this method does if it has the same name for different controllers. and if is better for testing it is totally fine to have both activate and constructor. what do you think about continue using activate name?

PR #28, opened 2020-03-06
VSCODE-52: Add connect disconnect and drop database collection explorer actions

Anemy:
https://jira.mongodb.org/browse/VSCODE-43

This PR adds the ability to `Drop Database` on a Database tree item, `Drop Collection` on a Collection tree item, and adds `Connect` and `Disconnect` to Connection tree items. Connection tree items now have different actions based on if they are connected or not.

Dropping databases and collections both have text input prompts that require the user to type in the name of the database or collection they are trying to drop.

I also lumped into this pr a bit of cleaning up on the error messages, previously we were letting vscode create error messages when commands failed, now we display them ourselves so we can remove a bit of the extra text vscode would add into the messages. Sorry for the large diff.

I added tests as end to end command handling tests, which is adding a good amount of code into `mdbExtensionController.test.ts` I'm planning on refactoring this a bit or moving code into more descriptive files although the initial command handling does happen in `mdbExtensionController.ts`... maybe in this pr or just in a future smaller one for a smaller diff. Open to thoughts.

A consideration is we can add groups to the tree actions, which separate the actions by a line (try right clicking). I'm thinking about how we could best lay these out. Not doing that in this PR.

Also this pr shrinks the size of the leaf icon on the left a bit - maybe shrink it a bit more? @fredtruman 
*Before*
Screen Shot 2020-03-05 at 5 58 06 PM

*After*  Made it a little smaller than this screenshot shows
Screen Shot 2020-03-05 at 5 57 44 PM

**Screenshots**
___
*Disconnected connection tree item actions*
Screen Shot 2020-03-05 at 7 19 57 PM

*Connected connection tree item actions*
Screen Shot 2020-03-05 at 7 19 49 PM

*Dropping actions & prompt*
Screen Shot 2020-03-05 at 7 19 34 PM
Screen Shot 2020-03-05 at 7 19 20 PM
Screen Shot 2020-03-05 at 7 19 07 PM

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -314,8 +314,6 @@ export default class ConnectionController {
     });
   }
 
-  // TODO: Shrink size of the leaf on the left view.
-
   public removeConnectionConfig(connectionId: string): void {
     delete this._connectionConfigs[connectionId];
     this._storageController.removeConnection(connectionId);
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -57,7 +57,7 @@ export default class ConnectionController {
     this._storageController = storageController;
   }
 
-  activate(): void {
+  loadSavedConnections(): void {
     // Load saved connections from storage.
     const existingGlobalConnectionStrings = this._storageController.get(
       StorageVariables.GLOBAL_CONNECTION_STRINGS
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -250,7 +250,7 @@ export default class ConnectionController {
     });
   }
 
-  public async connectWithInstanceId(connectionId: string): Promise<any> {
+  public async connectWithInstanceId(connectionId: string): Promise<boolean> {
     if (this._connectionConfigs[connectionId]) {
       return this.connect(this._connectionConfigs[connectionId]);
     }
@@ -265,9 +265,15 @@ export default class ConnectionController {
     );
 
     if (this._disconnecting) {
-      // TODO: The desired UX here may be for the connection to be interrupted.
       return Promise.reject(
         new Error('Unable to disconnect: already disconnecting from an instance.')
+      );
+    }
+
+    if (this._connecting) {
+      // TODO: The desired UX here may be for the connection to be interrupted.
+      return Promise.reject(
+        new Error('Unable to disconnect: currently connecting to an instance.')
       );
     }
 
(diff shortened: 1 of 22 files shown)


src/connectionController.ts
```diff
@@ -313,20 +325,19 @@ export default class ConnectionController {
...
-      );
+      vscode.window.showErrorMessage('Unable to remove connection: currently disconnecting.');
+      return Promise.resolve(false);
     }
 
     if (!this._connectionConfigs[connectionId]) {
       // No active connection(s) to remove.
-      return Promise.reject(new Error('Connection does not exist.'));
+      vscode.window.showErrorMessage('Connection does not exist.');
+      return Promise.resolve(false);
```
Anemy: Here's an example of where instead of `reject`ing the promise and letting vscode handle the message displaying, we're doing it ourselves to make the messages a bit clearer.

alenakhineika: Do you have screens of how the old error looked like and how the new one looks now? I run this branch and can't notice the difference... :)

src/test/suite/explorer/explorerController.test.ts
```diff
@@ -204,7 +152,7 @@ suite('Explorer Controller Test Suite', function () {
           `Expected active connection to be 'localhost:27018' found ${instanceId}`
         );
 
-        // This will timeout in 1s, which is enough time for us to just check.
+        // This will timeout in 500ms, which is enough time for us to just check.
```
alenakhineika: From where we got this 500ms value?

alenakhineika:
Oh and I can see that mongodb leaf is not gigantic anymore! 😄👍

src/test/suite/explorer/explorerController.test.ts
Anemy: The driver url we try to connect on is `mongodb://shouldfail?connectTimeoutMS=1000&serverSelectionTimeoutMS=500`
Which will fail/timeout in 500ms.

src/connectionController.ts
Anemy: Hmmm interesting, I tried myself. It appears that the error messaging is bad only when using commands originating from the tree view. Here's an example of what those look like:
Screen Shot 2020-03-09 at 1 40 27 PM

Interestingly though, commands from the command prompt just display the error message. Since some of these methods are used in command handlers from both the command prompt and tree view I think we want to probably want to go with `vscode.window.showErrorMessage('Connection does not exist.')` and `resolve(false)`

src/test/suite/explorer/explorerController.test.ts
alenakhineika: Got ya!

PR #29, opened 2020-03-09
VSCODE-47 integration with mongosh

alenakhineika:
## Description

- Use [browser-runtime-electron](https://github.com/mongodb-js/mongosh/tree/master/packages/browser-runtime-electron) to evaluate code from the playground.
- Use vscode code lenses do display the active db in the playground.

The new ticket was created to update mongosh dependencies with fixed runtime commands:  https://jira.mongodb.org/browse/VSCODE-72

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


package.json
```diff
@@ -302,6 +317,7 @@
     ]
   },
   "devDependencies": {
+    "mongosh": "file:../mongosh-65",
```
alenakhineika: My local project for `mongosh`.

src/editors/playgroundController.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  runAllPlaygroundBlocks(): Promise<boolean> {
+    return new Promise(async (resolve, reject) => {
+      const activeEditor = vscode.window.activeTextEditor;
+      const codeToEvaluate = activeEditor?.document.getText() || '';
+      const outputChannel = vscode.window.createOutputChannel(
+        'Playground output'
+      );
+      const res = await this.evaluate(codeToEvaluate);
```
Anemy: Small thing, this is something I was adding a bit of in https://github.com/mongodb-js/vscode/pull/28
For error handling, I think we want to to only reject legitimate errors and instead handle things users might run into ourselves, so here we might want to catch the error and display it:
```javascript
let result;
try {
  result = await this.evaluate(codeToEvaluate);
} catch(err: Error) {
  vscode.window.showErrorMessage(`Unable to run playground: ${err.message}`);
  resolve(false);
}
```
And then when for testing we can test on a false `resolve` value instead of a reject. Mostly this is so we control the error messaging instead of vscode, making errors a bit more understandable for users.

src/editors/playgroundController.ts
```diff
@@ -0,0 +1,96 @@
...
+    );
+
+    if (!this._runtime) {
+      this._runtime = new ElectronRuntime(serviceProvider);
+    }
+
+    const res = await this._runtime.evaluate(codeToEvaluate);
+    const value = formatOutput(res);
+
+    return value;
```
Anemy: Small thing - should this be `return Promise.resolve(value);` or does this work? Maybe we can change the return type of the function to `Promise<string>`

src/utils/formatOutput.ts
```diff
@@ -0,0 +1,89 @@
+import { inspect as utilInspect } from 'util';
+
+type EvaluationResult = {
+  value: any;
+  type?: string;
+};
+
+const inspect = (output): any => {
+  return utilInspect(output, {
```
Anemy: Ah didn't know about node js's inspect. cool.

src/connectionController.ts
```diff
@@ -165,26 +182,29 @@ export default class ConnectionController {
             return reject(new Error(`Failed to connect: ${error.message}`));
           }
 
-          const {
-            driverUrl,
-            instanceId
-          } = newConnectionConfig.getAttributes({
+          const { driverUrl, instanceId } = newConnectionConfig.getAttributes({
```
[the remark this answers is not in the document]
alenakhineika: I actually wanted to write about it here in the description, that I don't do it manually :) and sorry that real changes are being lost in automatic code prettifying. Lucas added settings file to the project and now vscode does it automatically. It reads settings from ` .vscode/settings.json`. I am surprised that it does not happen to you since the `settings.json` file is in git and you also use vscode for development, right?

src/editors/playgroundController.ts
alenakhineika: Ha! Yes, it should! But it works like this anyway :) Probably async/await is smart enough to resolve literals even tho they are not promise like. But I will update it to promise to be more explicit 👍

src/mdbExtensionController.ts
```diff
@@ -300,5 +305,6 @@ export default class MDBExtensionController implements vscode.Disposable {
     // TODO: Cancel active queries/playgrounds.
     this._connectionController.disconnect();
     this._explorerController.deactivate();
+    this._explorerController.deactivate();
```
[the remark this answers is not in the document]
alenakhineika: Yeah, one should be `this._playgroundController.deactivate();`

src/connectionController.ts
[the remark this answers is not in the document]
Anemy: hmmm yup. I'll look into it ty

alenakhineika:
@Anemy as I understood @mmarcon is working on publishing `browser-repl`, it should happen shortly. @mcasimir tried some other ways of temporary usage of mongosh, but said that only the option I use here worked.

Yeah VSCODE-46 is kind of the same thing that is happening here, but maybe we should think about acceptance criteria as a part of this other ticket. There are several issues I can see already:

- mongosh does not provide all API yet, so we also can't run ALL yet
- I would think more about formatting results. cli and browser version of mongosh each uses their own formatter of results. None of each actually worked completely for vscode, therefore I also created a separate formatter. But when new API will be provided, probably new format methods should be added. Maybe it would be beneficial to somehow unify and reuse it across all projects.
- According to formatting I also noticed that `show dbs` for example returns expected output, as the old shell does. But `help` for example returns some cursor, that is not being iterated. Or I need to do more work on it, or API is not finished yet.
- We need to think how to handle situations when a user has several lines of code like this:

```
show dbs
use test_db
```

The current evaluator does not understand it. It expects only one command at a time.

So my point is, we can consider this PR as integration with mongosh and should discuss all these issues and handle them as separate tickets/prs.

Anemy:
@alenakhineika sounds good. I like the plan for `VSCODE-46`. Let's let this sit for now then till we have the published dependency we need. lgtm.

alenakhineika:
@imlucas, @mcasimir so `npm i` work fine for the first time. but then when i do `rm -rf node_modules` and `npm i` again it throws:

```
lerna success exec Executed command in 14 packages: "npm install --no-package-lock"
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/mongosh-service-provider-server - Not found
npm ERR! 404
npm ERR! 404  'mongosh-service-provider-server@0.0.0' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/alenakhineika/.npm/_logs/2020-03-10T14_19_09_215Z-debug.log
```

It happens both for the single package as:
`"mongosh": "file:../.tmp-mongosh"`

and for:

```
"mongosh-service-provider-server": "file:../.tmp-mongosh/packages/service-provider-server"
"mongosh-browser-runtime-compass": "file:../.tmp-mongosh/packages/browser-runtime-compass"
```

imlucas:
@mcasimir what's blocking these modules from actually being published?

mmarcon:
@imlucas @alenakhineika @mcasimir I am not against publishing these modules, we don't need to wait for the PSC decision next week. We'll publish them again once the license has been finalized.

src/editors/activeDBCodeLensProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+      new vscode.Position(0, 0),
+      new vscode.Position(0, 0),
+    ))];
+
+    return this._codeLenses;
+  }
+
+  public resolveCodeLens?(codeLens: vscode.CodeLens): vscode.CodeLens {
+    if (this._runtime) {
+      this._activeDB = this._runtime.openContextRuntime.serviceProvider.nodeTransport.mongoClient.s.options.dbName;
```
Anemy: Hmmm where is `this. _runtime` set?

src/editors/activeDBCodeLensProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    codeLens.command = {
+      title: `Active db is ${this._activeDB}`,
+      tooltip: "Tooltip provided by Active DB CodeLens Provider",
+      command: "mdb.runDBHelpInPlayground",
+      arguments: ["Argument 1", { connectionInstanceId: '111111' }]
```
Anemy: I think we can get rid of the arguments here since we aren't using them them. When I hover I can't see the tooltip - is there a way to show a tooltip? Maybe it could read `Click to show information on db functions.`. Trying out clicking on this I think the behavior might not be totally intuitive for the user - it might hide something they were looking at in the output window - I like the idea of giving the user some help though.

src/editors/playgroundController.ts
```diff
@@ -0,0 +1,125 @@
...
+        }, reject);
+    });
+  }
+
+  runDBHelpInPlayground(): Promise<boolean> {
+    return new Promise(async (resolve, reject) => {
+      let result;
+
+      try {
+        result = await this.evaluate('db.help()');
```
Anemy: When I run this I get the output:
Screen Shot 2020-03-19 at 4 12 28 PM
I think we probably want to show output similar to how shell shows it:
Screen Shot 2020-03-19 at 4 15 20 PM
Is this something that shell handles on its own end when displaying output?

src/editors/playgroundController.ts
```diff
@@ -0,0 +1,125 @@
...
+      'Playground output'
+    );
+  }
+
+  createPlayground(): Promise<boolean> {
+    return new Promise((resolve, reject) => {
+      vscode.workspace
+        .openTextDocument({
+          language: 'mongodb',
+          content: '// The MongoDB playground'
```
Anemy: I wish we could also provide a starting name for the text document here so that users know to save it as a `.mongodb` extension, but I don't think we can... I think we've talked about this with Max - later on it'll be nice to have a good template for the user to start with so they have an understanding of the capability of playgrounds. Seeing if we have a ticket for that

.npmignore
```diff
@@ -0,0 +1 @@
+/mongosh
```
Anemy: Now that we have it as a dependency, I think we can remove this?

src/editors/activeDBCodeLensProvider.ts
alenakhineika: Oh good catch! it should be `this._activeDB` instead.

src/editors/activeDBCodeLensProvider.ts
alenakhineika: I was playing with configuration and didn't delete it. Yeah, we don't need it here at all. This PR is open for toooo long :)

src/editors/playgroundController.ts
alenakhineika: I think this API is not finished in mongosh yet. You can try to evaluate in the playground the regular `help`, it also does not return what shell does.

src/editors/activeDBCodeLensProvider.ts
alenakhineika: I will check how to do a tooltip.

src/editors/playgroundController.ts
alenakhineika: For new documents, VSCode shows an `untitled` filename without an extension. Then as you type it replaces it with the first line form the active document. And only after saving it adds an extension to the custom title. There is a way how to hack the title, but as I read the way how it works was intentionally implemented by VSCode. And the extension is being added only after saving anyway. This is actually the reason why I check `editorLangId` here and not `resourceLangId`.

https://github.com/mongodb-js/vscode/blob/VSCODE-47-integration-with-mongosh/package.json#L377

.npmignore
alenakhineika: Yes, thank you very much for noticing all the middle/test/debug code here! 🍬 I was experimenting so many times and I knew it will be a bit of a mess in this PR..)

alenakhineika:
@Anemy yeah as I mentioned it here in the description some commands currently do not work in the runtime. The https://github.com/mongodb-js/mongosh/pull/64 was merged already but wasn't published to npm yet. Therefore I have created this ticket to update it when it will be done: https://jira.mongodb.org/browse/VSCODE-72

src/editors/activeDBCodeLensProvider.ts
alenakhineika: I didn't find a way how to add a tooltip to code lens. It seems that the text we display there is just a string and codeLens can't be extensible at the moment. And I use this `db.help()` command because the command is required prop. Without a command, we can't specify a text.

PR #30, opened 2020-03-10
VSCODE-57: Collection schema tree view

Anemy:
https://jira.mongodb.org/browse/VSCODE-57

This PR adds viewing a collection or view's schema in the tree view. Using https://github.com/mongodb-js/mongodb-schema we show all of the fields found in the sample `MAX_DOCUMENTS_VISIBLE` (10) documents pulled from the collection. When we find arrays or objects which have a type probability of 1 (they're not polymorphic in the sample) we show a dropdown for these items. For arrays, we show a dropdown which contains an `Array` dropdown which has items for all of the types found in the array.

Currently we're just showing the name of the fields, we'll be adding icons and maybe some better tooltips in https://jira.mongodb.org/browse/VSCODE-58 

We currently show the first 15 fields found in a collection, and hide the rest behind a `Show more fields...` selection item. There's a gif at the bottom on this. Do ya'll think 15 is a good number?
We aren't caching the expanded state of the schema yet, so when it's hidden then re-opened it loses its nested expansion state. I've added this to a future ticket https://jira.mongodb.org/browse/VSCODE-59

In this PR we also add types for the `Connection` model we are using in connection manager and `DataService`, I haven't yet piped them around fully but it's a start in getting rid of the `any`.

For testing I added `dbTestHelper` which has a few functions that should help with mocking and clearing up data a bit so we can fully test against a real database.

**Screenshots**
___
*Show more fields...*
show more fields

*Showing an array of documents and string array.*
array of docs and more

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -1,6 +1,6 @@
 import * as vscode from 'vscode';
-import * as Connection from 'mongodb-connection-model/lib/model';
-import * as DataService from 'mongodb-data-service';
+import Connection = require('mongodb-connection-model/lib/model');
+import DataService = require('mongodb-data-service');
 
 const { name, version } = require('../package.json');
 
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -1,6 +1,8 @@
 import path = require('path');
 import * as vscode from 'vscode';
 
+import { ConnectionConfigType } from './connectionConfig';
+import { DataServiceType } from './dataServiceType';
 import { createLogger } from './logging';
 import { StatusView } from './views';
 import { EventEmitter } from 'events';
@@ -35,11 +37,11 @@ export enum DataServiceEventTypes {
 export default class ConnectionController {
   // This is a map of connection instance ids to their connection model.
   private _connectionConfigs: {
-    [key: string]: any;
+    [key: string]: ConnectionConfigType;
   } = {};
 
-  _currentConnection: any;
-  private _currentConnectionConfig: any;
+  _currentConnection: DataServiceType | null = null;
+  private _currentConnectionConfig: ConnectionConfigType | null = null;
   private _currentConnectionInstanceId: string | null = null;
 
   private _connecting = false;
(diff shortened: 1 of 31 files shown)


Anemy:
Hmmm seeing a warning in tests
Screen Shot 2020-03-10 at 7 46 34 AM

src/connectionController.ts
```diff
@@ -35,11 +37,11 @@ export enum DataServiceEventTypes {
 export default class ConnectionController {
   // This is a map of connection instance ids to their connection model.
   private _connectionConfigs: {
-    [key: string]: any;
+    [key: string]: ConnectionConfigType;
   } = {};
 
-  _currentConnection: any;
-  private _currentConnectionConfig: any;
+  _currentConnection: DataServiceType | null = null;
```
alenakhineika: What does `null = null` mean?

src/test/suite/snippets/stageAutocompleter.test.ts
```diff
@@ -41,8 +41,6 @@ const STAGE_LABELS = [
 ];
 
 suite('Stage Autocompleter Test Suite', () => {
-  vscode.window.showInformationMessage('Starting tests...');
```
alenakhineika: Hehe I was about to ask you if we really want to have these messages but you already deleted them 👍

src/test/suite/mdbExtensionController.test.ts
```diff
@@ -7,15 +7,15 @@ import { CollectionTreeItem } from '../../explorer';
...
+import { CollectionTypes } from '../../explorer/documentListTreeItem';
 import { mdbTestExtension } from './stubbableMdbExtension';
 import ConnectionController from '../../connectionController';
 import { StorageController } from '../../storage';
 
 const testDatabaseURI = 'mongodb://localhost:27018';
 
 suite('MDBExtensionController Test Suite', () => {
-  afterEach(function() {
+  afterEach(function () {
```
alenakhineika: Maybe use an arrow function? we don't care about context here i believe.

alenakhineika: Do we actually want to get rid of these old-fashioned function keywords everywhere? ;)

src/connectionController.ts
Anemy: ah just the type can be `DataServiceType` or `null` and then we're setting it to `null`. hah it does look a little funny though... I'll swap the types around just for readability

src/test/suite/mdbExtensionController.test.ts
Anemy: good call

src/connectionController.ts
alenakhineika: ahah i was reading it wrong! thx for explaining :)

PR #31, opened 2020-03-10
fix(deps): Add debug to fix fatal when installed from .vsix 

imlucas:
### fix(deps): Add debug to fix fatal when installed from .vsix 

Compass modules assume `debug` is a peerDependency so this will make `.vsix` work end-to-end. 9ca8823

### feat(logging): Redirect mongodb-js modules using debug to extension logs

Since we added debug to fix something in the previous commit, we can also get more logging out of data-service, connection-model and friends by [redirecting `debug`'s output stream](https://github.com/visionmedia/debug#output-streams) to the extensions log output channel.

### docs(playground): Add real example playground

Some data wrangling of plant sensor data is used to confirm VSCODE-63 fixed. Seemed handy to just add so we can use for testing/Q&A etc. The underlying data is also up on Compass Data Sets Shared Atlas.

Screenshot 2020-03-10 13 10 29

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


imlucas:
```
☉ [add-debug] vsc/ code --uninstall-extension mongodb.mongodb-vscode
Uninstalling mongodb.mongodb-vscode...
Extension 'mongodb.mongodb-vscode' was successfully uninstalled!
```

PR #32, opened 2020-03-11
VSCODE-59: Add schema actions and cache

Anemy:
https://jira.mongodb.org/browse/VSCODE-59

This PR adds caching to the state of the tree view for schema. So that nested elements' collapsed states are preserved when a top level tree item is collapsed & expanded.

This PR also adds groups to the tree view to make the actions a bit more navigable and user friendly. And we have a refresh action on schema now, which reloads the schema and reset dropdown states. Ya'll think we could improve these groupings a bit? Open to suggestions.

On refresh we fetch all of the fields if the user already clicked the `show more fields` button.

**Screenshots**
___
*Caching of tree schema*
cache

*Refresh action on schema*
right click refresh action

*Groups in tree view actions (those lines that separate the actions are the groups)*
Screen Shot 2020-03-11 at 3 36 30 PM
Screen Shot 2020-03-11 at 3 37 40 PM
Screen Shot 2020-03-11 at 3 36 21 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,11 +56,15 @@
   "activationEvents": [
     "onCommand:mdb.connect",
     "onCommand:mdb.connectWithURI",
+    "onCommand:mdb.createPlayground",
+    "onCommand:mdb.addConnection",
+    "onCommand:mdb.addConnectionWithURI",
     "onCommand:mdb.disconnect",
     "onCommand:mdb.removeConnection",
     "onCommand:mdb.openMongoDBShell",
     "onView:mongoDB",
-    "onLanguage:json"
+    "onLanguage:json",
+    "onLanguage:mongodb"
   ],
   "contributes": {
     "viewsContainers": {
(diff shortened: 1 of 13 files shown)


alenakhineika:
Looks great!

I am thinking about caching and I have noticed that if connections are collapsed in the sidebar and we connect via URI, we don't see an active connection in the sidebar. I was confused for a second and thought that something went wrong and I wasn't connected. Maybe we always should expand a current connection in the sidebar. What do you think?

Mar-12-2020 12-07-35

alenakhineika:
I also have noticed that data service uses the extended model and therefore we can see `Could not load electron Cannot find module 'electron'` errors in debug console. Probably we should extend connection in `compass-connect` plugin directly instead of doing it in the connection model. In this case, the connection model will be free from electron dependency 🤔

Screenshot 2020-03-12 at 13 02 32

Anemy:
@alenakhineika good find - I'll start expanding the `connections` when a new connection is added in - I'll do it in the current branch I'm working on for https://jira.mongodb.org/browse/VSCODE-66
Ah yup on `compass-connect`... 🤔

PR #34, opened 2020-03-12
VSCODE-66: Add connection renaming

Anemy:
https://jira.mongodb.org/browse/VSCODE-66

This PR adds the ability to rename connections. It also improves our connection controller's saved connection implementation so we don't require a dns fetch to add a connection and adding two connections with the same driver url aren't blocked anymore. It also makes it so when a user copies a connection string from a connection in the connection tree view we no longer attach the `appname` parameter to the uri. 

@alenakhineika 's suggestion we also now expand the `connections` tree element if it's collapsed when a connection is added.

When a connection is first added, its name is its derived instance id.

Here's the new type for connections in the connection controller:
```typescript
export type SavedConnection = {
  id: string; // uuidv4
  name: string; // Possibly user given name, not unique.
  driverUrl: string;
  storageLocation: StorageScope;
};
```

**Screenshots**
___
renaming a connection


src/explorer/connectionTreeItem.ts
```diff
@@ -98,7 +99,10 @@ export default class ConnectionTreeItem extends vscode.TreeItem
     }
 
     return new Promise((resolve, reject) => {
-      const dataService = this._connectionController.getActiveConnection();
+      const dataService = this._connectionController.getActiveDataService();
```
alenakhineika: So there is no more `getActiveConnection` only `getActiveDataService`? How can I get the active connection now to read all connection properties from it?

src/explorer/mdbConnectionsTreeItem.ts
```diff
@@ -69,48 +77,58 @@ export default class MDBConnectionsTreeItem extends vscode.TreeItem
...
 
     return Promise.resolve(Object.values(this._connectionTreeItems));
   }
 
   connectionsDidChange(): void {
-    this._childrenCacheIsUpToDate = false;
+    // When the connections change, like a connection is added or removed,
+    // we want to open the connections dropdown if it's collapsed.
+    if (!this.isExpanded) {
+      this.needsToRefreshExpansionState = true;
```
alenakhineika: Thx for adding it! ✨

src/storage/storageController.ts
```diff
@@ -18,6 +19,13 @@ export enum DefaultSavingLocations {
   'Session Only' = 'Session Only'
 }
 
+export type SavedConnection = {
+  id: string; // uuidv4
```
alenakhineika: 👍

alenakhineika:
Like how you did renaming! And I believe this is very nice to be able to save connections with the same URI. It is better to delete redundant connections later than throw error messages which might be confusing for users. Great job! 🚀

src/explorer/connectionTreeItem.ts
Anemy: Which connection properties are you looking for? Maybe I need to keep storing the connection model as well (this pr currently removes it and just keeps the basic connection and data service). We're storing the driver url on the `SavedConnection` now.

Anemy:
Ah one more thing I noticed I should add before we merge this pr is we want to show the prompt for where to save the connection right after the user inputs the connection string, right now it prompts after the connection is established.

src/explorer/connectionTreeItem.ts
alenakhineika: 1) For example, I am connected, and I want to know which host and port I am currently connected to. Or which database is the default for the current connection. How can I do it?

2) I want to check if I am currently connected or not. Is it the right way to do it?

```
const dataservice = this._connectionController.getActiveDataService();
if (dataservice === null) {
  vscode.window.showErrorMessage(
    `Unable to list documents: no longer connected to ${connectionId}`
  );
}
```

In Compass we use DataService more like API to the driver, when ConnectionModel is more about connection properties. If the getActiveDataService returns the connection model, then maybe it would be more descriptive to call it as before getActiveConnection.

Or maybe I just got something wrong, pls help to understand :)

Anemy: Ah k - I'll add a call for `getActiveConnectionModel` for connection details. Correct on checking data service, I'll add a helper for that in connection controller so we can just get a boolean. Dataservice is like the api for the driver yup

Anemy:
@alenakhineika Added two functions to `ConnectionController` that'll probably be helpful - `getActiveConnectionModel` and `isCurrentlyConnected`. Also updated adding connections to prompt right when the connection is added instead of after the connection is established.

I've been noticing the tests sometimes crash on mac with this error:
Screen Shot 2020-03-16 at 1 26 12 PM
Looking into it... not sure if it's from this pr or if we should wait to merge these changes on fixing it.

PR #36, opened 2020-03-18
VSCODE-27: View document in editor

Anemy:
https://jira.mongodb.org/browse/VSCODE-27

This PR adds the functionality of opening a document in the editor when the document is clicked in the tree view. While a document is being fetched, a status bar message is displayed: `Fetching document...`.
For views we no longer show a dropdown and instead when the `documents` item is selected we open the documents in the editor.
Also tweaked the action group for renaming a connection so that it's in the 2nd group in the dropdown.

~I added a ` --no-force-async-hooks-checks` flag to the vscode node `runTests` command, which I think might workaround the renderer process crashing.~ The issue might be having to do with async hooks not playing nicely in electron, I'll be taking a look at that in the `data-service` package to see if we can fix and remove the flag here.  https://jira.mongodb.org/browse/COMPASS-4087

**Screenshots**
___
*View a document by selecting it*
view doc

*View documents in a view*
view documents in view

*New connection action grouping*
Screen Shot 2020-03-18 at 2 08 05 AM


azure-pipelines.yml
```diff
@@ -1,6 +1,7 @@
 variables:
   npm_config_cache: $(Pipeline.Workspace)/.npm
   node_version: '12.4.0'
+  MONGODB_VERSION: '4.2.3'
```
Anemy: This is a temporary fix for missing tarball in CDN.

For now we run tests against mongo 4.2.3 since 4.2.4 isn't available yet. We'll want to remove this later.

src/connectionController.ts
```diff
@@ -25,9 +25,8 @@ function getConnectWebviewContent(): string {
         <title>Connect to MongoDB</title>
     </head>
     <body>
-      <img src="https://media.giphy.com/media/JIX9t2j0ZTN9S/giphy.gif" width="300" />
       <h1>MongoDB Connection Details Form</h1>
-      <h2>Also try connecting with URI.</h2>
+      <h4>Coming soon :)</h4>
```
alenakhineika: I actually would also add something like: `In order to connect please use a contextual menu or a command palette` to address the issue that Chuck had when he didn't know how to connect when a form is not ready yet.

src/editors/collectionDocumentsProvider.ts
```diff
@@ -99,7 +98,7 @@ export default class CollectionViewProvider implements vscode.TextDocumentConten
             operation.hasMoreDocumentsToShow = false;
           }
 
-          return resolve(EJSON.stringify(documents, null, 2));
+          return resolve(JSON.stringify(documents, null, 2));
```
alenakhineika: Why not EJSON?

src/editors/documentProvider.ts
```diff
@@ -0,0 +1,99 @@
...
+      const dataservice = this._connectionController.getActiveDataService();
+      if (dataservice === null) {
+        vscode.window.showErrorMessage(
+          `Unable to find document: no longer connected to ${connectionId}`
+        );
+        return reject(
+          new Error(
+            `Unable to find document: no longer connected to ${connectionId}`
+          )
+        );
```
alenakhineika: Why do we need both `vscode.window.showErrorMessage` and `return reject(new Error( ))`?

src/explorer/explorerTreeController.ts
```diff
@@ -82,11 +84,29 @@ export default class ExplorerTreeController implements vscode.TreeDataProvider<v
 
     treeView.onDidChangeSelection((event: any) => {
       if (event.selection && event.selection.length === 1) {
-        if (event.selection[0].isShowMoreItem) {
-          event.selection[0].onShowMoreClicked();
+        const selectedItem = event.selection[0];
+        if (selectedItem.isShowMoreItem) {
```
alenakhineika: Can we have `a rule` when we use empty lines? :) eg empty line after variables declaration, or around if/else statements. It is not necessary at all! Just some thoughts of a perfectionist :))

src/editors/collectionDocumentsProvider.ts
Anemy: I think this is something we might want to revisit when we work on editing documents. Chatted with @mmarcon and we were thinking JSON looks a little more intuitive to users.
Having objectId written as `5e32b4d67bf47f4525f2f8ab` vs
```json
{
  "$oid": "5e32b4d67bf47f4525f2f8ab"
}
```
for example.

src/connectionController.ts
Anemy: Ah good call, done.

src/editors/documentProvider.ts
Anemy: The show error message shows what occurred to the user, while the reject just has a meaningful message which is currently silent/unused.

src/explorer/explorerTreeController.ts
Anemy: Sure, so after variable statements and before `if` statements it would be preferable that there's an empty line? Fixed here - I think I sometimes don't leave an empty space when the variable we're setting is exclusively used in the if statements. I think empty space looks better now that it's added tho.

src/editors/collectionDocumentsProvider.ts
Anemy: I think best case might be to write output like `_id: ObjectId('5e32b4d67bf47f4525f2f8ab')` but we'll need to not display results in a json document then, and instead have a different output format...

alenakhineika: Will be `JSON.stringify()` able to handle any BSON document without throwing errors?

src/editors/documentProvider.ts
alenakhineika: Maybe we can define the text of the error once and then pass it to both `vscode.window.showErrorMessage` and `return reject(new Error( ))` instead of repeating it?

src/editors/collectionDocumentsProvider.ts
Anemy: Yup - added a test for it.

PR #38, opened 2020-03-19
Added script to install extension locally

mmarcon:
npm script to install the extension locally and `vsce` as a dev dependency so the npm scripts
work out of the box without requiring a `npm i -g`.

This way, I don't need to run the extension in debug mode from the project, I can just `git pull` and run `npm run local-install` and I have it running.


mmarcon:
Yes, you are right. I will also make the change in the other PR that refactors the README file.

mmarcon:
I am going to abandon this PR and do everything in https://github.com/mongodb-js/vscode/pull/37 so there is no need to resolve conflicts.

PR #39, opened 2020-03-20
VSCODE-51: Code playground template

mmarcon:
In addition to showing a default template to create the right expectations with users, we give them the possibility to disable the default template in settings.

## Motivation and Context
We want to help users understand what type of commands they can run in the playground.

However, after they are familiar with the extension, they may get annoyed because they need to delete the default template, so we give them the possibility of configuring that they want an empty playground file.

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

## Potential improvements
- We could use real collections from the Atlas test dataset suggest to users to spin up a free cluster and deploy the test datasets to run the test queries and aggregations in a pre-configured environment
- In comments, we can let users know how to disable the default template.


package.json
```diff
@@ -416,6 +414,19 @@
           ],
           "default": "Workspace",
           "description": "When the setting that hides the option to choose where to save new connections is checked, this setting sets if and where new connections are saved."
+        },
+        "mdb.playgroundTemplate": {
+          "type": "string",
+          "enum": [
```
Anemy: Are we thinking about adding more templates in the future? Otherwise maybe this can just be a boolean

src/test/suite/mdbExtensionController.test.ts
```diff
@@ -1103,9 +1103,15 @@ suite('MDBExtensionController Test Suite', () => {
...
 
+    const mockGetPlaygroundConfiguration = sinon.fake.returns({
+      get: (prop) => prop === 'playgroundTemplate' && 'default'
+    });
+    sinon.replace(vscode.workspace, 'getConfiguration', mockGetPlaygroundConfiguration);
+
     vscode.commands.executeCommand('mdb.createPlayground').then(() => {
       assert(mockOpenTextDocument.firstArg.language === 'mongodb');
-      assert(mockOpenTextDocument.firstArg.content === '// The MongoDB playground');
+      console.log('xxx', mockOpenTextDocument.firstArg.content);
```
Anemy: left over log

src/templates/playgroundTemplates.ts
```diff
@@ -0,0 +1,17 @@
+const defaultTemplate: string = `//select the database to use
```
Anemy: Just for readability for a template I think making the comments complete sentences might read smoother. 
`// Select the database to use.`
`// Run a find command.`

src/test/suite/mdbExtensionController.test.ts
mmarcon: Oh, yeah. Forgot about that. Thank you for spotting it!

package.json
mmarcon: I was thinking we may want more in the future but to be honest can't really think of a 3rd option now. I will change it to boolean.

mmarcon:
What am I doing wrong? Why do the CI jobs fail?

mmarcon:
> What am I doing wrong? Why do the CI jobs fail?

Never mind. Found the issue.

PR #40, opened 2020-03-20
VSCODE-72: create a new runtime instance for each evaluation

alenakhineika:
## Description

1) Use a new version of `mongosh` with https://github.com/mongodb-js/mongosh/pull/64 merged that fixes missing db interaction commands in browser runtime (eg. `use('db')`).

2) Create a new runtime instance for each run of the playground. It will allow to repeatedly run the same commands and avoid errors such as `Unable to run playground: Identifier ‘s’ has already been declared`.

Example from @mmarcon:

```
const s = 3, l = 1;
db.users.find().skip(s).limit(l);
```

_"When I run this the first time, this works well. But then if I re-run it, I get “Unable to run playground: Identifier ‘s’ has already been declared” which in the REPL makes sense but in the playground case I’d expect that every run starts with a clean context."_

3) Display active connection instead of active db. When switching connections update active connection name in code lenses.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/editors/activeConnectionCodeLensProvider.ts
+++ b/src/editors/activeConnectionCodeLensProvider.ts
@@ -3,7 +3,6 @@ import * as vscode from 'vscode';
 export default class ActiveConnectionCodeLensProvider implements vscode.CodeLensProvider {
   private _codeLenses: vscode.CodeLens[] = [];
   private _connectionController: any;
-  private _activeConnectionName?: string;
   private _onDidChangeCodeLenses: vscode.EventEmitter<
     void
   > = new vscode.EventEmitter<void>();
@@ -18,8 +17,7 @@ export default class ActiveConnectionCodeLensProvider implements vscode.CodeLens
     });
   }
 
-  public setActiveConnectionName(name: string): void {
-    this._activeConnectionName = name;
+  public refresh(): void {
     this._onDidChangeCodeLenses.fire();
   }
 
@@ -36,7 +34,8 @@ export default class ActiveConnectionCodeLensProvider implements vscode.CodeLens
   }
 
   public resolveCodeLens?(codeLens: vscode.CodeLens): vscode.CodeLens {
-    const message = `Active connection is ${this._activeConnectionName}`;
+    const name = this._connectionController.getActiveConnectionName();
+    const message = `Currently connected to ${name}`;
 
     codeLens.command = {
       title: message,
--- a/src/editors/activeConnectionCodeLensProvider.ts
+++ b/src/editors/activeConnectionCodeLensProvider.ts
@@ -1,10 +1,9 @@
 import * as vscode from 'vscode';
 
-export default class ActiveDBCodeLensProvider implements vscode.CodeLensProvider {
+export default class ActiveConnectionCodeLensProvider implements vscode.CodeLensProvider {
   private _codeLenses: vscode.CodeLens[] = [];
   private _connectionController: any;
-  private _runtime: any;
-  private _activeDB?: string;
+  private _activeConnectionName?: string;
   private _onDidChangeCodeLenses: vscode.EventEmitter<
     void
   > = new vscode.EventEmitter<void>();
@@ -19,8 +18,8 @@ export default class ActiveDBCodeLensProvider implements vscode.CodeLensProvider
     });
   }
 
-  public setActiveDB(activeDB: string): void {
-    this._activeDB = activeDB;
+  public setActiveConnectionName(name: string): void {
+    this._activeConnectionName = name;
     this._onDidChangeCodeLenses.fire();
   }
 
@@ -37,9 +36,12 @@ export default class ActiveDBCodeLensProvider implements vscode.CodeLensProvider
   }
 
   public resolveCodeLens?(codeLens: vscode.CodeLens): vscode.CodeLens {
+    const message = `Active connection is ${this._activeConnectionName}`;
+
     codeLens.command = {
-      title: `Active db is ${this._activeDB}`,
-      command: "mdb.runDBHelpInPlayground"
+      title: message,
+      command: "mdb.showActiveConnectionInPlayground",
+      arguments: [message]
     };
 
     return codeLens;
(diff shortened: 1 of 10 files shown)


src/editors/playgroundController.ts
```diff
@@ -17,9 +18,8 @@ const log = createLogger('editors controller');
 export default class PlaygroundController {
   _context?: vscode.ExtensionContext;
```
Anemy: Since we always assign this is the constructor I think we can lose the `?` (same with `_connectionController` and `_activeConnectionCodeLensProvider`) and that should allow us to get rid of an extra check or two.

src/editors/playgroundController.ts
```diff
@@ -125,4 +119,15 @@ export default class PlaygroundController {
       resolve(true);
     });
   }
+
+  public deactivate(): void {
+    if (this._connectionController) {
+      this._connectionController.removeEventListener(
+        DataServiceEventTypes.ACTIVE_CONNECTION_CHANGED,
+        () => {
+          this.showActiveConnectionInPlayground('Playground disconnected');
```
Anemy: I'm not sure I follow the expected behavior here, when the mongodb app deactivates we want to show this message?

src/templates/playgroundTemplate.ts
```diff
@@ -1,12 +1,15 @@
...
+db.myCollection.find({ foo: 'bar' });
+
+// Run an aggregation
 const agg = [
-  {$match: {foo: 'bar'}}
+  { $match: { foo: 'bar' } }
 ];
-db.my_collection.aggregate(agg);
+
+db.myCollection.aggregate(agg);
```
Anemy: 👍

src/test/suite/editors/playgroundController.test.ts
```diff
@@ -49,5 +54,43 @@ suite('Playground Controller Test Suite', () => {
...
       expect(await testPlaygroundController.evaluate('1 + 1')).to.be.equal('2');
     });
+
+    test('evaluate multiple commands at once', async () => {
+      testConnectionController.setActiveConnection(mockActiveConnection);
+
+      expect(await testPlaygroundController.evaluate(`
+        var x = 1;
+        x + 2
+      `)).to.be.equal('3');
```
Anemy: yesss! scripting!

src/editors/activeConnectionCodeLensProvider.ts
```diff
@@ -1,10 +1,9 @@
 import * as vscode from 'vscode';
 
-export default class ActiveDBCodeLensProvider implements vscode.CodeLensProvider {
+export default class ActiveConnectionCodeLensProvider implements vscode.CodeLensProvider {
   private _codeLenses: vscode.CodeLens[] = [];
   private _connectionController: any;
-  private _runtime: any;
-  private _activeDB?: string;
+  private _activeConnectionName?: string;
```
Anemy: If a playground is opened or created before a connection is established, would this be `undefined`? I think another way we could do this which might allow us not to keep the connection name here would be to call `this._connectionController.getActiveConnectionName()` in the method `resolveCodeLens` for the connection name, and then we can handle the case where there is no active connection there. It also might allow us not to have the event listener `ACTIVE_CONNECTION_CHANGED` altogether.

src/templates/playgroundTemplate.ts
```diff
@@ -1,12 +1,15 @@
...
 `;
```
mmarcon: @Anemy had a preference for closing the sentences in the comments with a `.`. I think it makes sense, although I don't have a strong opinion either way.

src/editors/activeConnectionCodeLensProvider.ts
```diff
@@ -37,9 +36,12 @@ export default class ActiveDBCodeLensProvider implements vscode.CodeLensProvider
   }
 
   public resolveCodeLens?(codeLens: vscode.CodeLens): vscode.CodeLens {
+    const message = `Active connection is ${this._activeConnectionName}`;
```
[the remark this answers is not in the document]
mmarcon: 👍

src/templates/playgroundTemplate.ts
alenakhineika: I don't have a strong opinion either, I prefer without, but I will return them back if having them makes sense for both of you.

src/editors/playgroundController.ts
alenakhineika: Yeah, I just added it to check when deactivation happens, but have never seen this message actually. What do you think is better here? Empty function?

src/editors/activeConnectionCodeLensProvider.ts
alenakhineika: If a playground is opened or created before a connection is established, it will be no code lenses at all. So we don't want to show an active connection name when there is no connection. In case if we get rid of ACTIVE_CONNECTION_CHANGED listener we won't be able to update code lenses with active connection state if we didn't change a state of a playground. I mean we need to type something to invoke code lenses update. And I wanted to change the name in real-time.  Pls, try to play with a playground and let me think how does it feel like it works now.

src/editors/playgroundController.ts
Anemy: I think empty function is probably better just so we preserve what is in the `output`.

src/editors/activeConnectionCodeLensProvider.ts
Anemy: Ah gotcha. It looks like `ACTIVE_CONNECTION_CHANGED` is definitely a useful event to have. I think we can avoid storing the `_activeConnectionName` on the code lens though if we get the name from `this._connectionController.getActiveConnectionName()` in the `resolveCodeLens` method, since the code lens is resolved everytime `this._onDidChangeCodeLenses.fire()` is called. I'd like to avoid duplicating the storage of the connection name if possible, tried it on local and it seemed to do the trick.

alenakhineika:
Thanks for the awesome recommendations in the review! So what I did:

- Returned `.` to the template 😏
- Renamed code lenses text prefix
- Removed the message after removing the listener
- Added a test that runs the playground two times in a raw and checks that the second result is correct. That covers the use case form the PR description
- Removed extra checks for context and connection controller
- Removed the connection name property from code lenses class

Anemy:
lgtm!

PR #42, opened 2020-03-24
Sort collections in tree-view

mmarcon:
## Description
Sorts collections in alphanumerical order.

### Checklist

## Motivation and Context

I expect collections to be in alphanumerical order, otherwise I have to scan them all to find the one I need.

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Open Questions
- Not sure why they don't come back already in the right order.

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/templates/playgroundTemplate.ts
+++ b/src/templates/playgroundTemplate.ts
@@ -1,27 +1,15 @@
 const template: string = `// Select the database to use.
-use("test");
-
-// Insert a few documents.
-db.sales.insertMany([
-  { "_id" : 1, "item" : "abc", "price" : 10, "quantity" : 2, "date" : new Date("2014-03-01T08:00:00Z") },
-  { "_id" : 2, "item" : "jkl", "price" : 20, "quantity" : 1, "date" : new Date("2014-03-01T09:00:00Z") },
-  { "_id" : 3, "item" : "xyz", "price" : 5, "quantity" : 10, "date" : new Date("2014-03-15T09:00:00Z") },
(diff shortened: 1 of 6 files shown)


alenakhineika:
Maybe it would be nice to have a test for it?

mmarcon:
@alenakhineika you are right. Let me add a test in a little bit.

mmarcon:
In fact, looks like I made a test fail. I will do the testing part to relax a bit once I am done with drafting the demo script.

Anemy:
Hmmm those might not be from this work - I got the same failures on my pr.
Ah yup - I'll fix in https://github.com/mongodb-js/vscode/pull/43

src/templates/playgroundTemplate.ts
```diff
@@ -1,27 +1,15 @@
...
 
 // Run a find command.
-db.sales.find({ date: { $gte: new Date("2014-04-04"), $lt: new Date("2014-04-05") } });
+db.myCollection.find({ foo: 'bar' });
 
 // Run an aggregation.
-const aggregation = [
-  { $match: { date: { $gte: new Date("2014-01-01"), $lt: new Date("2015-01-01") } } },
-  { $group: { _id : "$item", totalSaleAmount: { $sum: { $multiply: [ "$price", "$quantity" ] } } } }
+const agg = [
```
Anemy: Maybe aggPipeline? agg is good too.

mmarcon: This is just a revert of a commit that should not have been in this PR

PR #44, opened 2020-03-24
VSCODE-21: Connect form react app

Anemy:
https://jira.mongodb.org/browse/VSCODE-21

This is the first PR in adding a connect form to the vscode plugin. This PR adds a react/typescript app inside of `src/views/connect-form-app` which compiles into `connect-form/connectForm.js` which is required in the extension's webview. The styling of the app pulls in vscode's themed css variables. 

Currently the contents of the react app view is the same as the current just  html webview so there isn't much changing for a user - yet. This is the first step in setting up for adding components from compass-connect's project to build the connect form.

Considerations
- As we go forward, I think we'll want to make this react app more generalized as a single page app that uses react router for different views, but since this is the first view I think this is a good start.
- Currently the app is pulling from the global `package.json`. We could have it have it's own package json to keep things more isolated. Same with eslint, it's using the project's one. Open to thoughts here. I think the benefit of sharing the same dependencies is that we might in future want to reuse some of the files from the webview app inside of the extension.

Did a slight tweak in `update-grammer.js` to avoid using `URL` since it's a possibly reserved keyword
https://developer.mozilla.org/en-US/docs/Web/API/URL/URL


.eslintrc
```diff
@@ -4,18 +4,27 @@
...
-  "plugins": ["@typescript-eslint"],
+  "plugins": [
+    "@typescript-eslint"
+  ],
   "extends": [
     "eslint:recommended",
     "plugin:@typescript-eslint/eslint-recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:@typescript-eslint/recommended-requiring-type-checking",
-    "mongodb-js/node"
```
imlucas: fwiw: you can just swap `mongodb-js/node` to or add `mongodb-js/react` which uses that plugin.

Anemy:
Going to merge for now, if things or ideas come up we can change this app structure np. I'll swap in our eslint rules in the branch I'm working on now for pulling in the connect components

PR #45, opened 2020-03-24
VSCODE-67: add a confirmation when running a playground

alenakhineika:
## Description

Put running a playground behind a prompt to avoid users accidentally messing up their data or doing unintended queries.

- Before running commands in a playground show the `Are you sure you want to run this playground against [CONNECTION_NAME]?` confirmation message.
- Add a setting that can be turned off in case people would like to live on the wild side.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/editors/playgroundController.ts
+++ b/src/editors/playgroundController.ts
@@ -93,6 +93,24 @@ export default class PlaygroundController {
 
   runAllPlaygroundBlocks(): Promise<boolean> {
     return new Promise(async (resolve) => {
+      const activeConnection = this._connectionController.getActiveDataService();
+      const shouldConfirmRunAll = vscode.workspace
+        .getConfiguration('mdb')
+        .get('confirmRunAll');
+
+      if (activeConnection && shouldConfirmRunAll === true) {
+        const name = this._connectionController.getActiveConnectionName();
+        const confirmRunAll = await vscode.window.showInformationMessage(
+          `Are you sure you with to run this playground against ${name}?`,
+          { modal: true },
+          'Yes'
+        );
+
+        if (confirmRunAll !== 'Yes') {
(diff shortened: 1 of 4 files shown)


src/editors/playgroundController.ts
```diff
@@ -93,6 +93,24 @@ export default class PlaygroundController {
...
     return new Promise(async (resolve) => {
+      const activeConnection = this._connectionController.getActiveDataService();
+      const shouldConfirmRunAll = vscode.workspace
+        .getConfiguration('mdb')
+        .get('confirmRunAll');
+
+      if (activeConnection && shouldConfirmRunAll === true) {
+        const name = this._connectionController.getActiveConnectionName();
+        const confirmRunAll = await vscode.window.showInformationMessage(
+          `Are you sure you with to run this playground against ${name}?`,
```
Anemy: What are your thoughts on adding a second line here:
`Are you sure you with to run this playground on connection ${name}? This confirmation can be disabled in the extension settings.`

src/test/suite/editors/playgroundController.test.ts
```diff
@@ -116,5 +119,29 @@ suite('Playground Controller Test Suite', () => {
         }
       );
     });
+
+    test('show a confirmation message before running commands in a playground', (done) => {
```
Anemy: nice - I think this test also covers the case where a user doesn't hit `yes`.

alenakhineika: I replaced this test with two, where we can check both states when prompt should be shown and when not. Could you please take a look again.

PR #51, opened 2020-04-01
Retrieve the connection string from the right place

mmarcon:
The previous implementation was always returning null.

## Description

I realized the shell was always connected to localhost and I noticed the code was looking for the connecting connection instead of the active one.

Additionally, added error message when attempting to launch a shell without an active connection.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -511,11 +511,11 @@ export default class ConnectionController {
   }
 
   public getActiveConnectionDriverUrl(): string | null {
-    if (!this._connectingConnectionId) {
+    if (!this._currentConnectionId) {
       return null;
     }
 
-    return this._savedConnections[this._connectingConnectionId].driverUrl;
+    return this._savedConnections[this._currentConnectionId].driverUrl;
   }
 
   public addEventListener(
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
       },
       {
         "command": "mdb.openMongoDBShell",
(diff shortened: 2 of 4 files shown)


src/connectionController.ts
```diff
@@ -511,11 +511,11 @@ export default class ConnectionController {
   }
 
   public getActiveConnectionDriverUrl(): string | null {
-    if (!this._connectingConnectionId) {
+    if (!this._currentConnectionId) {
       return null;
     }
 
-    return this._savedConnections[this._connectingConnectionId].driverUrl;
```
alenakhineika, line 518: It doesn't do a connection itself, it just gets the connection name for code lense. I just tried these two ids `this._connectingConnectionId` and `this._currentConnectionId` and the playground behaves identically. How to reproduce your problem?

mmarcon: It's not the playground, it's the shell launcher: Command + Shift + P, "Launch MongoDB Shell".

alenakhineika: Oh got it!

Anemy:
Added a test for making sure `openMongoDBShell` shows an error when we aren't connected, and it sets the environment variable correctly to the current connection when we are connected.
lgtm once tests pass

mmarcon:
Thank you @Anemy!

PR #53, opened 2020-04-04
VSCODE-79: move playground to the language server

alenakhineika:
## Description

- In this PR I added instantiation of Language Server to the mdbExtensionController and moved playground execution to this Language Server.

- The playground makes a call to the Language Server Client and sends two parameters: code that needs to be evaluated and driverUrl.

- The Language Server Client sends a request with these values to the Language Server Server.

- Looks like we can't use DataService in the Language Server Server. Probably because it extends from EventEmitter. But we can use CliServiceProvider from '@mongosh/service-provider-server' instead. It accepts 2 params driverUrl and driverOptions. It provides similar methods as DataService does: https://github.com/mongodb-js/mongosh/blob/master/packages/service-provider-server/src/cli-service-provider.ts. Please let me know if you have any objections? The investigation ticket was created: https://jira.mongodb.org/browse/VSCODE-81

- I have noticed that the "Extension + Server Inspector" and "Attach to Language Server" runners from Launch file do not work. The first thing it should be 6009 instead of 6005, but even if I change it, Vscode complains anyway that a connection is refused.

- Moved code from the `activate` method of the Language Server to the `constructor` in order to set `this.client` value in the constructor and get rid of later checks on undefined in each method with sendRequest.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/language/server.ts
+++ b/src/language/server.ts
@@ -13,6 +13,8 @@ import {
   TextDocumentPositionParams,
   RequestType
 } from 'vscode-languageserver';
+import { ElectronRuntime } from '@mongosh/browser-runtime-electron';
+import { CliServiceProvider } from '@mongosh/service-provider-server';
 
 // Create a connection for the server. The connection uses Node's IPC as a transport.
 // Also include all preview / proposed LSP features.
@@ -275,9 +277,17 @@ connection.onCompletionResolve(
 /**
  * Execute the entire playground script.
  */
-connection.onRequest('executeAll', (event) => {
-  // connection.console.log(`executeAll: ${JSON.stringify(event)}`);
-  return '';
+connection.onRequest('executeAll', async (params) => {
+  try {
+    const connectionOptions = params.connectionOptions || {};
+    const runtime = new ElectronRuntime(
+      await CliServiceProvider.connect(params.connectionString, connectionOptions)
+    );
+
+    return await runtime.evaluate(params.codeToEvaluate);
+  } catch (error) {
+    throw new Error('Unable to connect to MongoDB instance. Make sure it is started correctly.');
+  }
 });
 
 /**
--- a/src/language/server.ts
+++ b/src/language/server.ts
@@ -278,16 +278,12 @@ connection.onCompletionResolve(
  * Execute the entire playground script.
  */
 connection.onRequest('executeAll', async (params) => {
-  try {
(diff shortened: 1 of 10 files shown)


alenakhineika:
Hey @imlucas, after our whiteboarding meeting, I have realized, that I need to try it right away, instead of getting deeper into theory. It will help me to better understand the concept. This is how I see the moving of playground to the language server. And what do you think about using  CliServiceProvider instead of DataService? Looking forward to your feedback.

src/language/server.ts
```diff
@@ -275,9 +277,17 @@ connection.onCompletionResolve(
...
-  // connection.console.log(`executeAll: ${JSON.stringify(event)}`);
-  return '';
+connection.onRequest('executeAll', async (params) => {
+  try {
+    const connectionOptions = params.connectionOptions || {};
+    const runtime = new ElectronRuntime(
+      await CliServiceProvider.connect(params.connectionString, connectionOptions)
+    );
+
+    return await runtime.evaluate(params.codeToEvaluate);
```
Anemy, line 287: I think we could move this error into a new try that way a code error can be better handled. Currently if there's an error in a playground's code, like calling a method that doesn't exist the error message is `Unable to connect to MongoDB instance. Make sure it is started correctly.`. Does the runtime throw an error that we can use?

src/test/suite/editors/playgroundController.test.ts
```diff
@@ -8,56 +11,169 @@ import {
...
+  const docPath = path.resolve(__dirname, '../../../../src/test/fixture', docName);
 
-const sinon = require('sinon');
+  return vscode.Uri.file(docPath);
+};
+
+/**
+ * Opens the MongoDB playground
+ */
+export async function openPlayground(docUri: vscode.Uri) {
```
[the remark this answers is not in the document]
alenakhineika: Removing it!

alenakhineika:
Created a ticket for cancellation token: https://jira.mongodb.org/browse/VSCODE-82

src/language/server.ts
alenakhineika: Removed try-catch at all, since they are being caught in the playground controller anyway.

PR #54, opened 2020-04-07
VSCODE: use cancellation token

alenakhineika:
## Description

Investigating how to cancel long-running playground scripts.

I use Node's `worker_threads` to evaluate playground scripts in a separate thread to be able to stop infinite loops. The language server design does not allow interaction with the language server server directly. All requests should go through the language server client. In this case, we can't stop the infinite loop in the language server server via using the `this.client.stop()` method, since at this point the server is already unresponsive.

I also use `CancellationTokenSource` to listen for cancelation events in the language server server that fired in the language server client via `this._source?.cancel();`.

Workers namespace conflicts with typescript that prints the error during lining: `Cannot redeclare block-scoped variable 'Worker'.` It doesn't appear each time, so we probably should just keep on eye on it.

The following ticket was created to double-check if the current implementation covers all edge cases: https://jira.mongodb.org/browse/COMPASS-4238

### Checklist
- [x] Show running progress in the notification area with support for cancellation 
- [x] Cancel playground scripts including ones with infinite loops
- [x] Improve async tests for showing confirmation messages before running a playground
- [x] New test for confirmation messages
- [x] New test for cancellation of playground scripts
- [x] Configure LSP inspector to stream logs to a webview

**How to use LSP inspector:**

- Install the [Language Server Protocol Inspector](https://marketplace.visualstudio.com/items?itemName=octref.lsp-inspector-webview) webview extension.
- Run the MongoDB extension.
- Activate the extension using any activation command.
- Open a playground.
- Check `mongodbLanguageServer.trace.server` settings.
- Run command "LSP Inspector: Start LSP Inspector" to open the LSP inspector webview.
- Run command "LSP Inspector: Start Stream LSP Logs" to start streaming logs to the LSP inspector port specified by the `languageServerExample.port` setting (The default value is 7000).

lsp-stream-logs

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/language/server.ts
```diff
@@ -277,13 +282,61 @@ connection.onCompletionResolve(
...
+      'The long running playground operation was canceled'
+    );
+
+    // Try to close mongoClient to free resources
+    worker.postMessage('terminate');
+
+    // Closing mongoClient...
+    // We can't wait for the actual result from cancelAll() function
+    // because it might never return a result in case of infinite loops
+    await sleep(3000);
```
alenakhineika, line 332: I have found a way to call `mongoClient.close` before the termination of the worker. This delay is not visible for a user since the cancelation message already sent to UI. We need this delay to let mongoClient to be closed before we terminate the worker. I am open to suggestions.

src/language/server.ts
```diff
@@ -277,13 +282,61 @@ connection.onCompletionResolve(
...
  */
-connection.onRequest('executeAll', async (params) => {
-  const connectionOptions = params.connectionOptions || {};
-  const runtime = new ElectronRuntime(
-    await CliServiceProvider.connect(params.connectionString, connectionOptions)
-  );
+connection.onRequest('executeAll', (params) => {
+  // Use Node worker threads to isolate each run of a playground
+  // to be able to cancel evaluation of infinite loops
+  const worker = new WorkerThreads(path.resolve(__dirname, 'worker.js'), {
```
alenakhineika, line 288: There is an issue with support for `.ts` files. Trying to run a `.ts` file in a worker thread results in error:
`The worker script extension must be “.js” or “.mjs”. Received “.ts”`

As a workaround, I require `.js` file from the out folder when creating a worker.

src/language/server.ts
mcasimir: I don't see anything bad with that, except that probably we don't need it.

I'd expect `terminate()` to close everything, socket connections included. If that is the case the client would be terminated, just not gracefully.

In case you'd like to close the connection more gently that seems ok to me.

src/language/languageServerController.ts
```diff
@@ -72,26 +72,48 @@ export default class LanguageServerController {
...
   }
 
-  deactivate(): Thenable<void> | undefined {
-    return this.client.stop();
+  deactivate(): void {
+    // Stop the language server
+    this.client.stop();
   }
 
   executeAll(codeToEvaluate: string, connectionString: string, connectionOptions: any = {}): Thenable<any> | undefined {
```
mcasimir, line 96: Super off-topic but isn't `Thenable<any>` enough?

src/language/server.ts
mcasimir: If you find this message passing with events a bit messy there are few libraries that helps a bit: https://www.npmjs.com/package/comlink

src/language/server.ts
mcasimir: Makes sense. I know you guys are going to use Webpack here, then maybe https://webpack.js.org/loaders/worker-loader/ or some similar 3rd-party plugin may help.

src/language/languageServerController.ts
alenakhineika: Before refactoring, the `executeAll` function was returning the `this.client.sendRequest` method and `this.client` could be undefined. Now when I wrapped it into promise you are right, it returns always `Promise<any>` type. Nice catch! Will update it.

src/language/server.ts
alenakhineika: Yes, we are going to webpackify the extension any time soon. Hope we will be able to configure it to get rid of this workaround. We also do compile before testing anyway, not because of this particular issue, and extension reads all files from the out folder when running. So probably this it not so much harm at this point. And do you use poor js workers only for tests in `mongosh`? The real `shell-api` uses the native workers, right?

src/language/server.ts
alenakhineika: I prefer using native modules when possible to not add any shady and unexpected behavior to the project. And to be honest I took a look at this package and it looks more confusing to me than original API 😀

src/language/worker.ts
```diff
@@ -1,16 +1,30 @@
...
+let serviceProvider: CliServiceProvider;
 
-const executeAll = async (codeToEvaluate, connectionString, connectionOptions = {}) => {
+// Close mongoClient after each runtime evaluation
+// to make sure that all resources are free and can be used with a new request.
+// The mongoClient.close method closes the underlying connector,
+// which in turn closes all open connections.
+// Once called, this mongodb instance can no longer be used.
+const cancelAll = (): void => {
+  (serviceProvider as any).mongoClient.close(false);
```
imlucas, line 19: Since we're cancelling, why not pass true here to force close?

src/editors/playgroundController.ts
```diff
@@ -140,26 +142,45 @@ export default class PlaygroundController {
...
-      try {
+      // Show a running progress in the notification area with support for cancellation
+      await vscode.window.withProgress({
+        location: ProgressLocation.Notification,
+        title: 'Running a playground...',
+        cancellable: true
+      }, async (progress, token) => {
+        token.onCancellationRequested(() => {
+          // If a user clicked the cancel button terminate all playground scripts
+          this._languageServerController.cancelAll();
```
imlucas, line 161: Couldn't we be executing more than 1 playground at a time? Perhaps something to come back to later.

src/language/server.ts
```diff
@@ -277,13 +282,64 @@ connection.onCompletionResolve(
...
+    // to be able to cancel evaluation of infinite loops.
+    //
+    // There is an issue with support for `.ts` files.
+    // Trying to run a `.ts` file in a worker thread results in error:
+    // `The worker script extension must be “.js” or “.mjs”. Received “.ts”`
+    // As a workaround require `.js` file from the out folder.
+    //
+    // TODO: After webpackifying the extension replace
+    // the workaround with some similar 3rd-party plugin
+    const worker = new WorkerThreads(path.resolve(__dirname, 'worker.js'), {
```
imlucas, line 297: This is excellent!

src/language/worker.ts
alenakhineika: We were experimenting with @mcasimir and only this one was actually working. Also, we can replace it soon with mongosh public function call: https://github.com/mongodb-js/mongosh/pull/104
Was merged recently to master, not sure if it is published already.

src/editors/playgroundController.ts
alenakhineika: We can open several playgrounds. I create a new Worker for every run of a playground. So it should be a new serviceProvider and mongoClient for each thread. Probably it should be fine 🤔 Or do you think they can somehow interfere each other?

src/language/server.ts
```diff
@@ -277,13 +278,69 @@ connection.onCompletionResolve(
...
-  return await runtime.evaluate(params.codeToEvaluate);
+      return resolve(result);
+    });
+
+    // Listen for cancellation request from the language server client
+    token.onCancellationRequested(async () => {
+      connection.console.log('Playground cancellation requested');
+      connection.sendNotification(
+        'showInformationMessage',
+        'The long running playground operation was canceled'
```
Anemy, line 321: Maybe `The running playground operation was canceled.` - since it might not be a long running operation that is cancelled. Cancelling a playground that is running is super smooth.

src/editors/playgroundController.ts
```diff
@@ -140,26 +147,45 @@ export default class PlaygroundController {
...
 
       const activeEditor = vscode.window.activeTextEditor;
       const codeToEvaluate = activeEditor?.document.getText() || '';
       let result;
 
-      try {
+      // Show a running progress in the notification area with support for cancellation
+      await vscode.window.withProgress({
+        location: ProgressLocation.Notification,
+        title: 'Running a playground...',
```
Anemy, line 161: Maybe `Running MongoDB playground...` maybe we could also put the name of the database it's running against?

src/test/suite/editors/playgroundController.test.ts
```diff
@@ -330,5 +354,26 @@ suite('Playground Controller Test Suite', () => {
...
+        async (dataService) => {
+          testConnectionController.setActiveConnection(dataService);
+
+          testLanguageServerController.executeAll('while (1===1) {}', 'mongodb://localhost:27018');
+
+          await testLanguageServerController.cancelAll();
+
+          const result = await testLanguageServerController.executeAll('4 + 4', 'mongodb://localhost:27018');
+
+          expect(result).to.be.equal('8');
```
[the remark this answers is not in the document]
alenakhineika: It is more about real-life testing. We had a call with Lucas and agreed that we omit manual connections closing since terminating the entire thread should do the job. But we created a ticket as kind of a reminder to review it later. This ticket can be closed in the future if no problem appears. The more complex tests can be implemented after emerging async rewriting in `mongosh`. Currently, when a playground has several commands they can be executed in the random sequence what leads to unexpected results. See: https://mongodb.slack.com/archives/GTTUKDWRH/p1586851227003500

src/language/worker.ts
```diff
@@ -0,0 +1,47 @@
...
+const executeAll = async (
+  codeToEvaluate: string,
+  connectionString: string,
+  connectionOptions: NodeOptions = {}
+) => {
+  try {
+    // Instantiate a data service provider
+    //
+    // TODO: update when `mongosh` will start to support cancellationToken
+    // See: https://github.com/mongodb/node-mongodb-native/commit/2014b7b/#diff-46fff96a6e12b2b0b904456571ce308fR132
```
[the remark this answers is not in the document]
alenakhineika: Don't think they planed it for the nearest future.

src/language/server.ts
```diff
@@ -277,13 +278,69 @@ connection.onCompletionResolve(
...
+      );
+
+      // If there is a situation that mongoClient is unresponsive,
+      // try to close mongoClient after each runtime evaluation
+      // and after the cancelation of the runtime
+      // to make sure that all resources are free and can be used with a new request.
+      //
+      // (serviceProvider as any)?.mongoClient.close(false);
```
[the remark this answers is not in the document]
alenakhineika: We don't close the mongo client. We terminate the worker that kills all running JS scripts including an instance of the runtime with opened mongo client connection.

package.json
```diff
@@ -393,6 +397,37 @@
...
       "title": "MongoDB",
       "properties": {
+        "mongodbLanguageServer.maxNumberOfProblems": {
+          "scope": "resource",
+          "type": "number",
+          "default": 100,
+          "description": "Controls the maximum number of problems produced by the server."
+        },
+        "mongodbLanguageServer.trace.server": {
+          "scope": "window",
```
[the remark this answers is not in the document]
alenakhineika: I prefer to have the possibility to manage logs level.

1) We can configure other logs verbosity using settings provided by VSCode by default.

Screenshot 2020-04-15 at 17 32 29

2) Since users of our extension are developers it can help them as well debugging the playground and open issues if our language server has a bug. I have seen requests to other extensions' developers from their users to allow configure logs, so I believe it can be valuable.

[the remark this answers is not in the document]
Anemy: Cool sounds good.

PR #55, opened 2020-04-08
VSCODE-74: Add SSL/TLS to Connect Form Webview

Anemy:
https://jira.mongodb.org/browse/VSCODE-74

This PR adds SSL configuration to the connect form web view.
Because the webview lives inside an iframe, we request to open a file in the extension, with a message from the iframe, and then pass the file path results back to the webview.
Because SSL values don't live on a connection string/driver url, we have to change up how we store connections to ensure we can store driver options. Now we're storing the connection model as well as the driver url. (We still store driver url to allow us to copy connection strings without having to first parse the connection model which skips the dns lookup). 

Note: Because of the changes in the storing of connections, all current connections will be invalid and not loaded.

Renamed `connectFormView` to `webviewController`, I'm thinking we might have more webviews in the future and they would probably live inside the same react app.

**Screenshots**
___
*Choosing a file*
Screen Shot 2020-04-08 at 12 39 13 PM

*File chosen example*
Screen Shot 2020-04-08 at 12 39 50 PM

*Invalid/No file error*
Screen Shot 2020-04-08 at 12 43 52 PM

Tested on my old laptop and this worked fine saving and loading ssl. (MacOS Catalina was a bit moody with hostnames). (And shows an error on invalid pem keys.)


Anemy:
@alenakhineika Yes we are currently storing as unprotected in the connection model and the driver URI. I don't like the duplication here, but to build a connection model to get the driver uri it currently requires a dns lookup for srv connection strings... Maybe there's a good work around here we can implement actually. I think it would be good to have it just in one place especially for the ticket coming up where we start managing credentials.

PR #56, opened 2020-04-14
Fix: missing appname when connecting with connectionId

mmarcon:
`appname` was being set only when connecting to a new connection but not when connecting to one that was already added. This fixes it.

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


mmarcon:
Ah, yeah, makes sense to move it there.

PR #59, opened 2020-04-16
VSCODE-85: Add default button to tree view

Anemy:
https://jira.mongodb.org/browse/VSCODE-85

This PR adds a button to the tree view for when the user has not yet added a connection.  This should improve the experience for users who have just installed the extension.

Note: If the user adds a connection and removes it, the button does not show up. If they refresh their vscode though, and there are no connections, it will show up. When they're first loading the extension and their connections have not yet loaded, the message also shows up - which might be enough of a reason not to use this button format. From what I've read it looks like there's no way for us to specify that our extension is being activated... https://github.com/microsoft/vscode/issues/59645 This will be less of an issue once we're webpacking and load times are hopefully significantly reduced.

**Screenshots**
___
Screen Shot 2020-04-16 at 7 11 41 PM

Apr-16-2020 18-06-38

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "viewsWelcome": [
       {
         "view": "mongoDB",
-        "contents": "No connections found.\n[Add Connection](command:mdb.connectWithURI)"
+        "contents": "No connections found.\n[Add Connection](command:mdb.connect)"
       }
     ],
     "languages": [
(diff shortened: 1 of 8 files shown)


Anemy:
Added create playground as a dropdown button:
Screen Shot 2020-04-17 at 5 44 24 PM

And changed the connect button behavior to opening the connect form instead of connection string input.

Anemy:
@alenakhineika I think it's useful to have access to it in the UI - right now it's a bit tough to discover just through the command palette. I think this is probably more of a temporary solution until we have a proper playground tree view: https://jira.mongodb.org/browse/VSCODE-86

PR #63, opened 2020-04-20
VSCODE-88: use mongosh parser for the shell symbols completion

alenakhineika:
## Description

- Use mongosh parser for the shell API symbols/methods completion.
- Switch to node driver v3.5.6.
- Remove `JSON.stringify` from the logger and use `util.inspect` instead.
- Async tests improvements.
- Show LSP debug in the language server output channel.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/logging.ts
```diff
@@ -16,35 +17,35 @@ class Logger implements ILogger {
...
-      `${message} ${optionalParams ? JSON.stringify(optionalParams) : ''}`
+      `${message} ${optionalParams ? util.inspect(optionalParams) : ''}`
     );
   }
 
   public debug(message?: any, ...optionalParams: any[]): void {
     this.append(
       'DEBUG',
-      `${message} ${optionalParams ? JSON.stringify(optionalParams) : ''}`
+      `${message} ${optionalParams ? util.inspect(optionalParams) : ''}`
```
Anemy, line 27: 👍

src/test/suite/language/languageServerController.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    fakeShowErrorMessage = sandbox.stub(vscode.window, 'showErrorMessage');
+
+    await openPlayground(getDocUri('test.mongodb'));
+  });
+
+  after(() => {
+    sandbox.restore();
+  });
+
+  suite('user is connected and runs math operations in playground', () => {
```
Anemy, line 46: Nice full server tests.

package.json
```diff
@@ -523,6 +523,7 @@
     "uuid": "^7.0.0",
     "vscode-languageclient": "^6.1.3",
     "vscode-languageserver": "^6.1.1",
+    "vscode-languageserver-textdocument": "^1.0.1",
```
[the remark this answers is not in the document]
alenakhineika: We need a document manager to create a connection in the language server.

https://github.com/mongodb-js/vscode/blob/VSCODE-9-shell-completion/src/language/server.ts#L23

alenakhineika:
@Anemy updated the MongoDB service to return no completions if connecting or disconnecting. About having description for completion yeah would be nice. But currently, mongosh does not return this information.  We can easily add it later when mongosh will start to return more detailed completions.

alenakhineika:
@Anemy there is no ticket about adding it to mongosh. They also need it for compass shell completion, but it wasn't planned for the nearest future.

PR #64, opened 2020-04-21
VSCODE-64: Webpack extension bundle

Anemy:
https://jira.mongodb.org/browse/VSCODE-64

This PR webpacks the extension into the `dist` directory.
There are 4 seperate webpack entries:
- extension - The main extension file.
- languageServer
- languageServerWorker
- webviewApp

One of the bigger code changes in this pr is making all of the local dependencies (images & other webpack bundles) pull relative to the extension path. I added a singleton for referencing the extension context for getting the extension path - I think this is should be an okay use for a singleton.

We set the `devtool` webpack setting as `source-map` to help debug errors, and we also don't minify for this reason. Benefits of open source :)

VSCode has its own `keytar` implementation, so we list it as an external dependency because the connection model would otherwise pull it in and we would get an error because it's an unsupported native node module.

**Current size of the dev bundles:**
extension.js - 11.1 mb
languageServer.js - 10 mb
languageServerWorker.js - 9.7 mb
webviewApp.js - 2.6 mb
**Prod bundle sizes (not minified):**
extension.js - 3.4mb
languageServer.js - 4.5mb
languageServerWorker.js - 4.4mb
webviewApp.sj - 1.1mb

.vsix file size: 4 mb

I'd like to add some benchmarks to watch the sizes of these and the build times. I'm also tweaking some of the node scripts. Once those are done I'll take this PR out of draft and ping reviewers.

Didn't get around to creating a good way of benchmarking the sizes and times. Creating new tickets for those to handle later: https://jira.mongodb.org/browse/VSCODE-92

For node commands I wanted to keep the same functionality and ease of debugging that we currently have. I'm thinking of ways we can improve this...


Anemy:
@alenakhineika Ah thanks for noticing - those are numbers from before the rebase - I've updated them now. I think the bundles are big though, we can probably find some ways to cut them down.

PR #65, opened 2020-04-22
Simplified M2/M5 Terraform template for Atlas

mmarcon:
## Description
Adds snippet to create M2/M5 Atlas clusters with Terraform

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


mmarcon:
@themantissa updated the snippet. Left local variables intentionally without placeholders that can be replaced by tabbing through so users don't do it too mechanically and think twice before checking the file as is into a public repo.

PR #68, opened 2020-04-28
Add language server file dependency tests

Anemy:
This PR adds two tests to ensure the language server bundles exist.

This PR also cleans up our commands a bit, we were showing some of the commands from the tree view in the command palette, so now it won't show those. When we add future commands lets keep in mind to hide them from the command palette if we don't want them used there.
I'd like to clean up the commands a bit more and have them live in an a `commands.ts` file where they exist as variable we can reference. That can wait though.

Small text change on the create playground action in the tree view: 
Screen Shot 2020-04-28 at 12 17 18 PM
(It was MongoDB: Create New Playground)

### DIFF
--- a/src/language/server.ts
+++ b/src/language/server.ts
@@ -9,7 +9,7 @@ import {
   CompletionItem,
   TextDocumentPositionParams,
   RequestType,
-  TextDocumentSyncKind,
+  TextDocumentSyncKind
 } from 'vscode-languageserver';
 import { TextDocument } from 'vscode-languageserver-textdocument';
 
@@ -54,14 +54,14 @@ connection.onInitialize((params: InitializeParams) => {
       // Tell the client that the server supports code completion
       completionProvider: {
         resolveProvider: true,
-        triggerCharacters: ['.'],
-      },
+        triggerCharacters: ['.']
+      }
       // documentFormattingProvider: true,
       // documentRangeFormattingProvider: true,
       // codeLensProvider: {
       //   resolveProvider: true
       // }
-    },
+    }
   };
 });
 
@@ -121,7 +121,7 @@ const getDocumentSettings = (resource: string): Thenable<ExampleSettings> => {
   if (!result) {
     result = connection.workspace.getConfiguration({
       scopeUri: resource,
-      section: 'mongodbLanguageServer',
+      section: 'mongodbLanguageServer'
     });
     documentSettings.set(resource, result);
   }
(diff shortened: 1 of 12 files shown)


src/language/server.ts
```diff
@@ -9,13 +9,12 @@ import {
   CompletionItem,
   TextDocumentPositionParams,
   RequestType,
-  TextDocumentSyncKind
+  TextDocumentSyncKind,
```
alenakhineika, line 12: I think you accidentally added a trailing comma.

src/language/server.ts
```diff
@@ -55,14 +54,14 @@ connection.onInitialize((params: InitializeParams) => {
       // Tell the client that the server supports code completion
       completionProvider: {
         resolveProvider: true,
-        triggerCharacters: ['.']
-      }
+        triggerCharacters: ['.'],
```
alenakhineika, line 57: And here.

src/language/server.ts
```diff
@@ -122,7 +121,7 @@ const getDocumentSettings = (resource: string): Thenable<ExampleSettings> => {
   if (!result) {
     result = connection.workspace.getConfiguration({
       scopeUri: resource,
-      section: 'mongodbLanguageServer'
+      section: 'mongodbLanguageServer',
```
alenakhineika, line 124: You added trailing commas everywhere to this file :)

src/test/suite/language/mongoDBService.test.ts
```diff
@@ -16,15 +18,26 @@ const INCREASED_TEST_TIMEOUT = 5000;
 
 suite('MongoDBService Test Suite', () => {
   const connection = {
-    console: { log: () => {} },
-    sendRequest: () => {},
-    sendNotification: () => {}
+    console: { log: (): void => {} },
+    sendRequest: (): void => {},
+    sendNotification: (): void => {}
```
alenakhineika, line 23: ✨

src/language/server.ts
Anemy: ahhhhhhhhhh hah. thanks

Anemy:
@alenakhineika Ah yeah - that's been done already. Closed that ticket. Also removed the commas hah thanks for spotting.

alenakhineika:
After fix that you merged in prev PR, Max ran the extension and got the `worker.js missing` error again. After he manually used `npm run watch` it fixes the problem. But isn't it suppose to be cleaned automatically when he ran the extension?

Anemy:
@alenakhineika Yes running the extension or running a local install runs a clean first. I noticed that tests don't run a clean, so I've added that now as well. I'll check with Max and see what the steps were.

mmarcon:
I simply checked out the branch and run the extension (F5). I think doing `npm run watch` is necessary, and that's ok.

alenakhineika:
I also have noticed this:

- I start watcher in terminal (`npm run watch`)
- I do some change in the web view 
- I start the extension
- The change is not applied 
- I restart the extension 
- The change it there

PR #69, opened 2020-04-28
VSCODE-93: db and collection names completion

alenakhineika:
## Description

> Make sure you are connected to a cluster before using a playground. You can't run a playground and you won't get completions if you are not connected.

This PR includes the following changes:

- Clean cached fields completions when reconnecting.
- Provide completion for DB and collection names.
- Remove `runtime.getCompletions()`, because by using our AST and `@mongosh/shell-api` directly we can provide more completion features. By doing so we enable the following completions:
  - `db.` (Database methods)
  - `db.collection.find().` (Cursor methods)
  - `db.collection.aggregate().` (AggregationCursor methods)
  - Support completion for collection names with dashes in them. Replace`db.coll-name.` with `db.['coll-name'].` to have a valid property name.
- Update read me to point to the following issues:

  - By default, VS Code prevents code completion in the snippet mode (editing placeholders in inserted code). Setting the `editor.suggest.snippetsPreventQuickSuggestions` option to `false` stops that and allows for the `db.collection.aggregate()` expression having both snippets completion (eg. `$match`, `$addFields`) and fields completion based on the document schema.
  - By default, VS Code prevents code completion inside string literals. To enable database names completions for `use('')` expression use the following setting:

```
"editor.quickSuggestions": {
  "other": true,
  "comments": false,
  "strings": true
}
```

These settings affect not only MongoDB extension but all installed extensions. If you don't want to change default settings, you can force VS Code to trigger suggestions by clicking `Ctrl+Space` inside a snippet or string literal.

More information here:
https://jira.mongodb.org/browse/VSCODE-93
https://jira.mongodb.org/browse/VSCODE-96

- Remove indentation rules that brake js formatting while typing.

Apr-28-2020 13-28-28

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


alenakhineika:
@Anemy about a second of delay before showing completions the answer is yes, this is because of parsing of the tree. I had a version, that could be a runtime delay, but it wasn't. But! When I was looking at runtime complications I started to think that we actually don't need the shell parser at all. Since we do parsing on our side and can identify more completion cases. The runtime:
- can't handle dots in literals
- can't provide completions for `db['collection'].` and `db.coll-name.`
- does not provide all symbols completions
So, I removed `runtime.getCompletions()` and use `@mongosh/shell-api` directly to support more completion features. By doing so we enable the following completions:
  - `db.` (Database methods)
  - `db.coll-name.` (Collection names with dashes in them)
  - `db.collection.find().` (Cursor methods)
  - `db.collection.aggregate().` (AggregationCursor methods)

alenakhineika:
@Anemy could you pls check if this is fixed:

> I've found that if I start typing in a collection name and then delete a few character, the completions are no longer available. Is this expected?

If not could you pls record a video to show it to me.

Anemy:
@alenakhineika Yup fixed - thanks! For collection names with dots in them should we autocomplete to `db.['collection-name']` instead of `db.collection-name` so that its usable syntax?
Nice getting the Cursor and AggregationCursor methods in 🔥

alenakhineika:
@Anemy pushed an update, now I replace `db.coll-name.` with `db.['coll-name'].` to have a valid property name. For collection names like this `db.companies.` i do not modify the name. Nice suggestion, thx!

alenakhineika:
I need to add one more fix. To remove the dot from `db.['coll-name'].`.

README.md
```diff
@@ -70,6 +72,22 @@ Open VS Code, launch the Commmand Palette (⌘+Shift+P on MacOS, Ctrl+Shift+P on
...
+`editor.suggest.snippetsPreventQuickSuggestions`: By default, VS Code prevents code completion in the snippet mode (editing placeholders in inserted code). Setting this option to `false` stops that and allows for the `db.collection.aggregate()` expression having both snippets completion (eg. `$match`, `$addFields`) and fields completion based on the document schema.
+
+`editor.quickSuggestions`: By default, VS Code prevents code completion inside string literals. To enable database names completions for `use('')` expression use the following setting:
+
+```
+"editor.quickSuggestions": {
+  "other": true,
+  "comments": false,
+  "strings": true
+}
```
Anemy, line 88: 👍

Anemy:
Once you've pushed up the fix and tests are good lgtm !

alenakhineika:
@Anemy uuuuhh this one was a hard one 😅 The documentation says that some properties are deprecated, but api does not match documentation. 

Merged update and tests for this use case.

PR #70, opened 2020-04-29
VSCODE-70: Secure connection credential storage

Anemy:
https://jira.mongodb.org/browse/VSCODE-70

This PR adds usage of the `keytar` module to store connection information. This updates our way of storing connections, we now store metadata about the connection, it's id and name in vscode storage, and then all connection information we store using the id in a string in keytar.

At first I had followed the way vscode recommends: https://code.visualstudio.com/api/advanced-topics/remote-extensions#persisting-secrets but found it wasn't working. Just `require` ing keytar and listing it as an external in webpack seems to do the trick though for local dev.

Unfortunately this will remove any existing saved connections. I think once we've released if we want to do updates to our storing methods we'll want migrations, but because we're still prerelease I think it's probably okay to wipe connections.

When testing we clear any credentials added to the keychain with our service name after the tests have run.

Keytar on linux has a dependency on `libsecret` and `gnome-keyring` which can potentially cause issues. I'd like to get my hands on a linux system and make sure this works. I think in cases on linux where a credential store is unavailable we should not save at all, but still allow a user to add/remove connections.


Anemy:
Opened up https://jira.mongodb.org/browse/VSCODE-110 to fix linux tests. Disabling for now 😬
 
To sum up the issue: When storing credentials in VSCode on headless linux with keytar we need to provide a pre-approved method of storing credentials; we need to unlock a keychain for usage. 
This is tricky because these modules usually require graphical dependencies and don't really provide easy to use command line tooling. So to store in headless linux we most likely need to use, unlock, and provide communication for gnome-keyring, since the more modern version libsecret doesn't really have cli capabilities and other alternatives feel half baked. In `compass-connect` linux tests we do this work around: https://github.com/mongodb-js/compass-connect/blob/master/.travis.yml#L23
In the workaround Lucas did we're relying on a package, `python-gnomekeying` which doesn't exist past ubuntu 16.04. I tried going back to that ubuntu version but I think I still don't have the buses right. Regardless I'd like to get it working on ubuntu latest.
The dbus provides the communication for storing the on the keychain so we need to set that to work with vscode and ensure it stores correctly.
vscode-docker has an approach where they use xvfb, but in their tests they're just mocking keytar :/
I think there's a quick and easily to use solution but I'm a line or two off. Opening up a new ticket for it so we can test these changes and then patch up the linux tests in parallel.

alenakhineika:
Does this affect only tests? Or if user will run vscode on Linux it won't work as well? maybe we can test it on linux virtual machine.

Anemy:
@alenakhineika Tested on my virtual machine and its runs fine - it's the headless environment for testing the linux which brings up the issue.

PR #71, opened 2020-04-30
VSCODE-101: Update a playground template

alenakhineika:
## Description

- Add the `db.sales.drop()` command to a playground template to allow users to run a template script more than once. Since running a script adds data to a collection and all the next runs documents with the same keys can not be inserted and causes the duplicate key error.
- Add a comment to the beginning of the template that reminds users to be connected before running a playground. Only connected users can run a playground or benefit from the completion feature.
- Add a comment that tells users how to use the completion feature inside snippets and string literals.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2263,83 +2263,6 @@
         "resolve": "^1.12.0"
       }
     },
-    "babel-helper-evaluate-path": {
-      "version": "0.5.0",
-      "resolved": "https://registry.npmjs.org/babel-helper-evaluate-path/-/babel-helper-evaluate-path-0.5.0.tgz",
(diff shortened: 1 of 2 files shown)


Anemy:
Is there a chance someone might have a `sales` db already and we drop unintended data? Maybe we could use a more obscure or namespaced name or  we could delete the documents by ids. Looks good - thanks for bumping the dependencies.

alenakhineika:
@Anemy @mmarcon how about `salesCollectionForPlayground` collection name instead of `sales`.

mmarcon:
I would use a different database.

Essentially, just change the `use()` call.

```javascript
use('mongodbVscodePlaygroundDB');
```

What do you think?

alenakhineika:
LGTM, updated.

PR #72, opened 2020-04-30
VSCODE-38: Send metrics when atlas link clicked

alenakhineika:
## Description

- Send metrics when "MongoDB Atlas" and "Create Free Cluster" links are clicked.
- Extract AST visiting methods to a separate Visitor class to improve code readability.
- Mock long-running functions in completion tests.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/views/webview-app/extension-app-message-constants.ts
```diff
@@ -5,7 +5,8 @@ export enum MESSAGE_TYPES {
   CONNECT_RESULT = 'CONNECT_RESULT',
   OPEN_CONNECTION_STRING_INPUT = 'OPEN_CONNECTION_STRING_INPUT',
   OPEN_FILE_PICKER = 'OPEN_FILE_PICKER',
-  FILE_PICKER_RESULTS = 'FILE_PICKER_RESULTS'
+  FILE_PICKER_RESULTS = 'FILE_PICKER_RESULTS',
+  LINK_CLICKED = 'LINK_CLICKED'
```
Anemy, line 9: I think we can make this `LINK_CLICKED` action a bit more descriptive, maybe `CREATE_ATLAS_CLUSTER_LINK_CLICKED`

src/telemetry/telemetryController.ts
```diff
@@ -8,7 +8,8 @@ import { StorageController } from '../storage';
 const log = createLogger('analytics');
 
 export enum TelemetryEventTypes {
-  PLAYGROUND_CODE_EXECUTED = 'playground code executed'
+  PLAYGROUND_CODE_EXECUTED = 'playground code executed',
+  LINK_CLICKED = 'link clicked'
```
Anemy, line 12: I think we can make this `LINK_CLICKED` action a bit more descriptive, maybe `CREATE_ATLAS_CLUSTER_LINK_CLICKED`? - Also maybe for the action inside the webview.

src/views/webview-app/extension-app-message-constants.ts
alenakhineika: We are going to have one metric `link clicked` for any link clicking. For now, it is the Atlas link, but later can be anything.

For example, now we have two links actually that send these metrics:
`{event: 'link clicked', properties: {screen: 'connectScreen', linkId: 'atlasLanding'}}`
`{event: 'link clicked', properties: {screen: 'connectScreen', linkId: 'freeClusterCTA'}}`

So I thought that would be nice to have one action and reuse it for any link. What do you think?

Anemy: ahhh. That makes sense. An idea we could do is in order to ensure these links come in that format is to add a type for tracking events. So in the telemetry controller we could change `public track(eventType: TelemetryEventTypes, properties: object): void { ... }` to 

```typescript
type PlaygroundTelemetryEventProperties = {
  type: string;
};

type LinkClickedTelemetryEventProperties = {
  screen: string;
  linkId: string;
};

type TelemetryEventProperties = PlaygroundEventProperties | LinkClickedTelemetryEventProperties;

public track(eventType: TelemetryEventTypes, properties: TelemetryEventProperties): void { ... }
```
Maybe this is a bit overtyping, but a suggestion.

alenakhineika: I like it a lot!! Pushed changes. Thx for suggesting ✨

PR #73, opened 2020-05-02
VSCODE-87: Add message when successful uri connection

Anemy:
https://jira.mongodb.org/browse/VSCODE-87

This PR adds a message on successful uri connection to the connection form just for some visual feedback. It does not show anything if the connection event was unsuccessful (vscode shows an information message for that in the bottom left). The message is hidden when connect with uri is clicked again, or an action is done on the form. We can give more feedback later on, I thought this will be a nice quick solution for now. Claudia has some thoughts/designs for this connect form so there's a good change this will change around.

May-02-2020 16-20-13

### DIFF
--- a/src/test/suite/views/webviewController.test.ts
+++ b/src/test/suite/views/webviewController.test.ts
@@ -171,7 +171,7 @@ suite('Connect Form View Test Suite', () => {
       html: '',
       postMessage: (message: any): void => {
         assert(message.connectionSuccess);
-        assert(message.connectionMessage === 'Connected.');
+        assert(message.connectionMessage === 'Successfully connected.');
 
         testConnectionController.disconnect();
         done();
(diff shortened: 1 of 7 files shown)


alenakhineika:
I would expect that after connecting I won't see the connect form anymore. I remember conversations that additionally to the connect form we will have an info page or something. Maybe we could switch to this page after connecting when we will have it. Current implementation reminds me of a behavior of the `compass-connect` plugin when it is running in a stand-alone way. Instead of redirecting a user to a database list, we just show a green information message `connected`. Knowing this behavior it is a bit weird to see this message on vscode connect form. It makes me feel that something went wrong. But definitely it is better than to not have a message at all.

Anemy:
@alenakhineika agreed - I think we can improve this a lot. Claudia has some designs. We can find one we like and give a better user experience. This is I think a bit of a quick temporary solution.

PR #74, opened 2020-05-02
Vscode 102 coll names completion improvement

alenakhineika:
## Description

- Allow collection names completion in the middle of expression `db..find()`
- Convert collection name with dashes to array like format in the middle of expression `db.coll-name.find()` ->  `db['coll-name'].find()`
- Provide only collection names in the middle of expression (not shell methods)
- Provide cursor methods only after `find()` and `aggregate()`, not inside objects
- Use different icons for collection names and db methods suggestions
- Refactor Visitor class to use visit node methods instead of multiple condition statements in the AST parsing method
- Disable completions after disconnecting
- Provide collection names completion in variable declarations `const a = db.`
- Provide only collection names and shell db symbol completion after using a cursor and then using `db.`
- Provide completions for db names in string templates:

```
use(`

`)
```

- Provide completions when trigger dot at the next line

```
db
.
```
```
db.collection
.
```

- Provide completions when trigger dot after space

```
db .
```
```
db.collection .
```

- Provide completions when trigger dot after comment

```
// ggg

db.
```
```
// ggg \n

db.
```
```
/*
 * jjj
*/
db.
```
```
use('berlin'); // Comment
db.
```

- Provide completions when code without semicolons
```
use('test')
db
```

- Provide collection names completion at the same line block comment starts

```
use('test');

db. /*
 * Comment
*/
```

- Provide collection names completion at the same line block comment ends

```
use('test');

/*
 * Comment
*/ db.
```

- Provide collection names completion at the same line with end line comment

```
use('test');

db. // Comment
```

- The `window.activeTextEditor` by design returns not only a current active editor, but also an output channel in case of user moved focus to it to scroll the output or copy text from the output. The output editor ended up in `window.activeTextEditor` since the first public API release by accident. Plus `OutputChannel` is just a frontend of OutputService and "Output" pane is managed by OutputEditorInput. When we running a playground we get data from the active editor, since several playgrounds and other files can be open at the same time. To make sure we run data from playground editor only we listen for `vscode.window.onDidChangeActiveTextEditor` events and filter them to get the last active playground editor.

https://github.com/microsoft/vscode/issues/65108
https://github.com/Microsoft/vscode/issues/2857
https://github.com/Microsoft/vscode/issues/72162
https://github.com/microsoft/vscode/issues/58869

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/language/visitor.ts
+++ b/src/language/visitor.ts
@@ -1,3 +1,5 @@
+import * as util from 'util';
+
 const estraverse = require('estraverse');
 const esprima = require('esprima');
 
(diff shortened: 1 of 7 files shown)


Anemy:
Trying this out I found that it wasn't suggesting collection names, rather the cursor methods on a different line from a complete statement:

Screen Shot 2020-05-02 at 6 13 49 PM

Maybe we're not registering it as a new node?

alenakhineika:
Pushed changes that cover the following use cases:

- Provide collection names completion in variable declarations `const a = db.`
- Provide only collection names and shell db symbol completion after using a cursor and then using `db.`
- Provide completions for db names in string templates:

```
use(`

`)
```

alenakhineika:
Also, added support for the following cases:

- When trigger dot is at the next line

```
db
.
```
```
db.collection
.
```

- When trigger dot is after space

```
db .
```
```
db.collection .
```

- When trigger dot is after comment

```
// ggg

db.
```
```
// ggg \n

db.
```
```
/*
 * jjj
*/
db.
```
```
use('berlin'); // Comment
db.
```

- When code without semicolons
```
use('test')
db
```

alenakhineika:
Move use cases covered:

- Provide collection names completion at the same line block comment starts

```
use('test');

db. /*
 * Comment
*/
```

- Provide collection names completion at the same line block comment ends

```
use('test');

/*
 * Comment
*/ db.
```

- Provide collection names completion at the same line with end line comment

```
use('test');

db. // Comment
```

src/language/visitor.ts
```diff
@@ -1,41 +1,179 @@
...
-  constructor() {}
+  _connection: any;
+  _state: CompletionState;
+  _position: { line: number; character: number };
+  _placeholder: string;
+
+  constructor() {
+    this._state = this.getDefaultNodesValues();
+    this._position = { line: 0, character: 0 };
+    this._placeholder = 'TRIGGER_CHARACTER';
```
Anemy, line 25: Does this placeholder change? If not maybe we can make it a constant.

alenakhineika: Does not change. Totally makes sense to make it a constant! Updated.

PR #77, opened 2020-05-06
VSCODE-108 Add command run event

alenakhineika:
## Description

- Fix unhandled promise rejection for `vscode.window.withProgress()`
- Add terraform language configuration since it is being used for snippets but VSCode does not know terraform by default
- Fix telemetry events after switching to webpack
- Add `command run` telemetry event

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


package.json
```diff
@@ -76,40 +76,61 @@
...
+        "configuration": "./languages/mongodb-language-configuration.json"
+      },
+      {
+        "id": "terraform",
+        "aliases": [
+          "Terraform",
+          "terraform"
+        ],
+        "extensions": [
+          ".tf"
```
alenakhineika, line 121: Adding terraform language support

alenakhineika:
@Anemy ah this one is interesting. So before I was creating a cancelable modal like this:

```
async vscode.window.withProgress()
```

Because of `withProgress` is thenable the extension was complaining that we need to handle rejection.

I moved `withProgress` to a separate function and wrapped it into a promise to be able to add catch function and get rid of this console error. This error did not affect any functionality, but I just have added a test anyway to address this issue.

src/test/suite/editors/playgroundController.test.ts
```diff
@@ -79,49 +86,70 @@ suite('Playground Controller Test Suite', () => {
     });
```
Anemy, line 86: Small thing but I think for this test to be a bit more catch all we should test that the error message is what we expect it to be.

PR #87, opened 2020-05-12
VSCODE-113: Add tests for the new connection event

alenakhineika:
## Description

Add tests for the `new connection` event.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -316,7 +316,7 @@ export default class ConnectionController {
     });
   };
 
-  public async getCloudInfoFromDataService(firstServerHostname) {
+  public async getCloudInfoFromDataService(firstServerHostname: string) {
     const cloudInfo = await getCloudInfo(firstServerHostname);
     let isPublicCloud = false;
     let publicCloudName: string | null = null;
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -65,7 +65,7 @@ export default class ConnectionController {
 
   private _statusView: StatusView;
   private _storageController: StorageController;
-  public _telemetryController?: TelemetryController;
+  private _telemetryController: TelemetryController;
 
   // Used by other parts of the extension that respond to changes in the connections.
   private eventEmitter: EventEmitter = new EventEmitter();
@@ -73,7 +73,7 @@ export default class ConnectionController {
   constructor(
     _statusView: StatusView,
     storageController: StorageController,
-    telemetryController?: TelemetryController
+    telemetryController: TelemetryController
   ) {
     this._statusView = _statusView;
     this._storageController = storageController;
@@ -102,6 +102,7 @@ export default class ConnectionController {
     }
 
     let loadedSavedConnection: LoadedConnection;
+
     try {
       const unparsedConnectionInformation = await this._keytar.getPassword(
         this._serviceName,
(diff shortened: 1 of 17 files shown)


src/telemetry/telemetryController.ts
```diff
@@ -110,21 +110,21 @@ export default class TelemetryController {
...
     properties: TelemetryEventProperties
   ): void {
-    const shouldSendTelemetry = vscode.workspace
-      .getConfiguration('mdb')
-      .get('sendTelemetry');
-
     log.info('TELEMETRY track', {
       eventType,
       segmentUserID: this._segmentUserID,
       properties
```
Anemy, line 124: I think when telemetry is disabled we don't want to show this log. How do you feel about moving it into the `if (this.needTelemetry()) {` below?

src/mdbExtensionController.ts
```diff
@@ -137,12 +137,14 @@ export default class MDBExtensionController implements vscode.Disposable {
   ): void => {
     const commandHandlerWithTelemetry = (args: any[]) => {
       // Send metrics to Segment
-      this._telemetryController?.track(
-        TelemetryEventTypes.EXTENSION_COMMAND_RUN,
-        {
-          command
-        }
-      );
+      if (this._telemetryController?.needTelemetry()) {
```
Anemy, line 140: I think because we check `if (this.needTelemetry()) {` in the telemetry controller we can remove this if?

src/editors/playgroundController.ts
```diff
@@ -168,9 +168,9 @@ export default class PlaygroundController {
       codeToEvaluate
     );
 
-    if (result) {
+    if (result && this._telemetryController?.needTelemetry()) {
```
Anemy, line 171: Same here - I think because we check if (this.needTelemetry()) { in the telemetry controller we can remove this if?

src/connectionController.ts
```diff
@@ -441,7 +441,7 @@ export default class ConnectionController {
         this.eventEmitter.emit(DataServiceEventTypes.CONNECTIONS_DID_CHANGE);
         this.eventEmitter.emit(DataServiceEventTypes.ACTIVE_CONNECTION_CHANGED);
 
-        if (this._telemetryController) {
+        if (this._telemetryController?.needTelemetry()) {
```
Anemy, line 444: And here  - I think because we check if (this.needTelemetry()) { in the telemetry controller we can remove this if?

alenakhineika: This one we actually want to keep, since `this.sendTelemetry()` requests cloud info what takes a while and can cause a timeout in tests. But I removed the rest of these checks from other controllers.

src/editors/playgroundController.ts
alenakhineika: I would keep this one as well. This parsing is not very heavy, but anyway we don't want extra time consumption, right?

alenakhineika:
@Anemy Refactored according to your comments ✨ Nice suggestion! While refactoring I have found a couple of fluky async tests and fixed them. Hopefully, this will improve stability of pipelines.

Issue #103, opened 2020-05-18
SSH Tunneling via Identity File fails on Windows


luastoned:
Hi,

I'm trying to access a mongo instance on a secure server with username/private key auth.
The path to the selected key file looks weird (as in being prefixed with `/C:/Users/.../pk`)

On connecting it shows `Failed to connect: Error creating SSH Tunnel: All configured authentication methods failed`

The same method works for Robo3T, so I would expect it to work here too.

Anemy:
@luastoned Hey Lua, we've just pushed the new version now which should fix this issue (0.0.3). Feel free to reopen if the issue hasn't gone away. Thanks for bringing up!

PR #110, opened 2020-05-20
VSCODE-117: Fix file pathing for windows connect

Anemy:
https://jira.mongodb.org/browse/VSCODE-117

Our file paths are relative to the file system (/ at the beginning), when they're read in with node https://github.com/mongodb-js/connection-model/blob/master/lib/connect.js#L42 they are resolved in a way that was adding an extra system /. Linux handled this alright but on windows we were getting errors like `Cannot read file 'C:/C:/some_path_to_file.file'`. This fix just removes the leading / so they can resolve correctly in connect.

Tested manually on windows & linux opening CA & SSH files. Still thinking about how we can best integrated test ssl and ssh auth as they've been breaking points here and in compass.

Removed a bit of unused url code in connection model as well.

### DIFF
--- a/src/views/webview-app/connection-model/connection-model.ts
+++ b/src/views/webview-app/connection-model/connection-model.ts
@@ -1,6 +1,3 @@
-import { format as toURL } from 'url';
-
-import AUTHENICATION_TO_AUTH_MECHANISM from './constants/auth-strategy-to-auth-mechanism';
 import AUTH_STRATEGIES from './constants/auth-strategies';
 import READ_PREFERENCES from './constants/read-preferences';
 import SSL_METHODS from './constants/ssl-methods';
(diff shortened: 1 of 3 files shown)


src/views/webview-app/connection-model/connection-model.ts
```diff
@@ -91,148 +86,6 @@ class ConnectionModel {
...
-    connectionModel.authStrategy === AUTH_STRATEGIES.KERBEROS &&
-    connectionModel.kerberosCanonicalizeHostname
-  ) {
-    result = `${result}&authMechanismProperties=CANONICALIZE_HOST_NAME:true`;
-  }
-
-  return result;
-};
-
 /**
```
alenakhineika, line 89: Nice cleanup!

Anemy:
Fixes https://github.com/mongodb-js/vscode/issues/103

PR #120, opened 2020-05-26
VSCODE-125: Connect with SSL to mongodb shell

alenakhineika:
## Description

- [x] Pass proper SSL options when launching MongoDB Shell. Since we send only connection string text to the terminal we are missing information from SSL files. In case of SSL connection add SSL options to the terminal text.
- [x] Update a connection form and replace certificate and key file pickers with a single `CertificateKey` file picker. In case of certificate and key files provided separately, we won't be able to launch mongodb shell with SSL, since shell expects a single path to the joined file: https://docs.mongodb.com/manual/reference/program/mongo/
- [x] Support SSL connections in playgrounds. Since we do not use data service in the language server and use CliServiceProvider instead we need to download SSL files and send binaries to `MongoClient.connect` instead of files path.
- [x] Use `path.resolve` to traverse the filesystem and maintain cross-platform compatibility for file pickers on the connect form.
- [x] Make tests great again.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/language/mongoDBService.ts
+++ b/src/language/mongoDBService.ts
@@ -59,30 +59,36 @@ export default class MongoDBService {
     );
   }
 
-  private readSslFileSync(sslOption?: string | string[]): any {
-    if (sslOption) {
-      if (Array.isArray(sslOption)) {
-        return fs.readFileSync(sslOption[0]);
-      }
-
-      if (typeof sslOption !== 'string') {
-        return;
-      }
-
-      return fs.readFileSync(sslOption);
+  private readSslFileSync(sslOption: string | string[]): any {
+    if (Array.isArray(sslOption)) {
+      return fs.readFileSync(sslOption[0]);
     }
+
+    if (typeof sslOption !== 'string') {
+      return;
+    }
+
+    return fs.readFileSync(sslOption);
   }
 
   private loadSslBinaries(): void {
-    this._connectionOptions.sslCA = this.readSslFileSync(
-      this._connectionOptions.sslCA
-    );
-    this._connectionOptions.sslKey = this.readSslFileSync(
-      this._connectionOptions.sslKey
-    );
-    this._connectionOptions.sslCert = this.readSslFileSync(
-      this._connectionOptions.sslCert
-    );
+    if (this._connectionOptions.sslCA) {
+      this._connectionOptions.sslCA = this.readSslFileSync(
+        this._connectionOptions.sslCA
+      );
+    }
+
+    if (this._connectionOptions.sslKey) {
+      this._connectionOptions.sslKey = this.readSslFileSync(
+        this._connectionOptions.sslKey
+      );
+    }
+
+    if (this._connectionOptions.sslCert) {
+      this._connectionOptions.sslCert = this.readSslFileSync(
+        this._connectionOptions.sslCert
+      );
+    }
   }
 
   public async connectToServiceProvider(params: {
--- a/src/language/mongoDBService.ts
+++ b/src/language/mongoDBService.ts
@@ -9,8 +9,15 @@ import { Visitor } from './visitor';
 import { ServerCommands, PlaygroundRunParameters } from './serverCommands';
 
 const path = require('path');
+const fs = require('fs');
 
 export const languageServerWorkerFileName = 'languageServerWorker.js';
+
+type SslFileOptions = {
+  sslCA?: string;
+  sslKey?: string;
+  sslCert?: string;
+};
 
 export default class MongoDBService {
   _serviceProvider?: CliServiceProvider;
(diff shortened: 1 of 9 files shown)


src/mdbExtensionController.ts
```diff
@@ -422,10 +438,31 @@ export default class MDBExtensionController implements vscode.Disposable {
       name: 'MongoDB Shell',
       env: { MDB_CONNECTION_STRING: mdbConnectionString }
     });
+
+    if (this.isSslConnection(activeConnectionModel)) {
+      mdbTlsOptionsString = '--tls';
```
alenakhineika, line 443: To test dev mode with self-signed certificates use `mdbTlsOptionsString = '--tls --tlsAllowInvalidCertificates';`

So the question stays here, how to test real case, the one without `--tlsAllowInvalidCertificates` flag.

src/mdbExtensionController.ts
```diff
@@ -422,10 +438,30 @@ export default class MDBExtensionController implements vscode.Disposable {
       name: 'MongoDB Shell',
       env: { MDB_CONNECTION_STRING: mdbConnectionString }
     });
+
+    if (this.isSslConnection(activeConnectionModel)) {
+      mdbTlsOptionsString =
+        '--tls --tlsAllowInvalidCertificates --tlsAllowInvalidHostnames';
```
alenakhineika, line 444: Our connection model also disables them: https://github.com/mongodb-js/connection-model/blob/master/lib/model.js#L634-L637

alenakhineika: TODO: check sslMethod to enable validation for some types of connections.

src/mdbExtensionController.ts
```diff
@@ -397,35 +397,71 @@ export default class MDBExtensionController implements vscode.Disposable {
...
-      mdbConnectionString = activeConnectionModel
-        ? activeConnectionModel.driverUrlWithSsh
-        : '';
-    }
+  public openMongoDBShell(): Promise<boolean> {
+    let mdbConnectionString: any;
+    let mdbTlsOptionsString = '';
 
-    if (!mdbConnectionString) {
+    if (!this._connectionController) {
```
Anemy, line 414: I think we still want to check if there is an active connection here for this check.
Maybe:
`if (!this._connectionController || !this._connectionController.isCurrentlyConnected()) {`

src/mdbExtensionController.ts
```diff
@@ -397,35 +397,71 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+      if (activeConnectionModel?.driverOptions.sslCA) {
+        mdbTlsOptionsString = `${mdbTlsOptionsString} --tlsCAFile ${activeConnectionModel?.driverOptions.sslCA}`;
+      }
+
+      if (activeConnectionModel?.driverOptions.sslCert) {
+        mdbTlsOptionsString = `${mdbTlsOptionsString} --tlsCertificateKeyFile ${activeConnectionModel?.driverOptions.sslCert}`;
+      }
+
+      if (activeConnectionModel?.driverOptions.sslPass) {
+        mongoDBShell['MDB_TLS_CERTIFICATE_KEY_FILE_PASSWORD'] =
```
Anemy, line 455: Ah didn't know environment variables could be set like this. 
Maybe we move the forming of `mdbTlsOptionsString` into a helper to make this function a little smaller?

src/language/mongoDBService.ts
```diff
@@ -43,6 +50,31 @@ export default class MongoDBService {
...
+        connectionOptions.sslCert ||
+        connectionOptions.sslPass)
+    );
+  }
+
+  private loadSslOptionsFromFile(connectionOptions: any): SslFileOptions {
+    const sslOptions: SslFileOptions = {};
+
+    if (connectionOptions.sslCA) {
+      sslOptions.sslCA = fs.readFileSync(connectionOptions.sslCA[0]);
```
Anemy, line 66: Should we have this wrapped in a try catch in case it fails? Might be a place where the user has moved the fail so it would be nice to give them a message saying file reading failed.

src/mdbExtensionController.ts
alenakhineika: Yes also was thinking about it!

alenakhineika: And you can not set env like this, used worn variable name. Thanks for noticing!

src/language/mongoDBService.ts
```diff
@@ -54,13 +102,25 @@ export default class MongoDBService {
...
     if (!this._connectionString) {
       return Promise.resolve(false);
     }
 
+    if (this.isSslConnection(this._connectionOptions)) {
+      try {
+        this.loadSslBinaries();
+      } catch (error) {
+        this._connection.console.log(
+          `SSL FILES read error: ${util.inspect(error)}`
```
Anemy, line 117: 👍

alenakhineika:
@Anemy I wish I could like JIRA tickets! :) looking forward to automating this type of tests ✨

amangoelfv:
i had create an express api system using nodejs. I wrapped it in serverless docs . Now thw api works when i start them with serverless offline start but when i deploy them the cloudwatch shows an error like:
ERROR	Uncaught Exception 	
{
    "errorType": "MongoServerSelectionError",
    "errorMessage": "connection timed out",
    "name": "MongoServerSelectionError",
    "reason": {
        "type": "ReplicaSetNoPrimary",
        "setName": null,
        "maxSetVersion": null,
        "maxElectionId": null,
        "servers": {},
        "stale": false,
        "compatible": true,
        "compatibilityError": null,
        "logicalSessionTimeoutMinutes": null,
        "heartbeatFrequencyMS": 10000,
        "localThresholdMS": 15,
        "commonWireVersion": null
    },
    "stack": [
        "MongoServerSelectionError: connection timed out",
        "    at Timeout._onTimeout (/var/task/node_modules/mongodb/lib/core/sdam/topology.js:430:30)",
        "    at listOnTimeout (internal/timers.js:549:17)",
        "    at processTimers (internal/timers.js:492:7)"
    ]
}


I used mongodb as my database 
please help

mmarcon:
@amangoelfv this error does not seem to be related to MongoDB for VS Code (out VS Code extension). I recommend posting in https://developer.mongodb.com/community/forums/c/drivers-odms-connectors/7.

PR #130, opened 2020-06-08
VSCODE-115: Launch mongoshell cross platform

Anemy:
https://jira.mongodb.org/browse/VSCODE-115

This PR tweaks our usage of `createTerminal` so that we pass the `mongosh` arguments as part of the command. This lets us not have to set/unset environment variables, while keeping the connection string credentials not showing in plain text in the console. This will also help the shell run across platforms since we're not relying on shell syntax for any commands.

- [x] Tested with SSL connection.
- [x] Tested on windows.
- [x] Tested on SSH Tunnel connection.

I think it would be nice to move the launching mongo shell and helpers into another file for organization. Didn't plan on doing it in this PR, we could though.

Jun-08-2020 14-07-53

### DIFF
--- a/src/views/webviewController.ts
+++ b/src/views/webviewController.ts
@@ -100,7 +100,7 @@ export default class WebviewController {
               action: message.action,
               files:
                 files && files.length > 0
-                  ? files.map((file) => path.toNamespacedPath(file.path))
+                  ? files.map((file) => path.resolve(file.path.substr(1)))
                   : undefined
             });
           });
--- a/src/views/webviewController.ts
+++ b/src/views/webviewController.ts
@@ -100,7 +100,7 @@ export default class WebviewController {
               action: message.action,
               files:
                 files && files.length > 0
-                  ? files.map((file) => path.resolve(file.path))
+                  ? files.map((file) => path.toNamespacedPath(file.path))
                   : undefined
             });
           });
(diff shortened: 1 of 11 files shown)


alenakhineika:
Do we also want to tag PRs with a version? Before I was assigning tag only to issues to mark them as ready to make it easier to find and close them after releasing.

Anemy:
>Do we also want to tag PRs with a version? Before I was assigning tag only to issues to mark them as ready to make it easier to find and close them after releasing.

Don't think so, was just trying it.

>the validation returns the following error: hostname required.

Ah - yup. We set a boolean when we its changed instead of just value checking it for default. I can fix when im on my laptop, should be quick.

Anemy:
Ah there's still an issue on windows. Seeing what we can do.

src/views/webviewController.ts
```diff
@@ -100,7 +100,7 @@ export default class WebviewController {
               action: message.action,
               files:
                 files && files.length > 0
-                  ? files.map((file) => path.resolve(file.path))
+                  ? files.map((file) => path.resolve(file.path.substr(1)))
```
Anemy, line 103: Windows was building the file string as `C:/C:/somepath/to/file.txt` I think this should cover us because vscode is giving us absolute file paths. Tested on windows and now we're getting the right file path. Also tested SSL on linux with a file path and it connected with shell.

PR #132, opened 2020-06-08
VSCODE-121: Remote WSL playground fix

Anemy:
https://jira.mongodb.org/browse/VSCODE-121

The node `worker_threads` package had a change in v12.5.0 that returned a promise instead of running synchronously: https://nodejs.org/api/worker_threads.html#worker_threads_worker_terminate
It appears remote WSL (Windows subsystem for Linux) is using the older version of worker threads, while vanilla vscode isn't. https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl I'm going to dig a bit deeper and see if there's a PR we can do so that the worker threads/node version gets bumped up, this fix should cover our use case though.

This PR changes how we terminate the worker to allow it to return possibly synchronously using `await`. No test changes, if we find more errors from WSL maybe it's worth making a testing environment for it.

- [x] Tested playgrounds on Windows with remote WSL ubuntu. (and reproduced the error on another branch).

### DIFF
--- a/src/language/mongoDBService.ts
+++ b/src/language/mongoDBService.ts
@@ -28,7 +28,7 @@ export default class MongoDBService {
   _cachedFields: object;
   _cachedDatabases: [];
   _cachedCollections: object;
-  _cachedShellSymbols: object;
+  _cachedShellSymbols: any;
   _extensionPath?: string;
   _visitor: Visitor;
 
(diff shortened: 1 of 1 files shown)


src/language/mongoDBService.ts
```diff
@@ -449,27 +446,27 @@ export default class MongoDBService {
           'VISITOR found shell collection methods completion'
         );
 
-        return resolve(this._cachedShellSymbols['Collection']);
+        return resolve(this._cachedShellSymbols.Collection);
```
alenakhineika, line 449: Nice! 😏

PR #138, opened 2020-06-23
VSCODE-116: Automate vscode publishing

Anemy:
https://jira.mongodb.org/browse/VSCODE-116

This PR adds a filesize check in our pipeline to make sure the bundle stays under 5MB. It also adds the built .vsix as a build artifact so we can more easily test and share builds.

These docs are very useful: https://code.visualstudio.com/api/working-with-extensions/continuous-integration https://code.visualstudio.com/api/working-with-extensions/publishing-extension

Also doing a bit of double checking/cleanup with this PR.
- Going over our .vscodeignore
- Looking at easy wins for making our extension size a bit smaller / faster

To download the artifact that each build creates on linux:
On the azure pipeline page there's an artifacts link:
Screen Shot 2020-06-24 at 10 54 51 AM
And then the download is in the dropdown on the right:
Screen Shot 2020-06-24 at 10 55 07 AM

To publish, we have a token for the marketplace, so we can use `vsce publish [minor/major/patch]`. This will bump the package json and tag the release with a commit for the version as well as update the marketplace. We should ideally include our change log changes before doing this. https://code.visualstudio.com/api/working-with-extensions/publishing-extension#autoincrementing-the-extension-version

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -73,7 +73,16 @@ steps:
       echo "##vso[task.setvariable variable=extension_version]`npx json -f package.json version`"
     displayName: 'Set extension_version variable from package.json version'
   - bash: |
-      ls -alh mongodb-vscode-$(extension_version).vsix
+      vsixFilename=./mongodb-vscode-$(extension_version).vsix
+      maxsize=5000000 # 5MB
+      filesize=$(stat -c%s "$vsixFilename")
+
+      echo "Size of $vsixFilename = $filesize bytes."
+
+      if (( filesize > maxsize )); then
+        echo "File is over 5MB."
+        exit 1
+      fi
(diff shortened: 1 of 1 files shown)


Anemy:
To download the artifact that each build creates on linux:
On the azure pipeline page there's an artifacts link:
Screen Shot 2020-06-24 at 10 54 51 AM
And then the download is in the dropdown on the right:
Screen Shot 2020-06-24 at 10 55 07 AM

To publish, we have a token for the marketplace, so we can use `vsce publish [minor/major/patch]`. This will bump the package json and tag the release with a commit for the version as well as update the marketplace. We should ideally include our change log changes before doing this. https://code.visualstudio.com/api/working-with-extensions/publishing-extension#autoincrementing-the-extension-version

PR #141, opened 2020-07-03
VSCODE-134: Show indexes in tree view

Anemy:
https://jira.mongodb.org/browse/VSCODE-150
Also fixes schema double click bug: https://jira.mongodb.org/browse/VSCODE-133

This PR adds an indexes to the tree view inside collections. Each index has nodes which are its keys (compound indexes have multiple). Currently the keys have descriptions like `Ascending`, `Descending`, or `Text`. These will be replaced once we add icons in https://jira.mongodb.org/browse/VSCODE-151
The icon for indexes is temporary and will also be replaced in https://jira.mongodb.org/browse/VSCODE-151

Indexes are fetched from the data service which uses https://github.com/mongodb-js/index-model

A good bit of the changes are some refactoring of how we cache previous data fetches with the tree view which fix the bug that crashes on schema multi clip. Apologies for putting the big diff in this PR - I thought it would be a quick fix and then it kept going hah. In the process also found out that we were calling `list dbs` way more than we needed to, so that is also fixed. 
The tree view should feel a bit more snappy and responsive now. I think it could be cleaned up a bit more and the caching could be generalized and moved into something to which all of the classes can inherit from, but we can do that another time.

*Screenshots*
___
Jul-03-2020 11-38-57
*Text index*
Screen Shot 2020-07-03 at 11 57 52 AM


Anemy:
Huh windows npm install issue is probably related to the windows version change in https://github.com/mongodb-js/vscode/commit/e8507d8fb498e154edfa106ec870643169a6819f#diff-fec826feae04e51c0d94076385408bdcL20
Not sure how that would break it though. 🤔 I think this PR can still be looked at in the meantime.

src/explorer/schemaTreeItem.ts
```diff
@@ -97,31 +97,27 @@ export default class SchemaTreeItem extends vscode.TreeItem
...
       if (!this.hasClickedShowMoreFields && this.hasMoreFieldsToShow) {
         return Promise.resolve([
-          ...Object.values(this.childrenCache).map(
-            (cachedField) =>
-              new FieldTreeItem(
-                cachedField.field,
-                cachedField.isExpanded,
-                cachedField.getChildrenCache()
-              )
-          ),
```
Anemy, line 110: Here was the bug with the schema. We were rebuilding the schema's tree items in order to control their expanded state, however we weren't updating the cache with the newly created items. Because of that we were accessing some old nodes.

src/explorer/collectionTreeItem.ts
```diff
@@ -84,53 +108,100 @@ export default class CollectionTreeItem extends vscode.TreeItem
...
+      this._documentListChild,
+      this._schemaChild,
+      this._indexListChild
+    ]);
+  }
+
+  // eslint-disable-next-line complexity
+  needsToUpdateCache(): boolean {
+    return (
+      (this._documentListChild.isExpanded &&
```
mcasimir, line 169: The complexity eslint is complaining about is kinda a real thing here. Is a bit hard to understand these conditions (especially for someone not having much context), and also probably quite easy to miscode them and introduce issues.
Could this code be factored further in a way that reveals the intent of the conditions? Just an idea: 

```
isChildCacheOutOfSync(child) {
   const isExpanded = child.isExpanded;
   const collapsibleState = child.collapsibleState;
   return isExpanded ? collapsibleState !== Expanded : collapsibleState !== Collapsed;
}
```

```
needsToUpdateCache() {
   isChildCacheOutOfSync(this._documentListChild) ||
   isChildCacheOutOfSync(this. _schemaChild) || ...
}
```

src/explorer/connectionTreeItem.ts
```diff
@@ -95,21 +97,37 @@ export default class ConnectionTreeItem extends vscode.TreeItem
...
-        return reject(new Error('Not currently connected.'));
-      }
-      dataService.listDatabases((err: any, databases: string[]) => {
+      dataService.listDatabases((err: Error | undefined, databases: any[]) => {
         if (err) {
           return reject(new Error(`Unable to list databases: ${err.message}`));
         }
 
-        this._childrenCacheIsUpToDate = true;
+        this.cacheIsUpToDate = true;
```
mcasimir, line 130: How is this used?

src/explorer/databaseTreeItem.ts
```diff
@@ -53,7 +55,24 @@ export default class DatabaseTreeItem extends vscode.TreeItem
...
+          this._dataService,
+          pastChildrenCache[collectionName].isExpanded,
+          pastChildrenCache[collectionName].cacheIsUpToDate,
+          pastChildrenCache[collectionName].getDocumentListChild(),
+          pastChildrenCache[collectionName].getSchemaChild(),
+          pastChildrenCache[collectionName].getIndexListChild()
+        );
+      });
+
       return Promise.resolve(Object.values(this._childrenCache));
```
mcasimir, line 76: Why not using `async/await`? What happens if this function throws synchronously, would we need a `try / catch`?

src/explorer/connectionTreeItem.ts
Anemy: We store the `cacheIsUpToDate` value on tree items with asynchronous resources in order to avoid re-fetching the resource when the tree changes. The cache is used to persist the expanded state of its children and their resources, this makes it so that when a connection is collapsed and expanded and list databases is re-run, we can keep the previous expanded state of databases and their collections.

Hmmm... We could use the cache as the flag here and set it to null when it needs to be refreshed. I think that might clean up the constructors of these classes a bit. Some of them have a few more arguments than I like.

src/explorer/databaseTreeItem.ts
Anemy: Ah async await will be better. Easier to read and handle errors. Coming soon to a commit near you

src/explorer/collectionTreeItem.ts
Anemy: Good suggestion - updated. ty

src/explorer/collectionTreeItem.ts
```diff
@@ -41,31 +61,51 @@ export default class CollectionTreeItem extends vscode.TreeItem
...
+
     this._documentListChild = existingDocumentListChild
       ? existingDocumentListChild
       : new DocumentListTreeItem(
         this.collectionName,
         this.databaseName,
         this._type,
         this._dataService,
-        false, // Collapsed.
+        false,
```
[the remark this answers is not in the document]
Anemy: Ah thanks for catching. I do like the practice of leaving a comment where explicit values are passed to constructors/functions.

Anemy:
@alenakhineika Cleaned up the extra Promise & commented back in the constructor variables.
Also added a test for when schema fetches and there are no documents, and when theres an error fetching indexes.

We chatted in slack about public & private. I think this sums it up: we can add private when we think it'll help keep modules isolated and true to their function. Not too much to change in these tree item classes, if we see other parts of the extension wanting to interact with the tree items I think it would be valuable to be more explicit about making things private. Implicit public is alright. Open to revisiting in the future and being more intentional if we feel its valuable.

PR #146, opened 2020-07-08
VSCODE-103: partial playground execution

alenakhineika:
## Description

- When a user selects line/lines from a playground show code lens as a way to start partial playground execution.
- A user can also use a play button on the top right-hand side corner of the editor for a partial playground execution. If nothing is selected the whole playground will be executed.
- Support two commands in command palette: `MongoDB: Run Selection From Playground` and `MongoDB: Run All From Playground`
- If `MongoDB: Run Selected Lines From Playground` is called when noting is selected show a warning: "Please select one or more lines in the playground".
- Do not show code lens if a user selected just part of the line because there is a big chance it can not be properly executed or a user selects eg a field name to copy-paste to another place in code.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/editors/playgroundController.ts
```diff
@@ -76,6 +85,34 @@ export default class PlaygroundController {
         log.info('Active editor path', editor.document.uri?.path);
       }
     });
+
+    vscode.window.onDidChangeTextEditorSelection((editor) => {
+      this._selection = editor.selections
```
Anemy, line 90: Are the selections always in order here? If a user selects a few lines lower in a file and then `alt+click` selects a few more lines higher up in the code are they in chronological order?

src/editors/playgroundController.ts
```diff
@@ -76,6 +85,34 @@ export default class PlaygroundController {
...
+
+  public showCodeLensForSelection(item: vscode.Range): void {
+    const selectedText = this.getSelectedText(item).trim();
+    const firstSelectedLine =
+      this._activeTextEditor?.document.lineAt(item.start.line).text.trim() ||
+      '';
+
+    if (
+      selectedText.length > 0 &&
+      selectedText.length >= firstSelectedLine.length
```
Anemy, line 110: One small change we could do for usability, is making this case show a code lens:
Screen Shot 2020-07-09 at 1 05 31 PM
I think if we change this conditional to see if the first selected line selection is empty might cover it.

src/editors/partialExecutionCodeLensProvider.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  public resolveCodeLens?(codeLens: vscode.CodeLens): vscode.CodeLens {
+    const message = `► Run Selection From Playground`;
+
+    codeLens.command = {
+      title: message,
+      command: 'mdb.runSelectedPlaygroundBlocks',
+      arguments: [message]
+    };
```
Anemy, line 42: After a selection is run, the codelens disappears. Do we think users would expect it to remain? I think I was looking for it. Not sure if that's something we can easily change.

src/editors/playgroundController.ts
alenakhineika: Very nice catch! 👏

alenakhineika:
@Anemy updated according to your comment! very nice comments actually :) about making sure that partial playground is being run on a proper db, probably we want to make sure that any playground is being run on a proper db, so we could add a db name to the confirmation model something like: "Are you sure you want to run this playground against `host:port/db`? This confirmation can be disabled in the extension settings." We can discuss it and add as a separate PR.

PR #154, opened 2020-07-23
VSCODE-97: Connect with code lenses

alenakhineika:
## Description

Use code lenses to connect to the new or saved connection:
- If a user is not connected show code lenses with a disconnected message. The code lenses' message is a link that opens a command palette with a list of saved connections started with the 'Add new connection' item.
- Clicking on the 'Add new connection' item opens a quick input to enter a new connection string.
- Clicking on any other connection item starts the connection process to the selected connection.
- After connecting the code lenses message shows the current connection host and port.
- If the user is connected and click on the code lenses again the same connection will be shown to allow the user select another connection or create a new connection.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -783,7 +783,7 @@ export default class ConnectionController {
     this._disconnecting = disconnecting;
   }
 
-  private getСonnectionQuickPicks(): any[] {
+  public getСonnectionQuickPicks(): any[] {
     if (!this._connections) {
       return [
         {
@@ -803,22 +803,16 @@ export default class ConnectionController {
         }
       },
       ...Object.values(this._connections)
-        .map((item) => ({
+        .sort((connectionA: any, connectionB: any) =>
+          (connectionA.name || '').localeCompare(connectionB.name || '')
+        )
+        .map((item: any) => ({
           label: item.name,
           data: {
             type: NewConnectionType.SAVED_CONNECTION,
             connectionId: item.id
           }
         }))
-        .sort((collectionA: any, collectionB: any) => {
-          if (collectionA.name < collectionB.name) {
-            return -1;
-          }
-          if (collectionA.name > collectionB.name) {
-            return 1;
-          }
-          return 0;
-        })
     ];
   }
 
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -37,6 +37,11 @@ export type SavedConnectionInformation = {
 // A loaded connection contains connection information.
 export type LoadedConnection = SavedConnection & SavedConnectionInformation;
 
+export enum NewConnectionType {
+  NEW_CONNECTION = 'NEW_CONNECTION',
+  SAVED_CONNECTION = 'SAVED_CONNECTION'
+}
+
 export default class ConnectionController {
   // This is a map of connection ids to their configurations.
   // These connections can be saved on the session (runtime),
@@ -586,13 +591,13 @@ export default class ConnectionController {
     const connectionNameToRemove:
       | string
       | undefined = await vscode.window.showQuickPick(
-        connectionIds.map(
-          (id, index) => `${index + 1}: ${this._connections[id].name}`
-        ),
-        {
-          placeHolder: 'Choose a connection to remove...'
-        }
-      );
+      connectionIds.map(
+        (id, index) => `${index + 1}: ${this._connections[id].name}`
+      ),
+      {
+        placeHolder: 'Choose a connection to remove...'
+      }
+    );
 
     if (!connectionNameToRemove) {
       return Promise.resolve(false);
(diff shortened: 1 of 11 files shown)


src/editors/playgroundController.ts
```diff
@@ -171,13 +176,67 @@ export default class PlaygroundController {
...
+      }
+
+      if (
+        selectedQuickPickItem.data.type === NewConnectionType.NEW_CONNECTION
+      ) {
+        return this._connectionController.connectWithURI();
+      }
+
+      return this._connectionController.parseNewConnectionAndConnect(
+        selectedQuickPickItem.data.connectionModel
```
Anemy, line 238: Looks like this creates a new connection with each connection. I think when the user clicks on an existing connection we want to connect to that connection and not recreate a saved connection.

Maybe we can use `connectWithConnectionId` on the `connectionController` instead?

```js
return this._connectionController.parseNewConnectionAndConnect(
  selectedQuickPickItem.data.connectionId
);
```

and above we can change the `data` property to `connectionId` instead of `connectionModel`.

src/editors/activeConnectionCodeLensProvider.ts
```diff
@@ -22,27 +22,18 @@ export default class ActiveConnectionCodeLensProvider implements vscode.CodeLens
...
-
-    return this._codeLenses;
-  }
-
-  public resolveCodeLens?(codeLens: vscode.CodeLens): vscode.CodeLens {
-    const name = this._connectionController.getActiveConnectionName();
-    const message = `Currently connected to ${name}`;
+    const message = activeConnection
+      ? `Currently connected to ${this._connectionController.getActiveConnectionName()}. Click here to change connection.`
+      : 'Disconnected. Click here to add connection.';
```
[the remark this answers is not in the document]
alenakhineika: The current message is from design, but I like your option more! I will change it then and later show Claudia to double-check with her. Also, I like the idea of showing 'Connecting...' message, will add it as well 👍

src/editors/playgroundController.ts
```diff
@@ -171,13 +176,67 @@ export default class PlaygroundController {
...
+      );
```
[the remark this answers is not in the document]
alenakhineika: I don't have a strong opinion about it, so I don't mind moving it to the connection controller!

alenakhineika:
@Anemy thanks for such good comments in the review! Pushed updates ✔️

src/connectionController.ts
```diff
@@ -777,4 +782,69 @@ export default class ConnectionController {
...
+            type: NewConnectionType.SAVED_CONNECTION,
+            connectionId: item.id
+          }
+        }))
+        .sort((collectionA: any, collectionB: any) => {
+          if (collectionA.name < collectionB.name) {
+            return -1;
+          }
+          if (collectionA.name > collectionB.name) {
+            return 1;
```
Anemy, line 818: I think this sort might be doing not quite what we expect:
Screen Shot 2020-07-28 at 10 24 31 AM

Looks like we're comparing `name` when the object builds the name into `label`, I think if we use `label` in the comparison it should be good.

src/editors/activeConnectionCodeLensProvider.ts
```diff
@@ -22,27 +22,23 @@ export default class ActiveConnectionCodeLensProvider implements vscode.CodeLens
...
     codeLens.command = {
       title: message,
-      command: "mdb.showActiveConnectionInPlayground",
-      arguments: [message]
+      command: 'mdb.changeActiveConnection',
+      arguments: []
     };
 
-    return codeLens;
+    return [codeLens];
```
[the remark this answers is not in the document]
alenakhineika: Me too! But I wanted to have all data preparation in provideCodeLenses for testing purposes and it worked! :)

src/connectionController.ts
alenakhineika: Interesting. I wanted it to be sorted the same way as in the sidebar. I assumed that it happens in `databaseTreeItem.ts` in the `getChildren` method that sorts by name. Is it some different list? Where we sort sidebar connections in code?

Anemy: Ah looks like collections aren't case insensitive also:
Screen Shot 2020-07-28 at 11 28 11 AM
I think most of the collections I have are lower case and never noticed it!
If you are up for fixing that as well that would be 🔥 we can also do it another time though.
Nice catch

alenakhineika: Np I will fix both places!

alenakhineika: Pushed updates, now connection quick pics and collections are case insensitive.

PR #156, opened 2020-07-29
VSCODE-155: Allow connecting to a new connection when already connecting

Anemy:
VSCODE-155

This PR adds versioning to connecting so that a new connection can be attempted while there's already a connection in progress. Only the most recent connection attempt will successfully open the connection to be used.

This PR also allows users to remove connections while they're currently connecting. I think this PR adds a bit of complexity to how we connect to new connections and disconnect, so there can be some risk of introducing a possible race condition in future, it's something we should be aware of.

Something that happens sometimes when connecting to a connection is the connection string is misformed or the database is currently offline. In these cases we would freeze up for 30s waiting for the connection to resolve. This might confuse a user and is frustrating when it's a small issue like an incorrect port number. I think there's also a chance that the connection never resolves with some ssh connections, I haven't been able to reproduce, but this PR should also address that case. 

many connections

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -55,6 +55,12 @@ export default class ConnectionController {
   _activeConnectionModel: null | ConnectionModelType = null;
   private _currentConnectionId: null | string = null;
 
+  // When we are connecting to a server we save a connection version to
+  // the request. That way if a new connection attempt is made while
+  // the connection is being established, we know we can ignore the
+  // request when it is completed so we don't have two live connections at once.
+  private _connectingVersion = 0;
+
   private _connecting = false;
   private _connectingConnectionId: null | string = null;
   private _disconnecting = false;
@@ -231,7 +237,7 @@ export default class ConnectionController {
   public sendTelemetry(
     newDataService: DataServiceType,
     connectionType: ConnectionTypes
-  ) {
+  ): void {
     // Send metrics to Segment
     this._telemetryController.trackNewConnection(
       newDataService,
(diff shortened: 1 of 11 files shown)


Anemy:
@alenakhineika Updated - now when the connection attempt finishes it returns the overall connection result from the connection controller. So if a successful connection attempt was made which overrode the connection in the connection form it will still display `Successfully connected`. I think we can revisit in the connection form epic and plug more directly into the current state of the connection controller (listen directly to `CONNECTIONS_DID_CHANGE` events and display the overall connection status).

Anemy:
@alenakhineika Updated so that when a connection is attempted from the connection form, the webview controller listens to connection events from the connection controller. When a connection event occurs, like the connection didn't succeed, or it did succeed it will message the connection form with the name of the connection which succeeded. This way if the attempting connection is overridden and the overriding connection succeeds it is shown as the connected connection.
Maybe give it a test and see if it fits the UX you're looking for? Happy to improve more.

alenakhineika:
Yes, the functionality works as a user I would expect, nice! Not a blocker for release, but I would improve UX and styles here a bit more.

Screenshot 2020-08-07 at 10 29 43

Currently, there are two places with user messages: the connect button changes to `Connecting...` state while connecting and the `Successfully connected` line appears above the button. Plus the color of the text is different. I would expect that the styles are matching, the `Connect` button does not change the title, maybe becomes disabled while connecting. And all info about the current connecting state is shown as a green message line above the button.

Anemy:
@alenakhineika How's this? Do you prefer darker text in the message or lighter text color?
button and message styling

PR #161, opened 2020-08-13
VSCODE-138: Quick index creation

Anemy:
VSCODE-138

This PR adds a button and right click action to create a playground with a template to create an index quickly. The PR also adds a refresh indexes right click action. It also adds caching to the expanded state of indexes which should be nice (previously they'd collapse if a parent folder was expanded/collapsed).

How does the ensureIndex/createIndex playground template look? Went with `ensureIndex` instead of `createIndex`

lower res low fps quick create
*Using older template in gif*

### DIFF
--- a/src/explorer/indexTreeItem.ts
+++ b/src/explorer/indexTreeItem.ts
@@ -11,10 +11,10 @@ export enum IndexKeyType {
   HASHED = 'hashed',
   GEO = '2d', // flat, cartesian geometry
   GEOSPHERE = '2dsphere', // index assuming a spherical geometry
-  GEOHAYSTACK = 'geoHaystack'
+  GEOHAYSTACK = 'geoHaystack',
 }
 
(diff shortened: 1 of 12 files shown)


src/explorer/indexTreeItem.ts
```diff
@@ -10,10 +11,10 @@ export enum IndexKeyType {
   HASHED = 'hashed',
   GEO = '2d', // flat, cartesian geometry
   GEOSPHERE = '2dsphere', // index assuming a spherical geometry
-  GEOHAYSTACK = 'geoHaystack'
+  GEOHAYSTACK = 'geoHaystack',
```
alenakhineika, line 14: 😶

alenakhineika (approved):
LGTM! I think in standup everyone was ok with having ensureIndex, so we can keep it.

PR #171, opened 2020-09-23
Adjusted linkId for Atlas CTA in Help panel

mmarcon:
## Description

The `linkId` of the Atlas CTA in the help panel to distinguish it from the CTA in the connect screen in Amplitude.

### Checklist

## Motivation and Context
- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/explorer/helpTree.ts
+++ b/src/explorer/helpTree.ts
@@ -65,35 +65,35 @@ export default class HelpTree
       const whatsNew = new HelpLinkTreeItem(
         'What\'s New',
         'https://github.com/mongodb-js/vscode/blob/master/CHANGELOG.md',
-        'whatsNew',
+        'whatsNewHelpTree',
         'megaphone'
       );
 
       const extensionDocs = new HelpLinkTreeItem(
         'Extension Documentation',
         'https://docs.mongodb.com/mongodb-vscode/',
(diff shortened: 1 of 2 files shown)


alenakhineika:
Shouldn't we change this place as well?

https://github.com/mongodb-js/vscode/blob/master/src/views/webview-app/components/help-panel/help-panel.tsx#L160

mmarcon:
@alenakhineika If I understand correctly what that file does, it's the one that shows this:
image

If that's the case then no, we should not change the id in that place too because the reason for this tweak is exactly to distinguish the 2 links.

Anemy (commented):
lgtm. the linkIds on the other `HelpTree` events are good right? Do we want to postfix all of the link ids in the help panel with `HelpTree`? 
Ex:
`whatsNew` -> `whatsNewHelpTree`
Might help distinguishing events from the new overview page links.

mmarcon:
Good point. I'll change them all.

Anemy (approved):

mmarcon:
Never mind, I just realized that we also have a `screen` property set to `helpPanel`. That's all I need. I will close this PR.

PR #172, opened 2020-09-24
VSCODE-172: Bump mongosh version to 0.4.0

Anemy:
0.0.1-alpha.19 -> ^0.4.0

Using the new future [mongosh](https://github.com/mongodb-js/mongosh).

Fixed a couple lints warns and updated vscode stubs with new variables in the interface.
Added acorn plugins to the webpack externals since webpack was using its version of acorn and overriding the ones the shell api is pulling in resulting in an error. *Didn't investigate deeper if its something we could update in mongosh, will do that a little now.
Here's where the acorn plugins are pulled in: https://github.com/mongodb-js/mongosh/blob/master/packages/async-rewriter/src/await.js#L9


addaleax (approved):

Anemy:
hmmm in the built .vsix the built language server is not resolving the acorn plugins. Looking into and fixing before merging. Probably will wait to release till monday then.

Anemy:
Looks like it something we need to change on mongosh to make sure we use the acorn dependencies correctly, we can pull in a new version of mongosh once we merge in this pr and release: https://github.com/mongodb-js/mongosh/pull/344

PR #173, opened 2020-09-29
VSCODE-123: Use asWebviewUri to load our bundled webview js file

Anemy:
VSCODE-123

Instead of loading the connect form's webview js file as a `vscode-resource:` URI we now use the URI created from `asWebviewUri`.

Addresses: https://github.com/mongodb-js/vscode/issues/102

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,445 +5,253 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.5.5",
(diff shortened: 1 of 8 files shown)


Anemy:
Looks like this PR also updated the package-lock.json which will pull in the acorn plugins which break the build. Tests are expected to fail - we can merge this in after we've pinned those deps in mongosh https://github.com/mongodb-js/mongosh/pull/344

PR #176, opened 2020-10-05
Update connection uri input to not hide on focus out

Anemy:
Improves the connect with uri experience. Somehow totally missed this option when building it xD
https://code.visualstudio.com/api/references/vscode-api#InputBoxOptions

keep focus prompt


addaleax (approved):

mmarcon:
*Very* nice little improvement! It does happen sometimes that I go out of the palette to copy-paste the string from somewhere else.

PR #178, opened 2020-10-06
VSCODE-98: Add overview page

Anemy:
VSCODE-98 This PR adds an overview page which can be accessed with the command palette (`Cmd + Shift + P`) with the command name `MongoDB: Open Overview Page`. 

This area of the code base could use a good amount of tests, I started to add in jest and updating tests in this branch but it ended up being a lot more work/diff than should go in a PR with some things already, so it'll come in a new PR. Also might end up moving back to mocha & enzyme for the react tests since jest really doesn't play well with running in a vscode instance and sharing the global extension & vscode attributes. I think it would be best to only use 1 test runner, rather than testing w/ mocha and then the react tests in jest, but more on that will come in that PR.

Something we've thought a bit about going into this work is how we can improve how we're doing styles in this webview. Initially I was thinking maybe we use css in js and use an approach like [leafygreen](https://github.com/mongodb/leafygreen-ui), but decided less was simple enough and just stuck with it. I think theming maybe isn't as declarative, but I don't see too much reason to change.

The Atlas banner at the bottom will change in VSCODE-169, the connection status comes in VSCODE-164, the resources panel in VSCODE-168, the connection form in VSCODE-173, and lastly we'll update the navigation the user uses to get to this overview page in VSCODE-170

Open Qs:
- How is the text? The mocks have the message in the header as `Navigate your databases and collections, use playgrounds for queries and aggregations` this PR changes it a bit: `Navigate your databases and collections, use playgrounds for exploring and transforming your data` - Which do ya'll prefer - or have a suggestion?

*Here's how it looks*
___
overview:home page dark theme
Screen Shot 2020-10-06 at 4 10 51 PM
Screen Shot 2020-10-06 at 4 10 32 PM


src/views/webview-app/components/connect-helper/connect-helper.tsx
```diff
@@ -0,0 +1,102 @@
...
+  ActionTypes,
+  LinkClickedAction,
+  OpenConnectionStringInputAction,
+  SetCurrentViewAction
+} from '../../store/actions';
+import { WEBVIEW_VIEWS } from '../../extension-app-message-constants';
+
+const styles = require('./connect-helper.less');
+
+type dispatchProps = {
```
mcasimir, line 14: extremely silly question, isn't the convention for types to be like classes 'DispatchProps'?

src/views/webview-app/components/connect-helper/connect-helper.tsx
```diff
@@ -0,0 +1,102 @@
...
+                styles['connect-helper-connection-form-button']
+              )}
+              onClick={() => onOpenConnectionFrom()}
+            >
+              Open form
+            </button>
+          </div>
+        </div>
+        <div className={styles['connect-helper-message']}>
+          <strong>{getOSCommandShortcutName()} + Shift + P</strong> for all MongoDB Command Palette options
```
mcasimir, line 80: isn't there an api to just get the keybinding for the palette?

src/views/webview-app/components/overview-page/overview-header.tsx
```diff
@@ -0,0 +1,57 @@
...
+type props = dispatchProps;
+
+class OverviewHeader extends React.Component<props> {
+  onLinkClicked = (screen: string, linkId: string): void => {
+    this.props.onLinkClicked(screen, linkId);
+  };
+
+  renderMongoLogo(): React.ReactNode {
+    return (
+      <svg width="188" height="50" viewBox="0 0 188 50" fill="none" xmlns="http://www.w3.org/2000/svg">
```
mcasimir, line 20: Could we put this in a separate file, either with a 👿  webpack 👿  loader, or just as a separate component for the logo?

mcasimir: Just noticed that we need classes there, so probably not a webpack loader :)

mcasimir (approved):
Looks great!

src/views/webview-app/components/connect-helper/connect-helper.tsx
Anemy: Ah it is hah. Updated ty.

src/views/webview-app/components/connect-helper/connect-helper.tsx
Anemy: I thought so, but I couldn't find the hotkeys on the configuration object https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration  `vscode.workspace.getConfiguration()` so I just went with showing the default :/

src/views/webview-app/components/overview-page/overview-header.tsx
Anemy: good call, updated

Anemy:
Merging this in - since we're working on other components in the overview page in the next few tickets we can keep iterating on anything from this if we want to change it.

PR #179, opened 2020-10-08
VSCODE-98: Add webview tests

Anemy:
VSCODE-98 This PR adds tests for our webview which shows the overview and connection form pages. Since this webview is running in its own context without the main extension's scope we run our tests using a different test runner (jest). This PR adds jest, and enzyme for react testing. In our CI we now parse for two xml test outputs `test-extension-results.xml` and `test-webview-results.xml`.
Bumped typescript to make the typings not have conflicts between jest and mocha. Fixed our tree items which were using accessors for tooltip, description, and iconPath in vscode which are now not part of the `vscode.TreeItem` interface. Could move that into another pr if it helps to look at.

Also did VSCODE-176 added a tooltip for schema fields showing their type, since it was a quick tooltip fix:
Screen Shot 2020-10-08 at 3 16 21 PM

Example of jest test running output:
Screen Shot 2020-10-08 at 3 11 53 PM


src/explorer/documentListTreeItem.ts
```diff
@@ -55,6 +55,27 @@ export const formatDocCount = (count: number): string => {
   return `${numeral(count).format('0a')}`.toUpperCase();
 };
 
+function getIconPath():
+    | string
+    | vscode.Uri
+    | { light: string | vscode.Uri; dark: string | vscode.Uri } {
```
addaleax, line 61: I know this is from the previous code and occurs in a couple places, but we only ever use the last type signature here, and the other ones could be dropped, right?

src/views/webview-app/components/connect-form/authentication/kerberos.tsx
```diff
@@ -150,15 +150,14 @@ class Kerberos extends React.Component<props> {
...
+          {/* <Toggle
             className={styles['form-toggle']}
             name="kerberosCanonicalizeHostname"
             onChange={this.onCnameToggle}
             checked={kerberosCanonicalizeHostname || false}
             size="small"
-            variant="default"
             disabled={false}
-          />
+          /> */}
```
addaleax, line 160: Is this temporary? I’m obviously not having the full context here, but it’s not obvious from the diff why this is being commented out :)

Anemy: ah ty meant to mention that before opening the pr hah

Anemy: Added it back in, moved PR out of WIP. Had to add the `esModuleInterop` flag in tsconfig to get the leafygreen import working right, cleaned up some of our other imports for that to work.

src/explorer/documentListTreeItem.ts
Anemy: Good call, I think maybe we can clean these up a bit more in future too. Pushed update.

addaleax (approved):
LGTM :)

PR #183, opened 2020-10-12
VSCODE-164: Display connection status in overview

Anemy:
VSCODE-164 *edit and also VSCODE-166

This PR adds a connection status to the overview page.

The webview requests the connection status from the extension every 1000 ms (1s). I went with polling because the overview page can be hidden and not receiving messages from the extension when it is not one of the active editors. I think polling is a bit more fail safe, and ensures we're showing the right status in those cases where the overview page has been reopened. I think because it's a lightweight call 1000ms is an alright interval for requesting the status - what do ya'll think? We also could send the connection status on connection change events, but I'm not sure it's needed because we request the status so often.
Also includes a bit of cleanup on the shared message types that our webview and extension use when passing messages to each other.

The mocks include a button to edit the connection name when connected, that'll come in a future PR.

connection status overview page
*light mode*
Screen Shot 2020-10-12 at 5 17 19 PM


alenakhineika:
New overview page! Niceee 😍

I tested this PR locally and sometimes the status id being stuck at the `Disconnecting...` status and throw the following error:

`Error running command mdb.disconnectFromConnectionTreeItem: Running the contributed command: 'mdb.disconnectFromConnectionTreeItem' failed.. This is likely caused by the extension that contributes mdb.disconnectFromConnectionTreeItem.`

Steps to reproduce:

- Connect to localhost from saved connections
- Open the overview page from the command palette
- Click `Connect with connection string`
- Do not fill in anything
- Click outside the input in order to the connection string input lose a focus
- Click Esc to close the connection string input
- Click `Connect with connection settings`
- Close the connect page
- Open the connect page
- Disconnect using the contextual menu
- Click on the same connection to connect back

Maybe there is a shorted way 😀 but this is how I can get this error each time a repeat steps.

I also have a concern about the size of the `All set. Ready to start? Create a playground` text. It looks huge in compressing with other text blocks on the page.

And one more small thing, do we want to have a dot after `Not Connected`? Other sentences don't have it at the end.

alenakhineika:
> We also could send the connection status on connection change events, but I'm not sure it's needed because we request the status so often

Do we really need to ping status each second? Won't be this enough to update the active status only when the `connection change` event occurs?

Anemy:
Ah - nice find - looks like it's not cleaning up a listener properly when the webview is closed. It sets up a listener when it messages the extension to connect to a connection string or connection. It's listening for connection events which it then tries to send to the webview. But when the webview is closed - like in your repro, messaging the webview fails because it no longer exists. I'll push up a fix and ping when it's up. I think it's an existing bug with the connect form as well.

I was thinking pinging from the webview ensures messages are only sent while the webview is open - so that in instances where it's closed it isn't the extension's responsibility to message it - kind of related to this bug lol.

src/views/webview-app/store/store.ts
```diff
@@ -5,16 +5,23 @@ import ConnectionModel, {
...
   WEBVIEW_VIEWS
 } from '../extension-app-message-constants';
 
+interface VSCodeApi {
+  postMessage: (message: MESSAGE_FROM_WEBVIEW_TO_EXTENSION) => void;
+}
+
 // eslint-disable-next-line no-var
-declare var acquireVsCodeApi: any;
+declare var acquireVsCodeApi: () => VSCodeApi;
```
alenakhineika, line 20: Won't window. acquireVsCodeApi work here the same way? To not mix vars with consts and lets.

Anemy:
Pushed up a fix for that command failed bug and removed that old `var` 👍 

For the text styling and the `Not connected.` text - does it sound to sync with Claudia when she's back? Maybe we merge this in now and smaller commit when she's back and we can chat on it?

alenakhineika:
Max wants to release the console.log feature, and if we want to ship this pr along with it, I have concerns about these moments that look a bit like a broken css. I also think that creating a playground from the overview page has the same value as connecting with a connection string or connection settings, therefore the size should be pretty much the same. I asked Max about it, lets wait for his feedback.

alenakhineika:
Checked the last version of PR and the disconnecting... bug fixed! Nice 👏

PR #185, opened 2020-10-12
VSCODE-181: Move extension commands into a single file in `src/commands`

Anemy:
VSCODE-181

Opening this a bit more as a discussion - do ya'll think this is a good change - or not quite useful/overly verbose?
I think the good thing this change brings is reducing the chance for string miss-spellings if we have commands being called in multiple places.
How's the command variable naming? `EXTENSION_COMMANDS` and commands prefixed with `MDB_` ?


alenakhineika:
Doesn't vscode throw an error during building process if we try to use the unexisting command (eg misspelled)?

Anemy:
We get an error when we try to register a command that isn't in the `package.json`, but we don't get an error until runtime when we try to call a command that doesn't exist in package.json. Hopefully our tests would catch all those but I think this PR makes it a bit safer.

alenakhineika (approved):

PR #190, opened 2020-10-20
VSCODE-190: Move connect form into modal

Anemy:
VSCODE-190

This PR moves the connect form into a modal on the overview page.
Much of the diff is from moving the connect form styling out of a single place, and breaking the form components of the connection form into a more generalized place.

This does not update the styles of the connect form. That's coming in VSCODE-191 (And will have some more tests 😄 )

connect modal


mcasimir (approved):
👍

PR #193, opened 2020-10-26
Atlas sign-in/up uses a localhost redirect

mmarcon:
This constitutes the prep work to eventually introduce a sign-in/sign-up functionality for Atlas.
By using a short-lived local server, we can redirect the user to the right Atlas sign-in/up page and wait for an auth redirect in VS Code.

### Checklist

## Motivation and Context
- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Open Questions
~I have a test that is currently commented out that I don't know how to make pass.~

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/explorer/helpTree.ts
+++ b/src/explorer/helpTree.ts
@@ -30,7 +30,7 @@ export class HelpLinkTreeItem extends vscode.TreeItem {
   url: string;
   useRedirect: boolean;
 
-  constructor(title: string, url: string, linkId: string, iconName?: string, useRedirect = false,) {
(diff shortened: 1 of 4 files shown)


Anemy (approved):
Sweet! Tried it out on local and works real smooth. I added a commit that adds a try catch to the openLink method - incase the server fails. It then defaults back to regular link opening on error. I'm thinking a user might double click a link or maybe even have something running on that port which could make something funny happen. If the commit looks good to you then lgtm!

PR #197, opened 2020-11-02
VSCODE-191: Add tabs to connect form

Anemy:
VSCODE-191

This PR adds tabs to the connection form, a few smaller ux improvements, and some more general connect form tests.

This PR also adds new ux improvements to the connection form functionality:
- Hide the replica set name input field in the advanced tab when srv record is true.
- The username field is now optional with x509 connections. (It might not even be used/required since mongodb 3.4). Manually tested with a working x509 connection without setting the username and it works.
- Show an error when the user does have SSL configured and they're using x509 authentication. VSCODE-195 
- Adds `.pem` file hint to the SSL file input descriptions.

Since this PR was already getting a bit big I moved out the changes to the select items (using toggle buttons instead of a select dropdown for items like authentication method and ssl method) to VSCODE-197

A bit of a design choice, I decided to not to use leafygreen's [modal](https://github.com/mongodb/leafygreen-ui/tree/master/packages/modal) and [tabs](https://github.com/mongodb/leafygreen-ui/tree/master/packages/tabs) components in favor of having our own. This is mostly because we want to use the vscode user's color themes, we'd have to be overriding every color in the components in their styles. I think it would be a quick swap if we ever decide its worth using these components.

___

*Connecting*
connect form new styles

*x509 error message when SSL method is not set*
Screen Shot 2020-11-02 at 1 33 06 PM

*Connection form tab active*
Screen Shot 2020-11-02 at 1 39 39 PM

*Connect form tab focus*
Screen Shot 2020-11-02 at 1 39 48 PM


src/views/webview-app/components/connect-form/ssl-tab/ssl-server-client-validation.tsx
```diff
@@ -64,15 +62,15 @@ class SSLServerClientValidation extends React.Component<props> {
         className={classnames(styles['form-group'])}
       >
         <FileInputButton
-          label="Certificate Authority"
+          label="Certificate Authority (.pem)"
```
alenakhineika, line 65: 👍

alenakhineika (approved):
Works very well! I like having the connect form as a modal over the overview page. I think people will like that they are not being redirected to some other window.

Small wish: would be nice to have a possibility to reset the connection setting. Currently, if i did some changes at different tabs i can't reset them to default values, even if i close the modal and open to again. I think advanced users can easily find proper values for inputs, but for others it can be challenging to do so if they were experimenting with values and messed up something.

src/views/webview-app/components/connect-form/ssl-tab/ssl-server-client-validation.tsx
Anemy: Hmm good thought on resetting the fields. We can chat w/ Claudia on that.

Issue #196, opened 2020-10-30
Bug Report: vscode "run all" command no longer shows up when searching for vscode commands


benatshippabo:
#### Current Behavior

Regression on vscode command to run all mongodb playgrounds. Before you could `cmd + shift + p` and then search to run all mongodb playgrounds, but now it's no longer present and we only have the play button.

#### Environment

```
Version: 1.50.1
Commit: d2e414d9e4239a252d1ab117bd7067f125afd80a
Date: 2020-10-13T14:53:05.704Z (2 wks ago)
Electron: 9.2.1
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Darwin x64 19.6.0
mongodb.mongodb-vscode: v0.2.1
```

#### Possible Solution

Please add the vscode command back so it is searchable again. 🙏 

#### Additional Context/Screenshots

![image](https://user-images.githubusercontent.com/66273043/97645876-56e8ec80-1a0b-11eb-9295-614f5c2258dc.png)

Anemy:
Thanks for the report - definitely a bug. We'll wrap a fix for this into the current work around improving playgrounds and get it out soon! I'll ping here when it's live.

mmarcon:
The fix will be in the next release.

PR #198, opened 2020-11-06
VSCODE-177: display playground output in editor

alenakhineika:
## Description

- Move playground results from the output channel to the virtual document opened in a split-view mode.
- The results of a playground should have syntax highlighting if they represent a JSON document.
- The results that are not Ja SON document should be printed as plain text with a single color.
- Users can open as many playgrounds they want, but results should be always printed to the same Results editor.
- The Results editor can be drag and dropped to any side of a split-view mode and should keep receiving results from any opened playground. 
- The Results file can be saved on a disc and in this case the current Results file stops acting as a Results editor and the new Results editor will be open with the next successful playground execution. It will allow users to freeze some results and compare them with each other.
- Keep printing debug information (eg console.log) in the output channel.

Also, this PR includes a small package.json fix to address the following issue: https://github.com/mongodb-js/vscode/issues/196

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/editors/playgroundController.ts
+++ b/src/editors/playgroundController.ts
@@ -344,7 +344,6 @@ export default class PlaygroundController {
       vscode.workspace.openTextDocument(playgroundResultURI).then((doc) => {
         vscode.window.showTextDocument(doc, { preview: true, viewColumn })
         .then(editor => {
-          this._playgroundResultTextEditor = editor;
           editor.edit(editBuilder => {
             let fullRange = new vscode.Range(new vscode.Position(0, 0), editor.document.positionAt(editor.document.getText().length));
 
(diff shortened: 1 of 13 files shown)


src/editors/playgroundController.ts
```diff
@@ -303,28 +346,49 @@ export default class PlaygroundController {
...
 
-      const result: ExecuteAllResult = await this.evaluateWithCancelModal();
+      const evaluateResponse: ExecuteAllResult = await this.evaluateWithCancelModal();
 
-      if (!result) {
-        this._outputChannel.clear();
+      this._outputChannel.clear();
+      if (evaluateResponse.outputLines) {
+        for (const line of evaluateResponse.outputLines)
+          this._outputChannel.appendLine(line.content);
```
addaleax, line 354: So `console.log()` etc. go to the output channel now but not the actual result? That seems like a very nice way to separate the two :+1:

addaleax (approved):

src/editors/playgroundController.ts
alenakhineika: We were considering different approaches during design sync, but everyone liked this one the most :)

Issue #186, opened 2020-10-13
Open mongo shell fails for git bash shell


ValYouW:
Hi,

When I try to launch the mongo shell it fails with this error:
```
The terminal process failed to launch: A native exception occurred during launch (File not found: ).
```

My default terminal shell is Git bash (`C:\Program Files\Git\bin\bash.exe`)
When I change the default terminal shell to PowerShell all works good.
When I manually open the terminal (git bash) and type `mongo` the shell starts (so it is in the PATH).

Any ideas?

Windows 10
VSCode version: 1.49.3
MongoDB extension version: 0.2.0

PR #201, opened 2020-11-10
VSCODE-200: Open mongodb shell with git bash

Anemy:
VSCODE-200

Fixes https://github.com/mongodb-js/vscode/issues/186

This PR adds support for launching MongoDB shell when the user has their default shell set to git bash.
It also fixes an issue with loading files with spaces in them.

The way we do the environment passing for these different shell environments is a bit manual and hacky. I've looked around for a bit for a node package which provides support for building the shell command for various shells/environments but haven't found one. Do ya'll know of one? I think this works for now but it would be nice to have a quick catch all to prevent any lack of support for other environments.

Manually tested with SSL certs on windows (with a cert file with a space in it which is how I came across that other bug xD).

### DIFF
--- a/src/commands/launchMongoShell.ts
+++ b/src/commands/launchMongoShell.ts
@@ -13,29 +13,52 @@ function isSslConnection(activeConnectionModel: any): boolean {
 }
 
 function getSslOptions(driverOptions: any): string[] {
-  const mdbSslOptions = [ '--ssl' ];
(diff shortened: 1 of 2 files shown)


rose-m (approved):
Code change LGTM, nice 👍 

Didn't test as I've not setup the Windows environment yet...

PR #204, opened 2020-11-12
chore: replace some hardcoded styling colors with variables

Anemy:
This is just a bit of less cleanup to remove some of the duplicated hard coded color values. We could use more variables in general, and have a more defined palette, but I think it's fine for now.

*pass of webview functionality to show nothing has changed:*
styling functionality pass


src/views/webview-app/components/connect-form/connection-form.tsx
```diff
@@ -33,27 +33,6 @@ type DispatchProps = {
...
-        <div className={styles['connection-message-container']}>
-          <div className={styles['connection-message-container-success']}>
-            <div
-              className={styles['connection-message']}
-            >{connectionMessage}</div>
-          </div>
-        </div>
-      );
-    }
-  }
```
Anemy, line 55: This was unused. Would probably be good to have some stricter typescript rules 🤔 although it probably wouldn't catch cases like this since it's a function in an exported class.

gribnoysup (approved):

alenakhineika (approved):
LGTM!

Issue #169, opened 2020-09-15
Adding connections using mongodb:// connection strings is broken


azec-pdx:
#### Current Behavior

Let's assume I start by creating new MongoDB Connection to my Replica with these parameters...

**Hostname:** `2.microservicedomain.mongo.us-east-1.my-aws-acct.local`
**Port:** `27017`
**Authentication:** `SCRAM-SHA-256`
**Username:** `myUsername`
**Password:** `C;Ib86n5b8{AnExew[TU%XZy,)E6G!dk`
**Authentication Database:** `admin`
**Replica Set Name:** `microservicedomain`

When I create connection by filling out form in VS Code, it connects fine.
If I now copy connection string from this connection (`Right click on connection` --> `Copy Connection String`) , it produces following connection string: `mongodb://myUsername:C%3BIb86n5b8%7BAnExew%5BTU%25XZy%2C)E6G!dk@2.microservicedomain.mongo.us-east-1.my-aws-acct.local:27017/?authSource=admin&authMechanism=SCRAM-SHA-256&replicaSet=microservicedomain&readPreference=primary&ssl=false`

This connection string is problematic in multiple ways:

* Password is not compliant with [rfc3986](https://tools.ietf.org/html/rfc3986#section-2.1) spec for percent-encoding that is used to sanitize MongoDB connection string usernames & passwords
* If I delete this connection added through UI and use this generated connection string to re-create same MongoDB connection, it fails. Therefore, connection string is not idempotent if I can not re-use it to create connection.
* If I try to feed this string to [mongosh](https://www.mongodb.com/blog/post/introducing-the-new-shell) it will not connect, again because Connection String is incorrect.

Additionally to this, creating the new Connection with valid Connection String which includes percent-encoded password is not working:

* Assume that I take valid Connection String where password was already correctly percent-encoded: `mongodb://myUsername:C%3BIb86n5b8%7BAnExew%5BTU%25XZy%2C%29E6G%21dk@2.microservicedomain.mongo.us-east-1.my-aws-acct.local:27017/?authSource=admin&authMechanism=SCRAM-SHA-256&replicaSet=microservicedomain&readPreference=primary&ssl=false` and feed it to VS Code in attempt to create New Connection from Connection String. **It fails to connect.** While this connection string does work with `mongosh` and classic `mongo shell` , it does not in VS Code plugin. 
* Assume that I take Connection String where password was NOT percent-encoded:
[...]

PR #207, opened 2020-11-17
VSCODE-198: Use RFC 3986 encoding

alenakhineika:
## Description

To be more stringent in adhering to RFC 3986, replace !, ', (, ), and * with a corresponding character code.

Related issue: https://github.com/mongodb-js/vscode/issues/169

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [x] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


Anemy (approved):
Nice! Tried with a few different configurations, including the one in the issue and they're working well. (can launch MongoDB Shell as well).

Unrelated, but it looks like we're keeping app name in the connection string when it's copied. Created VSCODE-207 for that.

rose-m (approved):
Nice! Also tested it and worked 👍 

One thing I noticed: connections that already existed were not updated. I had a connection in there with an `!` in the password _before_ I did the update to this PR version. _After_ the update when I copied the connection string it still had the pure `!` in it and it was not encoded. Is that intended? 🤔

alenakhineika:
@rose-m good catch! looks like derived properties stored as well to disk and not being calculated on the fly.

alenakhineika:
@rose-m pushed an update, please test again.

rose-m:
Yes, it works now 🎉

PR #208, opened 2020-11-17
feat(VSCODE-168): Add resources panel to overview page

Anemy:
VSCODE-168

This PR adds a resources panel to the overview page.

If possible, when reviewing, please take a look at the urls just to be sure I haven't missed something and they all make sense.

@mmarcon , does the telemetry screen id `overviewResourcesPanel` work for these links? How are the linkIds? https://github.com/mongodb-js/vscode/pull/208/files#diff-6e516781f35b2dd2aefe6b054aff9448af2c9434e6ed261f87eee4b401059d2bR19

*Gif and screenshot are up to date except `MongoDB Cloud` is removed from the `Contribute` list.
resources panel
*light mode*
Screen Shot 2020-11-18 at 12 34 22 PM
*some other theme*
Screen Shot 2020-11-18 at 12 34 56 PM


src/views/webview-app/components/resources-panel/resources-panel.tsx
```diff
@@ -0,0 +1,205 @@
...
+  ToggleShowResourcesPanelAction
+} from '../../store/actions';
+
+const styles = require('./resources-panel.less');
+
+const ResourceLinks = [{
+  title: 'Product overview',
+  description: 'Get an overview on MongoDB',
+  linkId: 'productOverview',
+  url: 'https://www.mongodb.com/'
```
Anemy, line 20: Not sure where this one should link to.

Anemy: Ah got it. Will link to https://docs.mongodb.com/

mmarcon:
I think we can remove the `MongoDB Cloud` link, it does not really belong to the `Contribute` section.
Other than that, LGTM.

@Sgrinfy are you happy with how this looks?

src/views/webview-app/components/resources-panel/resources-panel.less
```diff
@@ -0,0 +1,124 @@
...
+  z-index: -1;
+  background-color: rgba(0, 0, 0, 0.5);
+  animation-duration: 250ms;
+  animation-direction: forwards;
+  animation-name: backgroundDarkenAnimation;
+}
+
+@keyframes backgroundDarkenAnimation {
+  from {
+    width: rgba(0, 0, 0, 0.1);
```
rose-m, line 26:
```suggestion
    background-color: rgba(0, 0, 0, 0.1);
```
Seems like a copy and paste error? 😅

src/views/webview-app/components/resources-panel/resources-panel.less
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  to {
+    width: rgba(0, 0, 0, 0.5);
```
rose-m, line 30:
```suggestion
    background-color: rgba(0, 0, 0, 0.5);
```
And same here of course 😄

src/views/webview-app/components/resources-panel/resources-panel.less
```diff
@@ -0,0 +1,124 @@
...
+
+.resources-panel-content {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  width: 420px;
+  padding: 48px 24px;
+  background: var(--vscode-editor-background);
+  width: 420px;
```
rose-m, line 42: nit: The `width` is given twice (s. line 39)

src/views/webview-app/components/resources-panel/resources-panel.less
```diff
@@ -0,0 +1,124 @@
...
+}
+
+.resources-panel-links-container {
+  margin-top: 54px;
+}
+
+.resources-panel-link {
+  margin-top: 8px;
+  padding: 10px 20px;
+  background-color: rgba(50, 50, 50, 0.25);
```
rose-m, line 94: I'd suggest also using a variable here. With my theme (Material light) it looked pretty strange.
image


I suggest we might switch to `var(--vscode-sideBar-background)` as background and then use `var(--vscode-sideBar-foreground)` as the new `color`?

With the sidebar variant it looks like this:
image

rose-m (commented):
Looks good so far! There are some minor improvements we could fix I think 😊

src/views/webview-app/components/resources-panel/resources-panel.less
Anemy: Good call - here's what it looks like now:
Screen Shot 2020-11-18 at 12 34 56 PM
Screen Shot 2020-11-18 at 12 34 22 PM
Screen Shot 2020-11-18 at 12 34 07 PM

Anemy:
@rose-m Pushed up changes based on your feedback - thanks!
Also sent over the .vsix build of this branch to Claudia, if we find some styling updates I'll update here, but if this is already merged/closed by that time, styling tweaks can come in later PRs.

rose-m (approved):
Nice, LGTM 👍

PR #210, opened 2020-11-17
feat(VSCODE-170): Update connect form routes to overview page connect form

Anemy:
VSCODE-170

This PR updates how we show the connect form when the commands to navigate to the connect form are performed. Previously we would show a standalone connect form. Now we show the overview page (and then a user can select the form). If we find users want a direct link to the connection form back, we can always revisit.

Open Q:
Are there any places we want to explicitly link the overview page? We have a command palette command, however we could add a link in the `Help` panel for instance.

connect routing to overview page


rose-m (approved):
Nice, definitely feels nicer having it in this flow as compared to just the form on a blank page 👍 

Regarding your question - I think it would be nice to also have a link in the Help & Feedback section...
image

Anemy:
@rose-m Thanks. Sounds good - mentioned it to Claudia - going to merge this, but can wrap that into the next ticket since it's a two liner with an icon.

Issue #131, opened 2020-06-08
Unable to find document with NumberLong id


arsmn:

Anemy:
Could you describe the action you're doing?
Are you clicking on documents in the tree view with NumberLong ids? I'll give this a test.

arsmn:
Yes when I click on documents with NumberLong id, this error happens

Chronostasys:
I've got the same problem. In my case, I used a .Net Guid as _id field of my user collection, and when I clicked on documents it tells me "Unable to find document".    

--------
![image](https://user-images.githubusercontent.com/49905799/88354530-ceb3d500-cd93-11ea-9beb-55ee3c253606.png)  
document data  
![image](https://user-images.githubusercontent.com/49905799/88354588-f6a33880-cd93-11ea-86ac-1651cc31b810.png)  
![image](https://user-images.githubusercontent.com/49905799/88354598-03279100-cd94-11ea-8d64-2f0f36457a49.png)  
error after click

Anemy:
Hi @arsmn we've released version `0.4.0` which has a fix for showing documents with numberlong ids as well as binary ids. I'm going to close this, feel free to reopen if when you update you still run into this issue. Thanks for reporting the issue and helping us improve the extension!
I think there still might be a slight display bug around number longs since javascript doesn't play too nicely with them, but it should be unrelated to this issue.

arsmn:
Hi @Anemy
Thanks for the update, I checked that and it's fixed.

PR #213, opened 2020-11-20
fix(VSCODE-118): Fix opening documents with binary _id from tree view

Anemy:
VSCODE-118

Previously only object ids and primitives were able to be open.
This fixes part of https://github.com/mongodb-js/vscode/issues/131 however it looks like we can't hand numberLong types over a certain amount, which is a separate issue.

*The first document opened in gif*
Screen Shot 2020-11-20 at 4 42 34 PM

Also found out that Compass doesn't allow editing binary - or numberLong - values except in the json view, however when it's edited in the json view it silently fails. Need to see if a ticket already exists.
opening docs with binary, object id, and number ids

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3621,12 +3621,11 @@
       }
     },
     "bson": {
-      "version": "4.1.0",
-      "resolved": "https://registry.npmjs.org/bson/-/bson-4.1.0.tgz",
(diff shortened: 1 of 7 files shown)


gribnoysup (approved):

donavon:
This didn't really fix the issue of binary UUIDs. Here is a screenshot of v0.3.0

image

Most will at least open, but if I click on the first, I get this. Note that it happens to contain a "+" sign.

image

Anemy:
@donavon Ah - Thanks for trying it out and the good reproduce case.
This PR should cover these cases: https://github.com/mongodb-js/vscode/pull/220
Hopefully we can get it in the next release in the next few weeks.

PR #217, opened 2020-11-30
feat(VSCODE-196): Show connection form error in connection form

Anemy:
VSCODE-196

This PR updates how we show errors in our connection form. Previously we would show descriptive errors in vscode alerts. This updates the behavior to make the error in the connection form the verbose one, and move away from showing verbose vscode notifications, to show the error in the form ui. 

Wrapped a bit of cleanup in too, previously we were storing the connecting connection identifier as a number, now we use a uuid or the connection's id. The connection form now also stores the connecting connection identifier, so it can ignore overridden connection attempt results.

This PR should also handle the case where the user has closed the form window where they are attempting to connect. Previously this would cause a silent background error.

showing error in connection form

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -364,7 +364,7 @@ export default class ConnectionController {
 
           return resolve({
             successfullyConnected: false,
-            connectionErrorMessage: 'connection overriden'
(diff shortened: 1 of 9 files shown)


addaleax (approved):
Looks good to me :)

PR #219, opened 2020-12-01
[DNM] chore(467) rename NodeOptions to MongoClientOptions

aherlihy:
MONGOSH-467 Upgrade the Node driver to 4.0

## Description

Previously we were making our own options interface `NodeOptions` but now that the driver is in typescript we can use their `MongoClientOptions` interface directly. This will need to be merged after the shell is released with the change


addaleax (approved):

alenakhineika:
TODO: merge after `@mongosh/service-provider-server@0.7.0` published.

Anemy:
Merged in https://github.com/mongodb-js/vscode/pull/301 🎆

PR #220, opened 2020-12-04
fix(VSCODE-118): Fix opening binary _id

Anemy:
The solution introduced in https://github.com/mongodb-js/vscode/pull/213 didn't fully cover all the opening of documents with different typed `_id` cases. When we open documents in vscode, in the original solution we sort of hackily added the _id as a ejsoned string on the document's uri in vscode. This doesn't work well because we have to ejson stringify, parse, uri encode, uri decode, and also we're bound by uri constraints like character limits in the query.
This PR updates how we pass a document's `_id` to the [document provider in vscode](https://code.visualstudio.com/api/extension-guides/virtual-documents#:~:text=Document%20providers%20are%20first%20class,they%20are%20part%20of%20vscode.). It adds a store which we then pass a reference to, so that we can avoid any encoding/decoding of the `_id` value.
On the upcoming editing documents work we may revisit how we show these documents so that we can be a bit more generalized and we can accommodate opening documents from playground results.

Tested with binary ids with `+` characters, numberlongs, and some other possible _ids.

opening all sorts of uuids

### DIFF
--- a/src/editors/playgroundController.ts
+++ b/src/editors/playgroundController.ts
@@ -272,7 +272,10 @@ export default class PlaygroundController {
               this._outputChannel.clear();
               this._outputChannel.show(true);
 
-              return resolve(undefined);
+              return resolve({
+                outputLines: undefined,
+                result: undefined
+              });
             });
 
             // Run all playground scripts.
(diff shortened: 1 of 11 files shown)


src/editors/playgroundController.ts
```diff
@@ -286,7 +289,10 @@ export default class PlaygroundController {
         .then(undefined, (error) => {
           log.error('Evaluate playground with cancel modal error', error);
 
-          return resolve(undefined);
```
Anemy, line 289: typescript & lint wasn't happy with this line which makes sense. This is an unrelated change.
I'll make a ticket to look into adding lint checks to this project.

addaleax (approved):
Looks good! :)

rose-m (approved):

PR #224, opened 2020-12-07
feat(connect-form): enable multi hosts in connection form VSCODE-208

Anemy:
VSCODE-208

This PR updates our host input on the `General` tab in the connection form to allow for 3 connection types: `Standalone`, `Replica Set`, and `SRV Record`. The replica set option allows a user to configure a connection to have multiple hosts and ports.

When connecting to a replica set with multiple hosts and ports, we currently create the connection name based on the `hostname` and `port` in the standalone section. The name is generated with `connection-model`'s `instanceId` which doesn't properly handle names from multiple hosts or srv records. That work will happen in https://jira.mongodb.org/browse/COMPASS-4398 (related existing Compass bug) which I'll pull into this next sprint. For now the names will be a little counter intuitive for replica sets, so we should probably not release till that's in.

Mocks: https://www.figma.com/proto/4Fz3us4dY7YGHchhAQsBD2/VS-Code-%7C-Connect-%26-Overview?node-id=1416%3A774&viewport=653%2C-1916%2C1.0270851850509644&scaling=min-zoom 

connecting to replica set with different hosts and ports

Note: `SRV` is renamed to `SRV Record` which isn't in these screenshots.

Screen Shot 2020-12-07 at 10 33 29 AM


src/views/webview-app/components/connect-form/general-tab/general-tab.tsx
```diff
@@ -1,61 +1,17 @@
...
-        <FormGroup id="connection-host-information" separator>
-          <HostInput />
-          {this.renderPort()}
-          <SRVInput />
-        </FormGroup>
+      <React.Fragment>
+        <HostInput />
         <Authentication />
-      </div>
+      </React.Fragment>
```
addaleax, line 12: TIL :)

src/views/webview-app/components/connect-form/general-tab/host/host.tsx
```diff
@@ -0,0 +1,352 @@
...
+      this.props.setReplicaSet('');
+    }
+  };
+
+  onHostnameChanged = (event): void => {
+    this.props.onHostnameChanged(event.target.value);
+  };
+
+  onPortChanged = (event): void => {
+    this.props.onPortChanged(event.target.value.trim());
```
addaleax, line 133:
```suggestion
    this.props.onPortChanged(+event.target.value.trim());
```

because the typescript definitions say it’s a number? :)

addaleax (approved):
Nice :)

src/views/webview-app/components/connect-form/general-tab/host/host.tsx
Anemy: nice catch - updated - also added the event types to the function since it was thinking `event` was an `any`.

PR #226, opened 2020-12-17
VSCODE-219: edit documents opened form the sidebar

alenakhineika:
## Description

Edit documents that are opened from the tree-view.

- When a document is opened from the tree-view, the document is editable.
- Each document should be opened as a separate editor.
- Each document should be opened only once.
- If a user didn't make any changes the document can be simply closed without a confirmation modal.
- Editing happens in EJSON to preserve type information.
- When the document is saved (Cmd/Ctrl + S), the document is updated in MongoDB to reflect the changes that were made.
- For saving documents we use findOneAndReplace.
- When the document is saved, a notification informs the user that the change has been saved in MongoDB.
- When the update results in an error, the user is notified.
- It's possible to make edits to a document and save it more than once.
- When the extension is reopened the opened documents should have content.
- The documents can't be saved to MongoDB until a user connects to the proper namespace.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/editors/documentProvider.ts
+++ b/src/editors/documentProvider.ts
@@ -49,8 +49,6 @@ export default class DocumentViewProvider
         );
       }
 
-      // this._documentIdStore.remove(documentIdReference);
-
       // Ensure we're still connected to the correct connection.
       if (connectionId !== this._connectionController.getActiveConnectionId()) {
         vscode.window.showErrorMessage(
--- a/src/editors/documentProvider.ts
+++ b/src/editors/documentProvider.ts
@@ -12,7 +12,8 @@ export const DOCUMENT_ID_URI_IDENTIFIER = 'documentId';
 
 export const VIEW_DOCUMENT_SCHEME = 'VIEW_DOCUMENT_SCHEME';
 
-export default class DocumentViewProvider implements vscode.TextDocumentContentProvider {
+export default class DocumentViewProvider
+  implements vscode.TextDocumentContentProvider {
   _connectionController: ConnectionController;
   _documentIdStore: DocumentIdStore;
   _statusView: StatusView;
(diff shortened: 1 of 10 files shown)


src/test/suite/mdbExtensionController.test.ts
```diff
@@ -1217,38 +1225,317 @@ suite('MDBExtensionController Test Suite', function () {
...
+  test("if a user is not connected, documents won't be saved to MongoDB", async () => {
+    const fakeVscodeErrorMessage = sinon.fake();
+    sinon.replace(vscode.window, 'showErrorMessage', fakeVscodeErrorMessage);
+
+    sandbox.replaceGetter(vscode.window, 'activeTextEditor', () => ({
+      document: {
+        uri: {
+          query: [
+            '?documentLocation=mongodb',
+            'namespace=waffle.house',
```
addaleax, line 1402: I’m not sure if we have to care about this, but some characters like `#` would break URLs but are allowed in collection names … is that an issue here? :smile:

addaleax (approved):
Looks good! :)

src/editors/documentProvider.ts
```diff
@@ -36,20 +37,19 @@ export default class DocumentViewProvider implements vscode.TextDocumentContentP
...
         return reject(
-          new Error(
-            'Unable to fetch document: reference has expired.'
-          )
+          new Error('Unable to fetch document: reference has expired.')
         );
       }
 
-      this._documentIdStore.remove(documentIdReference);
+      // this._documentIdStore.remove(documentIdReference);
```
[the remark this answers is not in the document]
alenakhineika: Yes, this code can go! Forgot to remove it :) will do!

src/test/suite/mdbExtensionController.test.ts
alenakhineika: Vscode cares about it for us :) we shouldn't do here anything. but what I just noticed is that dataservice fails to save ejson. eg:

```
{
  "_id": {
    "$oid": "5ea8745ee4811fafe8b65ecb"
  },
  "koko": "nothing"
}
```

and throws the `Unable to save document: The dollar ($) prefixed field '$oid' in '_id.$oid' is not valid for storage.` error. Have you ever seen this before or tried to save ejson?

addaleax: I have not seen that error, no – that’s interesting … it sounds like it’s trying to insert the document as-is instead of first deserializing it as EJSON and then inserting it?

alenakhineika: Found the problem! the one I actually caused 😀 in the try-catch statement above I used JSON.parse() that broke the document format. Replaced this part with EJSON.parse() and now it saves correctly.

mmarcon:
When a document is reopened after VS Code is restarted and I try to edit it, I get this error.

image

The error is expected but it's not really helpful. Is it possible to show the connection name instead of the connection id?

MullinsN:
I like the changes that have appeared in the 0.4 update.

Please make sure that the marketplace extension documentation reflects some of the key changes. It helps with showing the growth in the extension.

mmarcon:
Thank you @MullinsN for trying out 0.4.0.

The market place homepage does talk about document editing and also includes a screenshot [here](https://marketplace.visualstudio.com/items?itemName=mongodb.mongodb-vscode#document-editing). Were you expecting something different?

If the database does not allow a change to be made, then VS Code will display an error that includes the server error message.

PR #227, opened 2020-12-18
VSCODE-221: Improve default connection name for connections

Anemy:
VSCODE-221

This PR updates how we give our connections names so that they fully represent the connection which was created. Currently, when a connection with multiple hosts and ports is added, we default to only showing one of the host and ports as the connections name. This is more misleading with srv records where we add a port to the end of the connection name. We now generate the name ourselves instead of relying on [connection-model's instanceId](https://github.com/mongodb-js/connection-model/blob/master/lib/model.js#L408).

It also pulls in bumped connection model and data-service, this updating driver to 3.6.3 from 3.6.1 and has the defaults in data service and connection model for direct connection and read preference secondary to continue previous behavior and not break old connections.

This is starting to bring a bit of extra logic into the connection controller, which I think might make sense to live somewhere else. That's being explored a bit in this wip branch https://github.com/mongodb-js/vscode/compare/VSCODE-205/remove-data-service . It also might make sense to make a `connections` folder and move the `ConnectionController` into there and find ways to break it down a bit more since it's starting to get a bit big. Open to ideas.

_Standalone connection naming_
basic with connect form

_Multiple host connection form naming_
multi host connection form naming

_SRV Record in form connection naming_
srv through form name

_Multiple host connection string naming_
multi host through connection string naming

_Ssh tunnel connection naming_
ssh tunnel naming


src/views/webview-app/components/connect-form/general-tab/host/host.tsx
```diff
@@ -247,7 +261,7 @@ export class HostInput extends React.PureComponent<StateProps & DispatchProps> {
               placeholder="27017"
               changeHandler={(event: React.ChangeEvent<HTMLInputElement>): void => this.onHostChanged({
                 host: host.host,
-                port: event.target.value
+                port: +event.target.value.trim()
```
mcasimir, line 264: isn't `parseInt` more clear?

addaleax (approved):

src/views/webview-app/components/connect-form/general-tab/host/host.tsx
addaleax: `parseInt()` would also accept values like `123 abc` … we don’t want that here, do we?

mcasimir: Actually `+event.target.value.trim()` is repeated quite a number of time, maybe a function like `parsePortFromEvent` would be nice

src/connectionController.ts
```diff
@@ -268,24 +269,50 @@ export default class ConnectionController {
...
+      return connectionModel.hostname;
+    }
+
+    if (connectionModel.hosts && connectionModel.hosts.length > 0) {
+      return connectionModel.hosts.map(
+        ({ host, port }) => `${host}:${port}`
+      ).join(',');
+    }
+
+    return connectionModel.hostname;
```
mcasimir, line 302: how is `hostname` used in VSCODE? don't we always have `hosts`?

mcasimir (approved):
Nice!

src/views/webview-app/components/connect-form/general-tab/host/host.tsx
Anemy: I think the trim here isn't actually needed since html's `number` input handles that for us. I'll remove that. I don't think we need a helper than for just having the `+`. Would be nice if typescript could pull types from the input type.

src/connectionController.ts
Anemy: I think it's there to make sure we work nicely with `connection-model`.  Yes `hosts` should always be there and this line should never really be reached.

src/views/webview-app/components/connect-form/general-tab/host/host.tsx
mcasimir: > parseInt() would also accept values like '123 abc' … we don’t want that here, do we?

Right + is even better, do we need to account for `NaN` though, i believe we have some validation though

Anemy: I don't think we need to. I don't think there's any easy way to get to `NaN` in an html number input field. I think it's probably best we don't perform much validation on the form level unless its something we really know the user probably doesn't want to do.

src/connectionController.ts
mcasimir: Ok cool!

src/views/webview-app/components/connect-form/general-tab/host/host.tsx
mcasimir: Makes sense!

mcasimir: > we don't perform much validation on the form level

I mean having a type=number field is the best kind of validation we can do there so is not possible to input any sort of bad value

PR #228, opened 2020-12-18
VSCODE-220: track document editing telemetry event

alenakhineika:
## Description

- Refactor MDBExtensionController to make sure that telemetry events are not being tracked in testing mode. In master some of them were sent to Segment.
- Track the `Document Updated` telemetry when a document is saved. Send the following values: `source = 'treeview'` and `success = true|false`.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/mdbExtensionController.ts
```diff
@@ -48,28 +48,20 @@ export default class MDBExtensionController implements vscode.Disposable {
   _languageServerController: LanguageServerController;
   _webviewController: WebviewController;
 
-  constructor(
-    context: vscode.ExtensionContext,
-    connectionController?: ConnectionController
```
alenakhineika, line 53: We haven't used it anywhere, right?

alenakhineika:
How can we share formatting config with each other to prevent these annoying lines indent changes in PRs? 🤔 It makes it hard to notice valuable code updates 😭

src/test/suite/connectionController.test.ts
```diff
@@ -984,17 +1001,13 @@ suite('Connection Controller Test Suite', function () {
       storageLocation: StorageScope.NONE
     };
 
-    const originalConnect = DataService.connect;
```
alenakhineika, line 987: DataService is a function and `DataService.connect` is always undefined. I have rewritten this part a bit to get access to the connect function.

src/extension.ts
```diff
@@ -34,7 +34,9 @@ export function activate(context: vscode.ExtensionContext): void {
     // Couldn't load keytar, proceed without storing & loading connections.
   }
 
-  mdbExtension = new MDBExtensionController(context);
+  const needTelemetry = true;
+
+  mdbExtension = new MDBExtensionController(context, needTelemetry);
```
alenakhineika, line 39: If it is not a test mode, we want to track telemetry events.

src/test/suite/explorer/helpExplorer.test.ts
```diff
@@ -55,12 +55,6 @@ suite('Help Explorer Test Suite', function () {
     const testHelpExplorer =
       mdbTestExtension.testExtensionController._helpExplorer;
 
-    sinon.replace(
-      mdbTestExtension.testExtensionController._telemetryController,
-      'trackLinkClicked',
-      sinon.fake.resolves()
-    );
-
```
alenakhineika, line 63: Since now all tests have telemetry disabled we don't need anymore to fake telemetry events individually.

src/test/suite/index.ts
```diff
@@ -29,52 +25,54 @@ export function run(): Promise<void> {
...
-      mdbTestExtension.testExtensionContext = new TestExtensionContext();
-      mdbTestExtension.testExtensionController = new MDBExtensionController(
-        mdbTestExtension.testExtensionContext
-      );
-      mdbTestExtension.testExtensionController.activate();
+        // Activate the extension.
+        mdbTestExtension.testExtensionContext = new TestExtensionContext();
+        mdbTestExtension.testExtensionController = new MDBExtensionController(
+          mdbTestExtension.testExtensionContext,
+          false
```
alenakhineika, line 43: We set a `neetTelemery` flat to false, which means all telemetry events will be disabled in test mode.

src/test/suite/index.ts
```diff
@@ -29,52 +25,54 @@ export function run(): Promise<void> {
...
-      // headless linux.
-      ext.keytarModule = new KeytarStub();
+        // We avoid using the user's credential store when running tests
+        // in order to ensure we're not polluting the credential store
+        // and because its tough to get the credential store running on
+        // headless linux.
+        ext.keytarModule = new KeytarStub();
 
-      // Disable metrics.
-      vscode.workspace.getConfiguration('mdb').update('sendTelemetry', false);
```
alenakhineika, line 54: This flag we don't need anymore in test mode.

src/editors/playgroundController.ts
```diff
@@ -315,15 +315,24 @@ export default class PlaygroundController {
...
         vscode.window.showTextDocument(doc, { preview: false, viewColumn });
-      });
+      },
+      (error) => {
+        vscode.window.showErrorMessage(
+          `Unable to open results document: ${error.message}`
+        );
+        log.error('Open playground ERROR', error);
+      }
+    );
```
alenakhineika, line 335: Missing error catching here was causing the same error in tests, that was described in this ticket:
https://jira.mongodb.org/browse/VSCODE-204

Let's further keep an eye on it.

src/editors/playgroundController.ts
```diff
@@ -357,7 +366,9 @@ export default class PlaygroundController {
 
       this._outputChannel.clear();
       if (evaluateResponse.outputLines) {
-        for (const line of evaluateResponse.outputLines) {this._outputChannel.appendLine(line.content);}
+        for (const line of evaluateResponse.outputLines) {
```
mcasimir, line 369: 👍

src/telemetry/telemetryController.ts
```diff
@@ -58,34 +58,44 @@ type NewConnectionTelemetryEventProperties = {
...
-  PLAYGROUND_LOADED = 'Playground Loaded'
+  PLAYGROUND_LOADED = 'Playground Loaded',
+  DOCUMENT_UPDATED = 'Document Updated'
 }
 
 /**
  * This controller manages telemetry.
  */
 export default class TelemetryController {
+  public _needTelemetry: boolean;
```
mcasimir, line 87: Is `public` with the underscore on purpose?

src/telemetry/telemetryController.ts
```diff
@@ -58,34 +58,44 @@ type NewConnectionTelemetryEventProperties = {
...
   PLAYGROUND_SAVED = 'Playground Saved',
-  PLAYGROUND_LOADED = 'Playground Loaded'
+  PLAYGROUND_LOADED = 'Playground Loaded',
+  DOCUMENT_UPDATED = 'Document Updated'
 }
 
 /**
  * This controller manages telemetry.
  */
 export default class TelemetryController {
```
mcasimir, line 86: Just curious, why is this called `Controller`? this doesn't strike me as something that receives commands or requests from the UI, but rather something more internally called by other parts of our code, how is the lingo for VSCode? Is this not a controller as in MVC or similar?

src/telemetry/telemetryController.ts
```diff
@@ -150,8 +160,20 @@ export default class TelemetryController {
     this._segmentAnalytics?.flush();
   }
 
-  public needTelemetry() {
-    return vscode.workspace.getConfiguration('mdb').get('sendTelemetry');
+  private needTelemetry(): boolean {
+    if (this._needTelemetry !== true) {
```
mcasimir, line 164: Isn't `!this._needTelemetry` more immediate?

What about to disambiguate a bit, we could call this `shouldSendTelemetry`, `_needTelemetry` could be something like `telemetryFeatureEnabled` and `confNeedTelemetry` could be `telemetryEnabledByUser`?

Maybe i'm giving horrible suggestions here.. is just to make sure is understandable that there are 3 different things and we call them in a different way:
- a feature flag to disable the telemetry during tests
- a configuration for the user
- something to really decide wether the telemetry should be sent or not.

src/extension.ts
mcasimir, line 39: Probably you already considered it, but we could do constructor injection here and pass the telemetry as a dependency instead:

```
new MDBExtensionController(context, telemetryController)
```

and use `sinon.createStubInstance` to mock the telemetry away with all the functions and events, and even spy to test that telemetry events will be actually sent.

For segment specifically another alternative is to use a fake or empty segment key. Last time i checked the segment library was not exploding and just not sending events.

src/test/suite/connectionController.test.ts
```diff
@@ -889,14 +900,20 @@ suite('Connection Controller Test Suite', function () {
       const currentConnectingVersion = testConnectionController.getConnectingVersion();
 
       assert(currentConnectingVersion !== null);
-      assert(currentConnectingVersion === testConnectionController._connections[
-        Object.keys(testConnectionController._connections)[0]
-      ].id);
+      assert(
```
mcasimir, line 903: `assert.equal` or `assert.strictEqual` will give you a better error message and output a diff

src/test/suite/index.ts
```diff
@@ -29,52 +25,54 @@ export function run(): Promise<void> {
   const testsRoot = path.join(__dirname, '..');
 
   return new Promise((c, e) => {
```
mcasimir, line 27: I'm not really sure why we use `new Promise` with that chain of `then` and `catch` instead of making `run` an async function.

If callbacks is the issue we can just import `util.promisify` and turn those into something that can be awaited. 

```
const globAsync = promisify(glob);
```

or 

```
const runMocha = promisify(mocha.run.bind(mocha));
```

This mixture of async code styles is something i've noticed frequently in VSCode, is so much easier to have `promisify` from the node standard library  and `async/await`  to take care of all the exception handling with all the silly corner cases. It makes the code a lot more readable, safe and easy to maintain.

src/test/suite/telemetry/telemetryController.test.ts
```diff
@@ -1,14 +1,11 @@
 import * as vscode from 'vscode';
-import { StorageController } from '../../../storage';
-import { TestExtensionContext } from '../stubs';
+import * as path from 'path';
```
mcasimir, line 2: isn't this just `import path from 'path';`?

mcasimir (approved):
Nice stuff!

src/mdbExtensionController.ts
mcasimir: Isn't this useful to pass a mock for tests?

alenakhineika: I replaced ConnectionController with needTelemetry flag because I think ConnectionController is not used here at all. Just wanted to double-check that ConnectionController isn't placed here for future needs.

src/extension.ts
alenakhineika: Personal preference :) i usually use `is` or `need` prefix.

src/editors/playgroundController.ts
alenakhineika: We previously had an issue with catching unhandled promise rejections for `vscode.workspace.openTextDocument` and only catching errors as a second argument solved the issue: `.then(() => { /* success */ }, () => { /* fail */ });`. You can see that we do it all the time here in playgroundController.ts.

But your point about awaiting `this.openResultAsVirtualDocument(viewColumn)` results is totally fair. The only thing that instead of returning `return this.openResultAsVirtualDocument(viewColumn)` tenable result I will just await result and return `return resolve(false);` because the expected result of the function is boolean.

src/connectionController.ts
```diff
@@ -284,19 +282,19 @@ export default class ConnectionController {
       sshTunnelOptions.host &&
       sshTunnelOptions.port
     ) {
-      return `SSH to ${connectionModel.hosts.map(
-        ({ host, port }) => `${host}:${port}`
-      ).join(',')}`;
+      return `SSH to ${connectionModel.hosts
```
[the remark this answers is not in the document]
alenakhineika: The connection model should always return hosts longer than zero:

https://github.com/mongodb-js/connection-model/blob/master/lib/model.js#L85

src/editors/playgroundController.ts
```diff
@@ -315,15 +315,24 @@ export default class PlaygroundController {
...
-        this.getVirtualDocumentUri(this.playgroundResult?.content)
-      )
-      .then((doc) => {
+  private openResultAsVirtualDocument(viewColumn: vscode.ViewColumn) {
+    const content =
+      this.playgroundResult && this.playgroundResult.content
+        ? this.playgroundResult.content
+        : '';
+
+    vscode.workspace.openTextDocument(this.getVirtualDocumentUri(content)).then(
```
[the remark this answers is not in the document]
alenakhineika: We previously had an issue with catching unhandled promise rejections for vscode.workspace.openTextDocument and only catching errors as a second argument solved the issue: .then(() => { /* success */ }, () => { /* fail */ });. You can see that we do it all the time here in playgroundController.ts.

src/mdbExtensionController.ts
```diff
@@ -48,28 +48,20 @@ export default class MDBExtensionController implements vscode.Disposable {
...
-  ) {
+  constructor(context: vscode.ExtensionContext, needTelemetry?: boolean) {
```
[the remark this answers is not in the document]
alenakhineika: Added returning type for the `openResultAsVirtualDocument` function.

I was actually choosing between `needTelemetry` as a prop and `{needTelemetry: true}` options object :) Since you also like it more, then let's use the object instead!

src/telemetry/telemetryController.ts
alenakhineika: No, it should be private. Left over after refactoring. Good catch!

src/telemetry/telemetryController.ts
alenakhineika: Do you have a better naming in mind? Maybe telemetryModule? 🤔

src/telemetry/telemetryController.ts
alenakhineika: Love your suggestions! ❤️ Also added comments to make it absolutely clear what is what.

src/extension.ts
alenakhineika: This is an entry point, we don't have controllers here yet.

src/test/suite/telemetry/telemetryController.test.ts
alenakhineika: I remember it was the safest way to import it. Was failing with just `import path from 'path';`, probably on windows. Don't remember exactly what.

PR #230, opened 2021-01-04
chore(code-lint): Add lint to check (VSCODE-215)

Anemy:
This PR adds a `lint` check for our `check` script, which runs in the precommit. Updated our eslint to search all the files, and added all of our entries to the dependency check which [mongodb-js-precommit](https://github.com/mongodb-js/precommit) runs. 

This PR includes fixes for the existing lint errors, and some of the errors. There are still a couple warnings (see screenshot below). Also updates a couple of our dependencies and our eslint config generation to be a bit cleaner and work with the new deps.

Here's what the commit output now looks like:
Screen Shot 2021-01-04 at 6 34 43 AM

Looks like [mongodb-js-precommit](https://github.com/mongodb-js/precommit) is throwing en error when trying to lint now. I think we want to add an optional parameter to skip the lint check. Haven't looked into what's causing it to fail yet.
Screen Shot 2021-01-04 at 6 38 46 AM


src/editors/editorsController.ts
```diff
@@ -229,117 +232,132 @@ export default class EditorsController {
     });
   }
 
-  saveDocumentToMongoDB(): Promise<boolean> {
```
Anemy, line 232: This method was causing a complexity lint error, this pr adds a `replaceDocument` function which does the MongoDB findOneAndReplace of it. cc @alenakhineika

mcasimir:
`Error: Callback was already called` is haunting VSCode too :)

package.json
```diff
@@ -53,7 +53,7 @@
     "test-webview": "jest",
     "posttest": "mongodb-runner stop --port=27018",
     "vscode:prepublish": "npm run clean && npm run compile:keyfile && npm run compile:resources && webpack --mode production",
-    "check": "mongodb-js-precommit './src/**/*{.ts}'",
+    "check": "npm run lint && mongodb-js-precommit './src/**/*.{js,ts,tsx}'",
```
mcasimir, line 56: nice, so this works with ts too, also the depcheck?

package.json
```diff
@@ -923,14 +924,17 @@
...
-    "typescript": "^4.0.3",
+    "typescript": "^4.1.3",
     "vsce": "^1.81.1",
     "vscode-test": "^1.3.0",
     "webpack": "^4.42.0",
     "webpack-cli": "^3.3.11",
     "xvfb-maybe": "^0.2.1"
   },
   "dependency-check": {
+    "entries": [
```
mcasimir, line 935: ohh i see, maybe we need to bump depcheck inside the mongodb-js-precommit thing: newer versions probably supports ts: https://www.npmjs.com/package/depcheck

src/editors/editorsController.ts
```diff
@@ -229,117 +232,130 @@ export default class EditorsController {
...
 
-      if (dataservice === null) {
-        // Send metrics to Segment.
-        this._telemetryController.trackDocumentUpdated('treeview', false);
+      const errorMessage = `Unable to save document: ${error.message}`;
 
-        vscode.window.showErrorMessage(
-          `Unable to save document: no longer connected to '${connectionName}'`
-        );
+      // Send metrics to Segment.
```
mcasimir, line 284: this code is duplicated here and there, what about:

``` js
failSaveDocument(errorMessage) {
   this._telemetryController.trackDocumentUpdated('treeview', false);
        );	


        return resolve(false);	      vscode.window.showErrorMessage(errorMessage);
}
```

src/editors/editorsController.ts
```diff
@@ -229,117 +232,130 @@ export default class EditorsController {
...
-        },
-        (error) => {
-          this._statusView.hideMessage();
+    if (
+      documentLocation !== 'mongodb' ||
+        !namespace ||
+        !connectionId ||
+        !documentId
+    ) {
+      vscode.commands.executeCommand('workbench.action.files.save');
```
mcasimir, line 317: The code from line `302` seems to contain a valuable piece of business logic that doesn't really stands out to someone that doesn't have a tons of context:

What is this condition for (is it cause the doc was never saved before? or some other special condition?), why do we save the file in this case and not in the other?

```
documentLocation !== 'mongodb' ||
        {	        !namespace ||
          _id: documentId	        !connectionId ||
        },	        !documentId
```

I think it would be helpful to break this method down with more functions to clarify a bit this behaviour or at least add some comment

src/editors/editorsController.ts
```diff
@@ -229,117 +232,130 @@ export default class EditorsController {
...
-        // Send metrics to Segment.
-        this._telemetryController.trackDocumentUpdated('treeview', false);
+    log.info('save document to MongoDB', activeEditor);
 
-        vscode.window.showErrorMessage(error.message);
+    if (!activeEditor) {
+      return true;
+    }
 
-        return resolve(false);
```
mcasimir, line 303: Nice that we are converting this to async/await

mcasimir (approved):
Nice! Maybe Alena wants to review this too.

package.json
Anemy: The `mongodb-js-precommit './src/**/*.{js,ts,tsx}'` part is specifying where it's going to lint (which is failing rn - and I don't think it lints typescript anyway). I'll open a pr adding a parameter to optionally skip linting in that package in a sec.

Anemy: The depcheck places are specified in package.json`dependency-check.entries`

package.json
Anemy: Looks like that package might be a bit better than the one we're using: https://www.npmjs.com/package/dependency-check 
Would be good to be checking the `src` and not built `dist` files. I'll make a ticket for that: https://jira.mongodb.org/browse/COMPASS-4551

package.json
Anemy: https://github.com/mongodb-js/precommit/pull/75 <- Adds an argument to `mongodb-js-precommit` to skip the lint stage.

src/editors/editorsController.ts
Anemy: Looks like this checks to make sure the document being saved has all of the correct uri parameters inside of vscode. Maybe there's a way the uri can be edited when a document is open, so this check hopefully protects against any malformed uris.
I think checking for `documentId` to be true here might make a user unable to save a document with an `_id` of `false`. Not sure if that's a bug we want to fix. 🤔 
I'd like to keep this PR mostly just about the linting, not much on the logic changing side, so I'll just leave a comment for now if that's cool.

src/editors/editorsController.ts
Anemy: Good call. Updated & updated the comment to be a bit more generic.

package.json
mcasimir: Oh now i get what is that for! 😄 

Why are we even using `mongodb-js-precommit`? it doesn't work with ts for eslint and depcheck anyway. Isn't it massively simpler to just add `eslint` and `depcheck` right in this project?

package.json
mcasimir: Oh i see i thought we were using `depcheck`. But for VSCode i'd really consider just doing everything in project, i think having those simple support stuff in a separate package makes some sense in Compass cause of the 100 repos, but doesn't really offer much help here, especially considering it was build for a different tech stack.

src/editors/editorsController.ts
alenakhineika: json document can be mongodb document or just any other json document that we don't want to save to mongodb. we check if the uri have all required properties that we need in order to save it to mongodb we don't get inside this if and later in code call `replaceDocument` function. But if not all the data presented it means there is some trouble with this document. Maybe user tried to  malform uri or this is just not the mongodb document that we opened from the sidebar - in this case instead of saving it to mongodb we just handle this like a regular editor and ask user if they want to save it on disc as a file.

src/explorer/playgroundsTree.ts
```diff
@@ -14,13 +14,17 @@ export class FileStat implements vscode.FileStat {
...
+      return vscode.FileType.File;
+    }
+    if (this.fsStat.isDirectory()) {
+      return vscode.FileType.Directory;
+    }
+    if (this.fsStat.isSymbolicLink()) {
+      return vscode.FileType.SymbolicLink;
+    }
+
+    return vscode.FileType.Unknown;
```
alenakhineika, line 27: Thx :)

src/language/server.ts
```diff
@@ -220,7 +220,7 @@ connection.onRequest(new RequestType('textDocument/codeLens'), (event) => {
   // return parsed;
 });
 
-connection.onDidChangeWatchedFiles((_change) => {
+connection.onDidChangeWatchedFiles((/* _change */) => {
```
alenakhineika, line 223: Nice.

src/language/visitor.ts
```diff
@@ -119,22 +119,22 @@ export class Visitor {
...
             break;
           case 'ExpressionStatement':
-            self.visitExpressionStatement(path.node);
+            this.visitExpressionStatement(path.node);
             break;
           case 'ObjectExpression':
-            self.visitObjectExpression(path.node);
+            this.visitObjectExpression(path.node);
+            break;
+          default:
```
alenakhineika, line 137: Yeah, that looks much better!

src/language/worker.ts
```diff
@@ -140,9 +140,9 @@ const getListDatabases = async (
     const result = await serviceProvider.listDatabases('admin');
     const databases = result
       ? result.databases.map((item) => ({
-          label: item.name,
-          kind: CompletionItemKind.Value
-        }))
+        label: item.name,
+        kind: CompletionItemKind.Value
+      }))
```
alenakhineika, line 145: Will this endless adding/deleting spaces be fixed by this PR with lint improvements?

package.json
Anemy: yup it's easier lol. I'll update this pr to remove `mongodb-js-precommit`.

src/telemetry/telemetryController.ts
```diff
@@ -34,14 +34,15 @@ type CloudInfo = {
 
 type LinkClickedTelemetryEventProperties = {
   screen: string;
-  link_id: string;
+  link_id: string; // eslint-disable-line camelcase
```
alenakhineika, line 37: Probably we can disable it for the whole file. Because we use camelcase props for telemetry events only in this file. And later we can have more such events.

src/language/worker.ts
Anemy: With https://github.com/mongodb-js/vscode/pull/231 yes

package.json
Anemy: Agreed. I'll update to just use `depcheck` and we can remove the `mongodb-js-precommit` dependency :D
The eslint config is about the same with https://github.com/mongodb-js/mongosh/blob/master/config/eslintrc.base.js except for a rule or two. Might be a nice thing to do at some point, don't see much need right now.

src/editors/editorsController.ts
mcasimir: @alenakhineika Thanks for the clarification! Yeah .. i think this as a comment would help a lot, this is quite deep to understand for someone who did not participated in the implementation.

I often find factoring this contrived concepts in dedicated methods helpful, for example with this:

```
_isMongodbDocument(document): boolean {
  const uriParams = new URLSearchParams(document.uri.query);
  const documentLocation = uriParams.get(DOCUMENT_LOCATION_URI_IDENTIFIER) || '';
  const namespace = uriParams.get(NAMESPACE_URI_IDENTIFIER) || '';
  const connectionId = uriParams.get(CONNECTION_ID_URI_IDENTIFIER);
  const documentIdReference =
      uriParams.get(DOCUMENT_ID_URI_IDENTIFIER) || '';
  const documentId = this._documentIdStore.get(documentIdReference);
  
  return documentLocation === 'mongodb' && namespace && connectionId && documentId;  
}
```

That `if` becomes:

``` js
if (!this._isMongodbDocument(activeEditor.document)) {
    vscode.commands.executeCommand('workbench.action.files.save'); // save to disk instead of mongodb
}
```

which both reduces the complexity of the main function, and also could help to put all those checks in their context.

Anemy:
@mcasimir Removed the `mongodb-js-precommit` dependency and added `depcheck`. Looks like it's doing a better job than `dependency-check` was doing too, updated package.json to remove a few unused deps and explicitly add `download` and `fs-extra` which we didn't have. Here's what the precommit output looks like now:
Screen Shot 2021-01-04 at 2 19 14 PM

src/editors/editorsController.ts
alenakhineika: I like it a lot! will refactor it!

src/editors/editorsController.ts
```diff
@@ -229,117 +232,126 @@ export default class EditorsController {
...
-      if (activeConnectionId !== connectionId) {
-        // Send metrics to Segment.
-        this._telemetryController.trackDocumentUpdated('treeview', false);
+    const findOneAndReplace = util.promisify(dataservice.findOneAndReplace);
 
-        vscode.window.showErrorMessage(
-          `Unable to save document: no longer connected to '${connectionName}'`
-        );
+    try {
+      await findOneAndReplace(
```
alenakhineika, line 263: Isn't it throws an error: `Unable to save document: Cannot read property 'client' of undefined`?

Anemy: Ah yup needs a bind. Updating...

package.json
```diff
@@ -884,20 +885,22 @@
     "@types/sinon": "^9.0.1",
     "@types/vscode": "^1.49.0",
     "@types/ws": "^7.2.4",
-    "@typescript-eslint/eslint-plugin": "^2.19.2",
-    "@typescript-eslint/parser": "^2.19.2",
+    "@typescript-eslint/eslint-plugin": "^4.11.1",
```
alenakhineika, line 888: This and "@typescript-eslint/parser" are already 4.12.0

Anemy: 👍  will rebase once the editing playground results pr goes through and then we can look at this again and merge.

Anemy:
Add https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/recommended-requiring-type-checking.ts#L21

Anemy:
@alenakhineika @mcasimir rebased off master, fixed those issues. Also added `parser: '@typescript-eslint/parser'` and `'plugin:@typescript-eslint/recommended-requiring-type-checking'` to the eslint config. They introduce a ton of warnings, but they all look like good warnings (we could reduce overtime).

PR #232, opened 2021-01-05
VSCODE-222: Edit documents from playground results

alenakhineika:
## Description

1) https://jira.mongodb.org/browse/VSCODE-222

- When one or more documents returned by a playground are displayed in the read-only editor with the results, documents will have a code lens to access the editing functionality when:
  - The playground result is a document returned by `findOne()`.
  - Or list of documents returned by `find()`.
  - Document's `_is` is presented.
- By clicking on the code lens, the user is taken to a new editable version of the document, displayed in EJSON (the editable document is shown with all its fields)
- When the document is saved (Cmd/Ctrl + S or with the menu), the document is saved to MongoDB.
- When the document is saved, a notification informs the user that the change has been saved in MongoDB.
- When the update results in an error, the user is notified.
- It's possible to make edits to a document and save it more than once.
- Update playground results behavior:
  - To provide a better user experience and to be able to drag and drop playground results to any position in the window (including splitting window down), replace results window reopening with content updating in the current virtual document. This solution removes `.txt` and `.js` file extensions from `uri.path`. Now users have to specify the extension manually when saving document to disc.

2) https://jira.mongodb.org/browse/VSCODE-223

Show and write mongodb documents using a FileSystemProvider that helps us to avoid storing files on the user's file system, improve the breadcrumbs we show on an open document (before we showed the /tmp/.../file).

3) https://jira.mongodb.org/browse/VSCODE-185

Improve how open documents are handled when vscode is re-opened. If the user had an open document editor, we want to ensure that such a document is no longer opened when the user closes and reopens VS Code.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/editors/documentController.ts
```diff
@@ -0,0 +1,262 @@
...
+
+    try {
+      newDocument = EJSON.parse(editor.document.getText());
+    } catch (error) {
+      return this._saveDocumentFailed(error.message);
+    }
+
+    this._statusView.showMessage('Saving document...');
+
+    dataservice.findOneAndReplace(
```
Anemy, line 182: Should we promisify this?
```ts
const findOneAndReplace = util.promisify(dataservice.findOneAndReplace.bind(dataservice));

try {
  await findOneAndReplace(...
```
Does the `replaceDocument` function return before this completes?
Do we have a test that uses a database for this function? Might be good for ensuring we don't introduce any regressions in future.

src/editors/documentController.ts
```diff
@@ -0,0 +1,262 @@
...
+    const uri: vscode.Uri = vscode.Uri.file(localDocPath).with({
+      query: `?${documentLocation}&${namespaceUriQuery}&${connectionIdUriQuery}&${documentIdUriQuery}`
+    });
+
+    return new Promise(async (resolve, reject) => {
+      vscode.workspace.openTextDocument(uri).then((doc) => {
+        vscode.window
+          .showTextDocument(doc, { preview: false, preserveFocus: true })
+          .then(() => resolve(true), reject);
+      }, reject);
```
Anemy, line 259: How do you feel about making this use async/await?
```ts
try {
  const doc = await vscode.workspace.openTextDocument(uri);
  await  vscode.window.showTextDocument(doc, { preview: false, preserveFocus: true });
} catch (err: Error) {
   throw err;
}
return true;
```

src/editors/playgroundController.ts
```diff
@@ -502,7 +541,19 @@ export default class PlaygroundController {
     });
   }
 
-  public deactivate(): void {
+  refreshPlaygroundResultDocument(
+    documentId: EJSON.SerializableTypes,
+    namespace: string
+  ): Promise<boolean> {
+    return new Promise(async (resolve) => {
+      return this._documentController.openEditableDocument(
```
Anemy, line 549: Doesn't look like the promise resolves, maybe we can just return the `this._documentController.openEditableDocument` result?

src/editors/playgroundResultProvider.ts
```diff
@@ -1,44 +1,145 @@
...
+        });
+      } else if (type === 'Document' && content._id && namespace) {
+        // When the playground result is the single document,
+        // show the single code lense after {.
+        codeLensesInfo.push({ line: 1, documentId: content._id, namespace });
+      }
+
+      this._editDocumentCodeLensProvider?.refresh(codeLensesInfo);
+
+      return resolve(JSON.stringify(content, null, 2));
```
Anemy, line 142: I might be missing something - maybe this function be synchronous?

src/editors/documentController.ts
```diff
@@ -0,0 +1,262 @@
...
+
+        // Send metrics to Segment.
+        this._telemetryController.trackDocumentUpdated('treeview', true);
+
+        // Save document changes to active editor.
+        activeEditor?.document.save();
+
+        // Update parent list of documents to reflect the changes that were made.
+        vscode.commands.executeCommand(
+          EXTENSION_COMMANDS.MDB_REFRESH_PLAYGROUND_RESULT_CONTENT,
```
Anemy, line 208: A bit of a functionality question - do we want to update the documents that were returned from the result of a playground here? I think it might be a little counter intuitive for the playground results to be updated when a document is updated. Would just the updated document be changed? What happens if other documents provided in the `find` of the playground results were updated in the background?

Anemy (commented):
nice work - this'll be a nice functionality add for playgrounds. 
Added a couple comments. I think one main one on the updating results content functionality.
I like separating out of the mongodb document operations from the editor controller. Something I noticed is the `documentController` is doing is the opening of the documents in the vscode editor - should that be something the editor controller does, and the `documentController` instead does just the mongodb document content providing side of things?

src/editors/documentController.ts
alenakhineika: Replaced callbacks with `util.promisify` and added a test from `replaceDocument `.

src/editors/documentController.ts
alenakhineika: It was written like this initially but sometimes caused an uncaught promise rejection error that was coming from somewhere. I also would like to see it async/await, but let's come back to it later in a separate PR for better code readability and additional testing.

src/editors/playgroundController.ts
alenakhineika: Refactored the part about opening and saving mdb docs. Liked a lot your suggestion about using the in-memory file system 👍

src/editors/playgroundResultProvider.ts
alenakhineika: Nice observation! Actually, I took as an example CollectionViewProvider and `provideTextDocumentContent` there is a Promise. But according to documentation can have different responses types: `T | undefined | null | Thenable<T | undefined | null>`. Refactored to `provideTextDocumentContent(): string` and `return JSON.stringify(content, null, 2);`.

src/editors/documentController.ts
alenakhineika: According to the acceptance criteria, we should update the parent list: `When the document is saved (Cmd/Ctrl + S or with the menu), the document in the collection is updated to reflect the changes that were made`

https://jira.mongodb.org/browse/VSCODE-178

But let's talk about it at the dosing sync.

alenakhineika:
@Anemy thank you very much for the review! ✨ Refactored branch according to your comment. PR is huge, sorry, but it would be much more work to implement and test intermediate solutions. As you can see not only lots of code was added but also removed! Please let me know if you want to review and discuss it together! Would be nice some time to do pair refactoring, I think it would be even cooler do discuss things and share knowledge before PR is ready!

src/editors/playgroundResultProvider.ts
```diff
@@ -1,44 +1,141 @@
...
+        if (this._playgroundResult && this._playgroundResult.content) {
+          this._playgroundResult.content[item] = document[item];
+        }
+      });
+    }
+
+    this.onDidChangeEmitter.fire(this._uri);
+  }
 
-      return resolve(this._playgroundController.playgroundResult?.content);
```
alenakhineika, line 41: This part potentially could cause https://jira.mongodb.org/browse/VSCODE-204 error, because provideTextDocumentContent expects Promise<string> result, and this._playgroundController.playgroundResult?.content could be undefined at some point. When manually tested different inputs and outputs of playground execution received once the same error that in the ticket. Refactored this function to check different result types and provided tests for them.

Anemy (commented):
Nice - here are the comments from our discussion, I think I didn't hit the post review button, just left the comments yesterday 😅

src/editors/documentController.ts
```diff
@@ -0,0 +1,251 @@
...
+
+      throw new Error(errorMessage);
+    }
+
+    const find = util.promisify(dataservice.find.bind(dataservice));
+
+    this._statusView.showMessage('Fetching document...');
+
+    try {
+      const documents = await find(
```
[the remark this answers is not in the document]
alenakhineika: The `findOne` method is not provided by data service: https://github.com/mongodb-js/data-service/blob/master/lib/data-service.js

src/editors/playgroundResultProvider.ts
```diff
@@ -1,44 +1,107 @@
...
+  _uri: vscode.Uri;
+
+  constructor(context: vscode.ExtensionContext) {
+    this._editDocumentCodeLensProvider = new EditDocumentCodeLensProvider();
+    this._playgroundResult = {
+      namespace: null,
+      type: null,
+      content: undefined
+    };
+    this._uri = vscode.Uri.parse('');
```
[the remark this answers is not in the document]
alenakhineika: Now uri lives in playgroundResultProvider.ts itself.

src/editors/playgroundResultProvider.ts
```diff
@@ -1,44 +1,107 @@
...
+  // When the playground evaluation response is received,
+  // use playgroundResult (namespace, type, content)
+  // to provide a virtual document content.
+  refreshPlaygroundResult(playgroundResult?: OutputItem): void {
+    if (playgroundResult) {
+      this._playgroundResult = playgroundResult;
+    }
+  }
+
+  async reopenResultAsVirtualDocument(
```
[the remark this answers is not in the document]
alenakhineika: One was updating only a class propery, another was refreshing the window. Refactored to clean it up and reflect other changes in the playground controller.

src/editors/playgroundController.ts
```diff
@@ -488,21 +516,17 @@ export default class PlaygroundController {
...
-        (doc) => vscode.window.showTextDocument(doc, 1, false),
-        (error) => {
-          vscode.window.showErrorMessage(`Unable to read file: ${filePath}`);
-          log.error('Open playground ERROR', error);
-        }
-      );
-
-      return resolve(true);
+  openPlayground(filePath: string): Promise<boolean> {
+    return new Promise(async (resolve, reject) => {
```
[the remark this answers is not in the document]
alenakhineika: Leftover after refactoring, removed.

src/editors/documentController.ts
```diff
@@ -0,0 +1,251 @@
...
+    const connectionId = uriParams.get(CONNECTION_ID_URI_IDENTIFIER);
+    const documentIdReference = uriParams.get(DOCUMENT_ID_URI_IDENTIFIER) || '';
+    const documentId = this._documentIdStore.get(documentIdReference);
+
+    // If not MongoDB document save to disk instead of MongoDB.
+    if (
+      activeEditor.document.uri.scheme !== 'VIEW_DOCUMENT_SCHEME' ||
+      !namespace ||
+      !connectionId ||
+      !documentId
```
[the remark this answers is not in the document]
alenakhineika: Nice catch, now I check explicitly for undefined and null.

src/editors/documentController.ts
```diff
@@ -0,0 +1,251 @@
...
+      this._statusView.hideMessage();
+
+      return JSON.parse(EJSON.stringify(documents[0]));
+    } catch (error) {
+      const errorMessage = `Unable to find document: ${error.message}`;
+
+      this._statusView.hideMessage();
+      vscode.window.showErrorMessage(errorMessage);
+
+      return null;
```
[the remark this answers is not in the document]
alenakhineika: Glad we had a call and discussed these returning types. Now it is much more clear for me when should be returned what. But pls take a look again at the refactored code to verify that it is what you meant.

src/editors/documentController.ts
```diff
@@ -0,0 +1,251 @@
...
+      const documents = await find(
+        namespace,
+        {
+          _id: documentId
+        },
+        {
+          limit: 1
+        }
+      );
+
```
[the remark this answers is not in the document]
alenakhineika: There were actually a couple of places where `this._statusView.hideMessage()` was missing. Updated to make sure we always hide this message.

src/editors/documentController.ts
```diff
@@ -0,0 +1,228 @@
...
+    let connectionName = '';
+
+    if (activeConnectionId) {
+      connectionName = this._connectionController.getSavedConnectionName(
+        activeConnectionId
+      );
+    }
+
+    this._statusView.showMessage('Fetching document...');
+    const document = await this._fetchDocument({
```
Anemy, line 185: I think we might want to check to make sure this document still exists here and show an error if not. For example when a user disconnects from their connection and then clicks the `Edit Document` code lens it currently shows a bit of a cryptic error which we could handle here I think.

Anemy: Looks like `_fetchDocument` throws an error when the document doesn't exist, so we can catch that here?

Anemy (commented):
Looking good, one comment on handling when a document can't be fetched

src/editors/documentController.ts
Anemy: * chatted - we can move that catching logic into the editors controller.

alenakhineika: It is already in editorsController, but let's move as we discussed the uri building to editorsController and rename some methods in documentController to resolve this confusion.

src/editors/mongoDBDocumentService.ts
```diff
@@ -0,0 +1,154 @@
...
+    const { documentId, namespace } = data;
+    const activeConnectionId = this._connectionController.getActiveConnectionId();
+    const connectionName = activeConnectionId
+      ? this._connectionController.getSavedConnectionName(activeConnectionId)
+      : '';
+    const dataservice = this._connectionController.getActiveDataService();
+
+    if (dataservice === null) {
+      return this._fetchDocumentFailed(
+        `no longer connected to ${connectionName}`
```
Anemy, line 122: I think connection name we show here will always be an empty string on a disconnected connection - is it possible to show the connection name of the connection which the playground result was run against?

Anemy (approved):
lgtm - I think we might want to associate playground results with the connection id which generated those results, so a user can't try to open up a document from playground results from a different connection. I think we could probably address that later. Left one comment regarding that, lgtm though nice work 🔥

src/editors/mongoDBDocumentService.ts
alenakhineika: I will address this in the next PR, thx for pointing at it!

Issue #223, opened 2020-12-05
Feature Request - Option to display Output panel


MatthewS2077:
## Feature Request

Option to display Output panel

## Detailed Description

Add option display the Output panel which defaults to true (the current behaviour).

If false, then running a query will not open the Output panel.

## Context

It's annoying to have the Output panel automatically opened, and especially because I like my panel on the right hand side of screen.

I prefer to open panels myself using the default key mapping (Ctrl-J) .

MatthewS2077:
Just to clarify, I'm talking about the Output "pane" in the "panel" (the panel shows the _Problems, Terminal, Output, Debug Console_ panes).

I always have the Terminal pane open, but whenever I run a command in Mongo Playground it switches to the Output pane (Playground Output). 

I've never seen anything come up in the Output pane, and it's annoying to have to continually switch back to the Terminal pane.

alyahyaoui:
same here
when running the selected lines in the playground it opens a new tab showing the result. but at the same time the output pane in the panel opens automatically and it's completely blank it doesn't show anything .
making possible to not opening the output pane for nothing is very useful .
extremely useful extension.

Anemy:
@alyahyaoui @MatthewS2077 Sorry it's so pestering. We've just release `0.4.0` which will only show the playground log output panel when a playground is run and there are logs to show - by default it won't open now. We can look into a setting to not open the log output by default, but I think this release should at least fix the ux annoyances. Thanks for using the extension and helping us improve it! :D

MatthewS2077:
@Anemy excellent! works great now! thanks for your efforts. 👍

PR #233, opened 2021-01-12
fix(playground): Only open playground output when there are output lines

Anemy:
This PR makes it so we only show the playground output panel when there are output lines to show (by removing the places we open it without output).
Addresses part of https://github.com/mongodb-js/vscode/issues/223

Here's where it's shown: https://github.com/mongodb-js/vscode/blob/master/src/editors/playgroundController.ts#L423


Anemy:
Fixed in https://github.com/mongodb-js/vscode/pull/234

PR #234, opened 2021-01-13
VSCODE-225: Associate results with connection

alenakhineika:
## Description

Associate playground results with the connection id which generated those results, so a user can't try to open up a document from playground results from a different connection.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


src/editors/playgroundController.ts
```diff
@@ -412,18 +395,26 @@ export default class PlaygroundController {
...
+
     const evaluateResponse: ExecuteAllResult = await this.evaluateWithCancelModal();
 
-    this._outputChannel.clear();
-    if (evaluateResponse.outputLines) {
+    if (
+      evaluateResponse &&
+      evaluateResponse.outputLines &&
+      evaluateResponse.outputLines.length > 0
+    ) {
```
addaleax, line 406: You should be able to do something like 
```suggestion
    if (evaluateResponse?.outputLines?.length) {
```
here :)

src/mdbExtensionController.ts
```diff
@@ -289,21 +299,24 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+          return Promise.resolve(false);
         }
 
         if (this._connectionController.isConnecting()) {
           vscode.window.showErrorMessage(
             'Unable to add database: currently connecting.'
           );
-          return false;
+
+          return Promise.resolve(false);
```
addaleax, line 319: I don't think you need to do this, since this is an `async` function (although there's nothing wrong with it of course)

addaleax (approved):
👍

src/editors/editorsController.ts
```diff
@@ -110,10 +93,15 @@ export default class EditorsController {
...
-  async openMongoDBDocument(data: {
-    documentId: EJSON.SerializableTypes;
-    namespace: string;
-  }): Promise<boolean> {
+  _openMongoDBDocumentFailed(text: string): Promise<boolean> {
+    const message = `Unable to open mongodb document: ${text}`;
+
+    vscode.window.showErrorMessage(message);
+
+    return Promise.resolve(false);
```
Anemy, line 101: Maybe this can be a `void` return type?

Anemy: Same with `_saveMongoDBDocumentFailed`

src/mdbExtensionController.ts
```diff
@@ -71,37 +77,50 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+      this._telemetryController,
+      this._playgroundResultViewProvider,
+      this._activeConnectionCodeLensProvider,
+      this._partialExecutionCodeLensProvider
     );
     this._webviewController = new WebviewController(
       this._connectionController,
       this._telemetryController
     );
+    this._editorsController.registerProviders();
```
Anemy, line 112: Not related to this ticket, I wonder if we should move calling the `activate` method into the constructor, so that the `extension.js` doesn't create the `mdbExtensionController` and then immediately activate it. Maybe we're doing extra steps there.

src/mdbExtensionController.ts
```diff
@@ -289,21 +299,24 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+          return Promise.resolve(false);
         }
 
         if (this._connectionController.isDisconnecting()) {
           vscode.window.showErrorMessage(
             'Unable to add database: currently disconnecting.'
           );
-          return false;
+
+          return Promise.resolve(false);
```
Anemy: Why do we wrap the false with the `Promise.resolve(` here and in a few other places?

Anemy (approved):
lgtm - one or two small comments. I think we might also be double wrapping the error message here:
Screen Shot 2021-01-13 at 3 53 18 PM

src/editors/editorsController.ts
alenakhineika: Hehe nice catch!

src/mdbExtensionController.ts
alenakhineika: The `registerProviders` function just wraps the registration of all providers from the constructor, where they were placed before. Moving `registerProviders to the `activate` function causes an error about trying to register an already registered scheme in tests because in tests we do this:

https://github.com/mongodb-js/vscode/blob/master/src/test/suite/index.ts#L45

in addition to this (that is also being called):

https://github.com/mongodb-js/vscode/blob/master/src/extension.ts#L41

Anemy (approved):
👍

PR #235, opened 2021-01-14
VSCODE-224: Telemetry document edited from playground

alenakhineika:
## Description

Track telemetry events when user decides to edit a document from a code lens and when the edited document is saved to MongoDB.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/instanceInfoResultType.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+  },
+  dataLake: {
+    isDataLake: boolean;
+  },
+  genuineMongoDB: {
+    isGenuine: boolean;
+    dbType: string;
+  }
+};
```
Anemy, line 25: Another way we could do this is add a typescript definitions file to the data-service itself. While this does help our code it doesn't ensure this can't change in the data-service module. Not too much we can do there though.
I think using the data-service's instance function in our telemetry is a bit overkill - it fetches all of the databases and collection info for example, which we don't use/need. Maybe at some point we update it to be a bit more like `mongosh` where we're only pulling what we need: https://github.com/mongodb-js/mongosh/blob/master/packages/service-provider-server/src/cli-service-provider.ts#L234 . Not related to this pr but I figure it's good to bring up.

src/telemetry/telemetryService.ts
```diff
@@ -79,66 +90,67 @@ export enum TelemetryEventTypes {
...
-        './constants.json'
-      );
-      const constants = fs.readFileSync(segmentKeyFileLocation);
-
-      this._segmentKey = JSON.parse(constants)?.segmentKey;
-      log.info('TELEMETRY key received', typeof this._segmentKey);
-    } catch (error) {
-      log.error('TELEMETRY key error', error);
-    }
+    this._segmentUserID = storageController.getUserID() || '';
```
Anemy, line 114: Isn't `getUserID` always returning a string?

src/telemetry/telemetryService.ts
```diff
@@ -352,4 +368,8 @@ export default class TelemetryController {
   trackDocumentUpdated(source: string, success: boolean): void {
     this.track(TelemetryEventTypes.DOCUMENT_UPDATED, { source, success });
   }
+
+  trackOpenMongoDBDocumentFromPlayground(source: string): void {
```
Anemy, line 372: Maybe we can make `source` here use the `DOCUMENT_SOURCE_PLAYGROUND` and `DOCUMENT_SOURCE_TREEVIEW` types

src/telemetry/telemetryService.ts
```diff
@@ -253,22 +267,24 @@ export default class TelemetryController {
     dataService: DataServiceType,
     connectionType: ConnectionTypes
   ): void {
-    dataService.instance({}, async (error: any, data: any) => {
```
Anemy, line 256: Since we are doing a lot of things depending on the instance results here - maybe we can promisify this and try catch it. Then if we do end up doing something that breaks it's handled well - maybe we also want to log that error to telemetry.

Anemy (commented):
nice - good tests too. Couple small comments, not blockers, nice renaming on the telemetry too.

src/instanceInfoResultType.ts
alenakhineika: Yeah, would be nice to have types in data service. But since we were talking about getting rid of data service completely for this project, I put a temporary solution here to fix the linting warning.

src/telemetry/telemetryService.ts
```diff
@@ -304,7 +305,9 @@ export default class TelemetryService {
 
         this.track(TelemetryEventTypes.NEW_CONNECTION, preparedProperties);
       }
-    });
+    } catch (error) {
+      log.error('TELEMETRY track new connection', error);
```
Anemy, line 309: Would it make sense to track the telemetry error here? (We'd probably want to sanitize the error somehow though)

alenakhineika: Max hasn't requested a telemetry event for failed connections, only for ones that succeeded.

Anemy: Gotcha - I was thinking about creating a telemetry event around letting us know we've missed tracking a connection because we errored parsing the connection instance information, not an event around a connection failing to connect. It's probably something we'd want a bug tracker to report, but since we don't have one it might make sense to report it to telemetry so we're aware if users are adding connections but we aren't properly tracking them.

alenakhineika: Good idea, pls create a ticket to triage it!

Anemy (approved):
👍

PR #240, opened 2021-01-20
feat(tree-view): add icons for connect and playground panel actions VSCODE-179

Anemy:
VSCODE-179

This PR adds icons to the connections and playgrounds tree panels.

updated icons


Anemy:
Synced w/ Claudia - will use plus icon from codicon instead of the circle plus we have.

alenakhineika (approved):
👍

PR #246, opened 2021-01-26
chore(telemetry): Add extension version to connected event VSCODE-126

Anemy:
VSCODE-126

Adds the extension version to the connection telemetry event.
Also adds a test w/ a real database for instance detail since I think we didn't have one before.


alenakhineika (approved):
👍

PR #250, opened 2021-02-02
New link for Atlas registration page

mmarcon:
With this new link for Atlas registration page we take users to a more contextual, VS Code-related experience.

image


mmarcon:
Page is not live yet, please check if it is before merging. Taking the PR out of draft as I am not sure if only the author can do that.

Anemy (approved):

PR #267, opened 2021-02-22
Autocomplete details

mmarcon:
Given that we are pulling the autocomplete symbols from the shell API, we can pull the documentation details from there too.

Feb-22-2021 08-19-58

## Description

This change pulls in the shell's `i18n` package and uses that to enrich the shell `CompletionItem`s with an example and a one-sentence explanation of what the command does. Unfortunately it does not look like the CompletionItem API supports links.

### Checklist

## Motivation and Context
- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions
That string that is passed to `Translator.translate()` is overly verbose but there does not seem to be a way around that.

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1792,6 +1792,17 @@
         "pretty-repl": "^2.3.1",
         "semver": "^7.1.2",
         "text-table": "^0.2.0"
+      },
+      "dependencies": {
+        "@mongosh/i18n": {
+          "version": "0.6.1",
+          "resolved": "https://registry.npmjs.org/@mongosh/i18n/-/i18n-0.6.1.tgz",
+          "integrity": "sha512-8Jsf1nKGNwLuQkM/WYFgKbr9Zvaqg6YBujuYe1rzKeHN51ihKTTPpI4U5Ytl+1HsgMbjTcNegpsz8EsDg5xp/Q==",
+          "requires": {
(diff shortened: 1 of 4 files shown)


mmarcon:
As a side note, if we had a flag for deprecated methods that the extension can read when it loads the shell API symbols, we could tag those `CompletionItem`s as deprecated and I believe VS Code will then render them with a strikethrough font.

rose-m (commented):
Awesome!! 😍 LGTM - but I'll let the others give their full approval!

mcasimir:
Wow 😄!

Would it be possible to move this to be part of the interface of the shell autocompleter? Is some change more, but if this is not urgent it would be good so we don't leak shell internals around and we'd get the same in all the shells.

mcasimir:
Never mind, forgot that the shell auto-completer works in a different way

gribnoysup (approved):
Looks good!

gribnoysup:
RE failing to update mongosh deps to latest, it's probably related to https://github.com/mongodb-js/vscode/pull/219 and for sure a few other things, there were a bunch of changes in latest mongosh and we definitely should take some time and update vscode to latest

mmarcon:
image

PR #270, opened 2021-03-02
VSCODE-227: Use a proper command format to launch shell in bash for ssl

alenakhineika:
## Description

Use a proper command format to launch shell in bash for ssl.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,7 +10,7 @@ trigger:
 strategy:
   matrix:
     linux:
-      imageName: 'ubuntu-latest'
+      imageName: 'ubuntu-18.04'
     mac:
       imageName: 'macos-latest'
     windows:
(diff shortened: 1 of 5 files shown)


azure-pipelines.yml
```diff
@@ -10,7 +10,7 @@ trigger:
 strategy:
   matrix:
     linux:
-      imageName: 'ubuntu-latest'
+      imageName: 'ubuntu-18.04'
```
alenakhineika, line 13: Mongodb url can't be resolved for the latest version of ubuntu, we should temporarily use ubuntu-18.04 until https://github.com/mongodb-js/download-url/ will be updated to fix the problem.

Issue #273, opened 2021-03-09
Does not return all data from a query


ssnyder:
#### Current Behavior
Querying a collection to gather a count does not return all rows of results in the output view (on VSCode using latest version of extension).  This was verified by running the same query in Robo3T and the last record was missing from VSCode return.
#### Code/Gist
I am trying to count all of my documents, grouping by the payload.annotationLanguage value.  All records contain this field with a String value. 

db.bestdb.aggregate([{
    $group: {
        _id: {
            "lang": "$payload.annotationLanguage"
        },
        "count": {$sum:1}
    } 
},{
    $project: {
        "lang": "$_id.lang",
        "count": "$count",
        "_id": 0
    }
}]);

#### Expected Behavior/Code

I am expecting 21 records returned, but the last record is not returned by VSCode extension.  Running the same query in Robo3T provides all records expected.

#### Environment
- Connection to Mongo 3.2 server using admin user for authentication
- VSCode Help-About:

> Version: 1.53.2 (system setup)
> Commit: 622cb03f7e070a9670c94bae1a45d78d7181fbd4
> Date: 2021-02-11T11:48:04.245Z
> Electron: 11.2.1
> Chrome: 87.0.4280.141
> Node.js: 12.18.3
> V8: 8.7.220.31-electron.0
> OS: Windows_NT x64 10.0.19042
>

Anemy:
Hi @ssnyder this actually works as intended, currently unless `.toArray()` is specified we [open a cursor and fetch the first 20 documents, similar to the mongo shell](https://docs.mongodb.com/manual/tutorial/iterate-a-cursor/). Thanks for reporting, I think we can probably improve how we show cursor playground results to users to help illustrate what they're seeing.

Does running 
```javascript
db.bestdb.aggregate([{
$group: {
_id: {
"lang": "$payload.annotationLanguage"
},
"count": {$sum:1}
}
},{
$project: {
"lang": "$_id.lang",
"count": "$count",
"_id": 0
}
}]).toArray();
```
Show all 21 results?

ssnyder:
This does work thank you for the quick response.  I must have missed that in the docs.  

I have set the Mdb: default limit to 1000 in the settings.  This does not seem to be honored?  Also, setting the batch size (DBQuery.shellBatchSize=300) in the script being run (after the use, but before the aggregate) is also not being honored?  

I would appreciate more information on this in the docs and the playground default template as well.

duard:
[...]

PR #274, opened 2021-03-10
fix: update playground template to add a comment for .toArray to aggregation run

Anemy:
Updates the playground template aggregation to have a comment for `.toArray` and explicitly note that a cursor is currently created. Thinking it might help cases like https://github.com/mongodb-js/vscode/issues/273


addaleax (approved):
👍

gribnoysup (approved):

PR #281, opened 2021-03-30
chore(telemetry): don't use Compass' data-service instance detail fetching VSCODE-235 

Anemy:
VSCODE-235

Previously we were using the [mongodb-data-service](https://github.com/mongodb-js/data-service) `instance` method to fetch details about a connection for telemetry. This PR makes it so we fetch our own information. This allows us to avoid the extra calls and processing the [instance fetching was performing](https://github.com/mongodb-js/data-service/blob/master/lib/instance-detail-helper.js#L603) which we didn't need and reduces our dependency on Compass' data-service. 

Screen Shot 2021-03-30 at 9 31 07 AM


src/connectionController.ts
```diff
@@ -248,10 +248,12 @@ export default class ConnectionController {
 
   public sendTelemetry(
     newDataService: DataServiceType,
+    connectionModel: ConnectionModel,
     connectionType: ConnectionTypes
   ): void {
     this._telemetryService.trackNewConnection(
-      newDataService,
+      newDataService.client.client,
```
alenakhineika, line 255: Maybe we can think in a feature about a better naming instead `client.client`. I saw it before and it every time confuses me :)

alenakhineika (approved):
LGTM!

Issue #222, opened 2020-12-05
Feature Request - Option to switch focus to playground result


MatthewS2077:
## Feature Request

Option to switch focus to playground result.

## Detailed Description

Add option _Switch focus to Playground Result_ which defaults to true (the current behaviour). 

If false, then running a query will not switch focus to the Playground Result but leave focus in the Playground.

## Context

It's annoying to have focus switch to Playground Result.  I prefer focus to stay where it is.

Anemy:
Hi @MatthewS2077 thanks for the good feature request. Created a jira ticket to track progress: https://jira.mongodb.org/browse/VSCODE-214

MatthewS2077:
Any news on this?  The extension is great but switching focus to the results pane automatically is a bad UX,  IMHO 🤷‍♂️

For my setup, my results pane is to the right of my editor pane, so I can see both at once. If I'm testing aggregations I make a quick change, run, check result... but focus is switched to the output pane and I need to switch focus back to my editor pane to make another change... this is very annoying 🤬.   Otherwise this extension is 👍👍👍

alenakhineika:
Hey, the team is going to triage the ticket and you can track progress here or in Jira: https://jira.mongodb.org/browse/VSCODE-214

alenakhineika:
Hey, `0.6.0` is live with an update to preserve focus in playgrounds after running them! Closing this issue, feel free to reopen if you find any issues with it.

PR #290, opened 2021-04-26
fix: save big files without error and preserve playground focus

alenakhineika:
## Description

1) Preserve focus after running playgrounds to be able quickly to make additional changes and run the playground again: https://jira.mongodb.org/browse/VSCODE-214

- Related issue: https://github.com/mongodb-js/vscode/issues/222
- Note: We made a decision to preserve focus by default and do not make it configurable in settings.

2) For large files VSCode turns off tokenization, wrapping, and folding in order to reduce memory usage and avoid freezing or crashing. This causes situations when the file is open, but the `vscode.window.activeTextEditor` is `undefined`. This PR removes throwing an error when there is no active editor and keeps the default behavior for such files, that is saving files to disc: https://jira.mongodb.org/browse/VSCODE-250

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/editors/editorsController.ts
+++ b/src/editors/editorsController.ts
@@ -158,7 +158,7 @@ export default class EditorsController {
     const activeEditor = vscode.window.activeTextEditor;
 
     if (!activeEditor) {
-      vscode.window.showErrorMessage('Unable to save mongodb document: The active editor cannot be found');
+      await vscode.commands.executeCommand('workbench.action.files.save');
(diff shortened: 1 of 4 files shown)


gribnoysup (approved):
Looks good to me!

PR #294, opened 2021-05-27
feat: create dbs and collections using playgrounds VSCODE-262

alenakhineika:
## Description

Change the way on how we create dbs and collections. Instead of using input boxes for db and collection names open a new playground template.

We should use one template for db and collection creation since there is no way to create db without a collection
The template should be a valid JS template with placeholders for db and collection names
If we create a collection for the existing db the template should pre-populate the db name
Show a time-series prototype form only for Server 5.0 and later.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/editors/playgroundController.ts
```diff
@@ -237,6 +244,36 @@ export default class PlaygroundController {
...
+    let content = playgroundCreateCollectionTemplate;
+
+    if (dataService) {
+      const adminDb = dataService.client.client.db('admin');
+      const buildInfo = await adminDb.command({ buildInfo: 1 });
+
+      if (buildInfo) {
+        const serverVersion = buildInfo.version.replace(/-.*$/, '');
+
+        if (semver.satisfies(serverVersion, '>= 5.0')) {
```
addaleax, line 260:
```suggestion
        const serverVersion = buildInfo.versionArray[0];

        if (serverVersion >= 5) {
```

since `versionArray` is the machine-accessible variant of `version`?

addaleax (approved):

src/editors/playgroundController.ts
alenakhineika: Replaced version with versionArray.

addaleax (approved):
Looks great!

PR #301, opened 2021-06-16
feat: bump mongosh APIs to latest VSCODE-240

addaleax:
Also includes the change from
https://github.com/mongodb-js/vscode/pull/219
and bumps mongodb-connection-model to a version that depends on the
3.x driver directly instead of using it as a peer dependency,
and minor drive-by things I’ve noticed (use `mongodb` itself only
for typed imports, use `assert.strictEqual` where it gives a better
error message, adjust types in preparation for a 4.0 driver update).

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


Anemy:
Changes look good, but I'm getting an error when running playgrounds:
Screen Shot 2021-06-16 at 8 11 42 PM

Looks like this is returning `null` https://github.com/mongodb-js/vscode/blob/master/src/editors/playgroundController.ts#L188
Maybe something changed with how the data service is providing the connection options.  Can look into more tmrw. Surprised tests didn't pick this up, maybe something we can cover.

gribnoysup:
Yeah, there was a breaking change awhile ago in connection-model https://github.com/mongodb-js/connection-model/pull/352 and a corresponding one in data-service https://github.com/mongodb-js/data-service/pull/308, so you might want to update data-service also to work with latest connection-model correctly

addaleax:
→ https://github.com/mongodb-js/compass/pull/2261

Anemy:
Getting this error now when running playgrounds:
Screen Shot 2021-06-17 at 11 19 57 AM


Maybe we need to do something like what this adapter is doing: https://github.com/mongodb-js/compass/blob/master/packages/compass-shell/src/modules/adapt-driver-v36-connection-params.js#L26

addaleax:
Hm yeah ... I'm not understanding why this is not happening for me, then

But if the 4.0 bump in Compass is a relatively quick thing maybe it's worth not trying to work around in vscode and just waiting for the Compass bump to happen first

Anemy (approved):
lgtm! nice updates!

PR #311, opened 2021-07-13
feat: add support for loadbalanced connections

mcasimir:
Updates `@mongosh/*`, `mongodb-connection-model`, `mongodb-data-service` and `mongodb` to add support for loadbalanced connections.

Screenshot 2021-07-13 at 10 44 30

Screenshot 2021-07-13 at 16 02 20

Other: 
- update `depcheck` since i was running into this: https://github.com/depcheck/depcheck/issues/610
- add an `.npmrc` file to always set `legacy-peer-deps=true`
- move to workflows as planned for https://jira.mongodb.org/browse/VSCODE-272 since azure pipeline was failing
- enable cache for eslint

PR #306, opened 2021-06-25
Make `mongosh` the default shell

mmarcon:
## Description

As we approach the GA of the new MongoDB Shell (`mongosh`), let's make it the default.

### Checklist
- [ ] New tests and/or benchmarks are included
- [x] Documentation is changed or added

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*
- [ ] Patch (non-breaking change which fixes an issue)
- [x] Minor (non-breaking change which adds functionality)
- [ ] Major (fix or feature that would cause existing functionality to change)

PR #298, opened 2021-06-08
feat: system collections are last in the list

mmarcon:
## Description

To separate system collections from regular collections, we display them last in the list.

image

This PR also includes a minor tweak to the Terraform snippet as one line was not properly indented.

### Checklist
- [x] New tests and/or benchmarks are included
- [ ] Documentation is changed or added

## Motivation and Context
- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*
- [x] Patch (non-breaking change which fixes an issue)
- [ ] Minor (non-breaking change which adds functionality)
- [ ] Major (fix or feature that would cause existing functionality to change)

PR #297, opened 2021-06-04
Update atlas-terraform.json

themantissa:
## Description
Example terraform snippet was quite out of date.  
-update versions
-update ip whitelist to access list
-update provider block
-remove disk size
-update connection string output

### Checklist
- [ ] New tests and/or benchmarks are included
- [ ] Documentation is changed or added

## Motivation and Context
- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [X] Misc

## Types of changes

- [ ] *Backport Needed*
- [X] Patch (non-breaking change which fixes an issue)
- [ ] Minor (non-breaking change which adds functionality)
- [ ] Major (fix or feature that would cause existing functionality to change)

PR #313, opened 2021-07-13
0.6.0

Anemy:
Changelog update:

### Added

- Added icons for time-series collections (VSCODE-263, #296)
- Updated our mongosh dependency to 1.0.0 and MongoDB node driver to a fork of the 4.0.0 driver (VSCODE-264, #308)
- Added support for load balanced connections (#311)

### Changed

- Made mongosh the default shell when launching the shell from a saved connection (#306)
- Updated how collections and databases are created to use playgrounds instead of vscode inputs (VSCODE-262, #294)
- Updated how playgrounds are run to preserve focus on the editor when the results are shown (VSCODE-214, #290, #222)
- Updated the Atlas terraform snippet (#297)
- Updated the collections list to sort system collections to the end of the list (#298)

### Fixed

- Fixed showing an error when large files are shown in vscode which have tokenization disabled (VSCODE-250, #290)
- Fixed showing recently dropped collections and databases in the explorer (VSCODE-244, #288)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "name": "mongodb-vscode",
   "displayName": "MongoDB for VS Code",
   "description": "Connect to MongoDB and Atlas directly from your VS Code environment, navigate your databases and collections, inspect your schema and use playgrounds to prototype queries and aggregations.",
-  "version": "0.5.1-dev.0",
+  "version": "0.6.0",
   "homepage": "https://github.com/mongodb-js/vscode",
   "qna": "https://developer.mongodb.com/community/forums/",
   "repository": {
--- a/src/test/suite/connectionController.test.ts
+++ b/src/test/suite/connectionController.test.ts
@@ -512,7 +512,7 @@ suite('Connection Controller Test Suite', function () {
 
   test('"getConnectionStringFromConnectionId" returns the driver uri of a connection', async () => {
     const expectedDriverUri =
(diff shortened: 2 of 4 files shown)


src/test/suite/connectionController.test.ts
```diff
@@ -512,7 +512,7 @@ suite('Connection Controller Test Suite', function () {
 
   test('"getConnectionStringFromConnectionId" returns the driver uri of a connection', async () => {
     const expectedDriverUri =
-      'mongodb://localhost:27018/?readPreference=primary&appname=mongodb-vscode%200.5.1-dev.0&directConnection=true&ssl=false';
```
mcasimir, line 515: Can we read the version from package.json? A separate PR is also ok if you want to go with this!

package.json
```diff
@@ -3,7 +3,7 @@
   "name": "mongodb-vscode",
   "displayName": "MongoDB for VS Code",
   "description": "Connect to MongoDB and Atlas directly from your VS Code environment, navigate your databases and collections, inspect your schema and use playgrounds to prototype queries and aggregations.",
-  "version": "0.5.1-dev.0",
```
mcasimir, line 6: Just a thought for later .. we could automate also this, is pretty simple .. although it would be even better to just sick with `0.0.0-dev-0` if that is only cosmetic.

mcasimir (approved):

src/test/suite/connectionController.test.ts
Anemy: Yeah let's grab in a separate pr hah

PR #315, opened 2021-07-14
chore: automate release VSCODE-116

mcasimir:
Automate the release of the extension

When the build is triggered with a tag `v*.*.*`:
- The extension `CHANGELOG.md`, the one visible in the Settings panel, is replaced with a link to the Github release
- The `package.json` and `package-lock.json` versions are set to the tag version
- A draft release is built with the `vsix` bundle attached

When the release is published:
- The `vsix` asset is downloaded from the release and uploaded to the vscode marketplace 
- The `CHANGELOG.md` in the root of the repo is regenerated and pushed to main

### DIFF
--- a/.github/workflows/publish-release.yaml
+++ b/.github/workflows/publish-release.yaml
@@ -1,0 +1,41 @@
+name: Publish Release
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  publish-release:
+    name: Publish Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Node.js Environment
+        uses: actions/setup-node@v2.1.2
+        with:
+          node-version: ^12.9.0
+
+      - name: Install npm@7
+        run: npm install -g npm@7
+
+      - name: Download release assets
+        run: |
+          filename=$(basename -- "$ASSET_URL")
+          extension="${filename##*.}"
+
+          if [[ ! $extension != "vsix" ]]; then
+            echo "Wrong asset extension ${extension} in ${ASSET_URL}"
+            exit 1
+          else
+            curl "${ASSET_URL}" > ${filename}
+            ls -al ${filename}
+            echo "VSIX_PATH=${filename}" >> $GITHUB_ENV
+          fi
+        env:
+          ASSET_URL: ${{ github.event.release.assets[0].browser_download_url }}
+
+      - name: Publish to marketplace
+        run: |
+          echo "Publishing ${VSIX_PATH} to marketplace"
+          # npx vsce publish --packagePath ${VSIX_PATH}
+        env:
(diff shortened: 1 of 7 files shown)


.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,41 @@
...
+            ls -al ${filename}
+            echo "VSIX_PATH=${filename}" >> $GITHUB_ENV
+          fi
+        env:
+          ASSET_URL: ${{ github.event.release.assets[0].browser_download_url }}
+
+      - name: Publish to marketplace
+        run: |
+          echo "Publishing ${VSIX_PATH} to marketplace"
+          # npx vsce publish --packagePath ${VSIX_PATH}
```
mcasimir, line 39: @Anemy would be great to try this out manually once before enabling it

.github/workflows/rebuild-changelog.yaml
```diff
@@ -0,0 +1,58 @@
...
+            echo "## [$name]($url) - $date" >> CHANGELOG.md
+            echo "" >> CHANGELOG.md
+            echo "$body" >> CHANGELOG.md
+            echo "" >> CHANGELOG.md
+            echo "" >> CHANGELOG.md
+          done
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Commit CHANGELOG.md changes
```
mcasimir, line 50: @Anemy let me know if you want we can update the version in the `package.json` like this:

```
# release view without a tag == latest release
lastReleaseVersion=$(gh release view --json=name | jq -r .name)
npx json -I -f package.json -e "this.version='${lastReleaseVersion}-dev.0'"
```

src/test/suite/connectionController.test.ts
```diff
@@ -512,7 +512,7 @@ suite('Connection Controller Test Suite', function () {
 
   test('"getConnectionStringFromConnectionId" returns the driver uri of a connection', async () => {
     const expectedDriverUri =
-      'mongodb://localhost:27018/?readPreference=primary&appname=mongodb-vscode%200.6.1-dev.0&directConnection=true&ssl=false';
+      'mongodb://localhost:27018/?readPreference=primary&appname=mongodb-vscode%200.0.0-dev.0&directConnection=true&ssl=false';
```
mcasimir, line 515: If we auto-update the dev version in `package.json` we also have to update this test

.github/workflows/rebuild-changelog.yaml
Anemy: I think it's fine as 0.0.0 - also makes us do something similar to what we're doing in mongosh.

Anemy (approved):
lgtm! 🔥

PR #318, opened 2021-07-16
VSCODE-247: Replace code lenses with code actions

alenakhineika:
## Description

Based on user feedback, we are going to resolve the annoyance of the code lenses to run portions of a playground that make the code jump up and down when the selection changes, with code actions.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions

At design sync let's discuss if this is ok to have a `Show Fixes` tooltip for the bulb since we run the code and do not fix anything. I didn't find an API to change this tooltip. You can change a tooltip for the action itself, but not for the bulb.

Screenshot 2021-07-14 at 15 21 38

UPD: As a temporary improvement we [moved code lenses to the bottom of the selected text](https://github.com/mongodb-js/vscode/pull/324) and then after discussion with the team at the design sync we decided to proceed with code actions solution. We might use the bulb in the future for other features that could fit better to the `Show Fixes` term.

## Types of changes

- [ ] *Backport Needed*


src/test/fixture/.vscode/settings.json
```diff
@@ -18,5 +18,6 @@
     "other": true,
     "comments": false,
     "strings": true
-  }
+  },
+  "security.workspace.trust.enabled": false
```
addaleax, line 22: This is an unrelated change, right?

src/test/suite/editors/codeActionProvider.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  test('expected provideCodeActions to return undefined when text is not selected', () => {
+    const testCodeActionProvider = new CodeActionProvider(testPlaygroundController);
+    const codeActions = testCodeActionProvider.provideCodeActions();
+
+    expect(codeActions).to.be.undefined;
+  });
+
+  test('expected provideCodeActions to return a run selected playground blocks action', () => {
```
addaleax, line 74: Would it be possible to add a test that actually runs the selected code? As far as I can tell, there wasn’t any before, but it might be a nice thing to have here

src/test/fixture/.vscode/settings.json
alenakhineika: Vscode 1.57 introduced a trusted authors confirmation modal, that interrupts tests, so this line disables this check during tests running.

PR #323, opened 2021-07-21
fix(ci): fix release github actions

Anemy:
Fixes a couple issues on building/releasing/building changelog. 
Updated our release version title to have a `v` in the front, this is mostly since our previous release titles are named like that. I read somewhere it's a recommended practice, but I think it could probably go either way - any preferences?

### DIFF
--- a/.github/workflows/publish-release.yaml
+++ b/.github/workflows/publish-release.yaml
@@ -22,7 +22,7 @@ jobs:
           filename=$(basename -- "$ASSET_URL")
           extension="${filename##*.}"
 
-          if [[ ! $extension != "vsix" ]]; then
+          if [[ $extension != "vsix" ]]; then
             echo "Wrong asset extension ${extension} in ${ASSET_URL}"
             exit 1
           else
--- a/.github/workflows/rebuild-changelog.yaml
+++ b/.github/workflows/rebuild-changelog.yaml
@@ -28,24 +28,21 @@ jobs:
           echo "# Change Log" > CHANGELOG.md
           echo "" >> CHANGELOG.md
 
-          gh release list | while read line; do
-            line=${line/Latest/" "}
-            tagName=$(echo $line | awk '{ print $2 }')
-            echo "line: $line --- tag: $tagName"
+          tags=$(gh api repos/mongodb-js/vscode/releases | jq -r .[].tag_name | grep -v internal | grep -v pre | grep -v beta)
 
-            if [ "$tagName" != "Pre-release" ] && [ "$tagName" != "Preview" ]; then
+          # NOTE: here the quotes around $tags are necessary
+          echo "$tags" | while read tagName; do
 
-              json=$(gh release view $tagName --json=name,publishedAt,url,body)
-              url=$(echo $json | jq -r .url)
-              name=$(echo $json | jq -r .name)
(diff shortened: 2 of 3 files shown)


.github/workflows/rebuild-changelog.yaml
```diff
@@ -33,16 +33,19 @@ jobs:
...
-            url=$(echo $json | jq -r .url)
-            name=$(echo $json | jq -r .name)
-            date=$(echo $json | jq -r .publishedAt | cut -f1 -dT)
-            body=$(echo $json | jq -r .body)
-            echo "## [$name]($url) - $date" >> CHANGELOG.md
-            echo "" >> CHANGELOG.md
-            echo "$body" >> CHANGELOG.md
-            echo "" >> CHANGELOG.md
-            echo "" >> CHANGELOG.md
+            if [ "$tagName" != "Pre-release" ] && [ "$tagName" != "Preview" ]; then
```
Anemy, line 36: We were getting `release not found` with the github cli for releases with these tag names. I think we probably don't want to include them in the changelog anyway.

.github/workflows/test-and-build.yaml
```diff
@@ -71,7 +71,7 @@ jobs:
           echo "See full release notes at: https://github.com/mongodb-js/vscode/releases/tag/${RELEASE_TAG}\n" > CHANGELOG.md
           npx json -I -f package.json -e "this.version='${RELEASE_VERSION}'"
           npx json -I -f package-lock.json -e "this.version='${RELEASE_VERSION}'"
-        if: startsWith(github.ref, 'refs/tags/')
+        if: ${{ startsWith(github.ref, 'refs/tags/') && runner.os == 'Linux' }}
```
mcasimir, line 74: Like this you'll get a weird set of assets where for 2 of the platforms you would be able to download a different content from the UI, and i'm not sure that would be obvious.

If the issue was that the build didn't work on windows did you consider trying something like `shell: bash` or even an `if:  !startsWith(github.ref, 'refs/tags/')` on the task that upload assets to avoid any confusion?

.github/workflows/publish-release.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
           filename=$(basename -- "$ASSET_URL")
           extension="${filename##*.}"
 
-          if [[ ! $extension != "vsix" ]]; then
+          if [[ $extension != "vsix" ]]; then
```
mcasimir, line 25: 🤦‍♂️ i'm sorry about this 😄

mcasimir (commented):
Thanks for fixing this .. i did my best to try things out on a playground repo, but didn't manage to make it 100% the same as here.

Left a few comments i think is worth to address, beside those I think we can merge.

.github/workflows/test-and-build.yaml
Anemy: Ah good call, maybe we could have just one asset built since it should be totally platform independent. I'll add the `shell: bash`, we can always remove the extra platforms if they ever cause confusion.

.github/workflows/rebuild-changelog.yaml
Anemy: Updated those releases, also cleaned up their names. Removed that line 👍

mcasimir (approved):
Nice! thanks for the changes!

PR #324, opened 2021-07-22
feat(playgrounds): Show code lens at end of selection

Anemy:
Chatting and pairing on this w/ @alenakhineika - we were thinking until we decide if code actions are a better alternative we could go back to putting the code lenses at the bottom of selections just to hopefully improve that ux.

This PR has a few fixes for situations where we weren't showing codelenses, for instance if a user had a multi selection and the last selection was an empty space it wouldn't show a code lens.

In cases where the selection includes the end of the playground, and that last line is not empty, we now add an empty line so we can show a code lens nicely at the end. This is shown in the video below. I don't totally remember, but I think this maybe have been the reason we changed to putting code lenses on top.

https://user-images.githubusercontent.com/1791149/126701660-93b22e68-56d8-4c49-ade6-2ad54350e182.mp4

### DIFF
--- a/src/editors/partialExecutionCodeLensProvider.ts
+++ b/src/editors/partialExecutionCodeLensProvider.ts
@@ -1,6 +1,57 @@
 import * as vscode from 'vscode';
 import EXTENSION_COMMANDS from '../commands';
 
+// Returns a boolean if the selection is one that is valid to show a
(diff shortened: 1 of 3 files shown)


alenakhineika (approved):
LGTM!

PR #326, opened 2021-07-23
chore(deps): use node 14

Anemy:
Seems to work fine, and tests are passing. Maybe let's hold off on merging until we look into bumping in Compass and mongosh so we don't have to `nvm` all the time.


addaleax (approved):
:+1:

addaleax:
> mongosh

mongosh is already Node.js 14, so I have to nvm anyway :)

Anemy:
@addaleax ah cool let's just merge this then lol

PR #334, opened 2021-07-27
fix: add github actions badge, remove azure pipelines badge

Anemy:
Not sure if this is all that helpful to have, maybe we should remove it all together? Do y'all find them useful in projects?

Here's how it looks: [Build Status](https://github.com/mongodb-js/vscode/actions/workflows/test-and-build.yaml)


alenakhineika:
I like badges, but the thing i don't like about the build badge, that it can be often red because of the latest build, but in readme, it looks like master is failing :)

PR #338, opened 2021-08-09
refactor: add no-floating-promises rule

alenakhineika:
## Description

Moving toward async/await instead of using promises, it would be nice to have a tool that ensures proper call of async functions that do or do not expect awaiting the result. For example, forgetting `await` before `connect()` would definitely cause the connectivity issue. But we do not want to await telemetry events or other processes that could happen parallel to the main flow.

This PR adds `no-floating-promises` aslant rule that complains when await keyword is missing before async function call. If the function doesn't have to wait for a result, the void keyword should be used instead. It makes code safer and easier to read because you explicitly can see the expected behaviour.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -210,7 +210,7 @@ export default class ConnectionController {
     try {
       return this.addNewConnectionStringAndConnect(connectionString);
     } catch (error) {
-      vscode.window.showErrorMessage(error.message);
+      void vscode.window.showErrorMessage(error.message);
 
       return false;
     }
@@ -244,7 +244,7 @@ export default class ConnectionController {
     connectionModel: ConnectionModel,
     connectionType: ConnectionTypes
   ): void {
-    this._telemetryService.trackNewConnection(
+    void this._telemetryService.trackNewConnection(
       newDataService.client.client,
       connectionModel,
       connectionType
@@ -323,7 +323,7 @@ export default class ConnectionController {
         connectionId,
         connectionInfoAsString
       );
-      this._storageController.storeNewConnection(newLoadedConnection);
+      void this._storageController.storeNewConnection(newLoadedConnection);
     }
 
     return this.connect(connectionId, connectionModel, connectionType);
@@ -412,7 +412,7 @@ export default class ConnectionController {
     }
 
     log.info('Successfully connected');
-    vscode.window.showInformationMessage('MongoDB connection successful.');
+    void vscode.window.showInformationMessage('MongoDB connection successful.');
 
     this._activeDataService = newDataService;
     this._activeConnectionModel = connectionModel;
@@ -449,7 +449,7 @@ export default class ConnectionController {
           : savedConnectionModel
       );
     } catch (error) {
-      vscode.window.showErrorMessage(`Unable to load connection: ${error}`);
+      void vscode.window.showErrorMessage(`Unable to load connection: ${error}`);
 
       return false;
     }
@@ -463,7 +463,7 @@ export default class ConnectionController {
 
       return connectResult.successfullyConnected;
     } catch (error) {
-      vscode.window.showErrorMessage(error.message);
+      void vscode.window.showErrorMessage(error.message);
 
       return false;
     }
@@ -476,7 +476,7 @@ export default class ConnectionController {
     );
 
     if (!this._activeDataService) {
-      vscode.window.showErrorMessage(
+      void vscode.window.showErrorMessage(
         'Unable to disconnect: no active connection.'
       );
 
@@ -502,10 +502,10 @@ export default class ConnectionController {
 
     try {
       await _disconnect();
-      vscode.window.showInformationMessage('MongoDB disconnected.');
+      void vscode.window.showInformationMessage('MongoDB disconnected.');
     } catch (error) {
       // Show an error, however we still reset the active connection to free up the extension.
-      vscode.window.showErrorMessage(
+      void vscode.window.showErrorMessage(
         'An error occured while disconnecting from the current connection.'
       );
     }
@@ -533,7 +533,7 @@ export default class ConnectionController {
   async removeMongoDBConnection(connectionId: string): Promise<boolean> {
     if (!this._connections[connectionId]) {
       // No active connection(s) to remove.
-      vscode.window.showErrorMessage('Connection does not exist.');
+      void vscode.window.showErrorMessage('Connection does not exist.');
 
       return false;
     }
@@ -559,7 +559,7 @@ export default class ConnectionController {
 
     await this.removeSavedConnection(connectionId);
 
-    vscode.window.showInformationMessage('MongoDB connection removed.');
+    void vscode.window.showInformationMessage('MongoDB connection removed.');
 
     return true;
   }
@@ -571,7 +571,7 @@ export default class ConnectionController {
 
     if (connectionIds.length === 0) {
       // No active connection(s) to remove.
-      vscode.window.showErrorMessage('No connections to remove.');
+      void vscode.window.showErrorMessage('No connections to remove.');
 
       return false;
     }
--- a/src/editors/editorsController.ts
+++ b/src/editors/editorsController.ts
@@ -121,7 +121,7 @@ export default class EditorsController {
       )) as EJSON.SerializableTypes;
 
       if (mdbDocument === null) {
-        vscode.window.showErrorMessage(`
+        void vscode.window.showErrorMessage(`
           Unable to open mongodb document: document ${JSON.stringify(data.documentId)} not found
         `);
 
@@ -148,7 +148,7 @@ export default class EditorsController {
     } catch (error) {
       const printableError = error as { message: string };
 
-      vscode.window.showErrorMessage(printableError.message);
+      void vscode.window.showErrorMessage(printableError.message);
 
       return false;
     }
@@ -196,9 +196,9 @@ export default class EditorsController {
       });
 
       // Save document changes to active editor.
-      activeEditor?.document.save();
+      await activeEditor?.document.save();
 
-      vscode.window.showInformationMessage(
+      void vscode.window.showInformationMessage(
         `The document was saved successfully to '${namespace}'`
       );
 
@@ -206,7 +206,7 @@ export default class EditorsController {
     } catch (error) {
       const printableError = error as { message: string };
 
-      vscode.window.showErrorMessage(printableError.message);
+      void vscode.window.showErrorMessage(printableError.message);
 
       return false;
     }
@@ -251,7 +251,7 @@ export default class EditorsController {
     } catch (error) {
       const printableError = error as { message: string };
 
-      vscode.window.showErrorMessage(
+      void vscode.window.showErrorMessage(
         `Unable to open documents: ${printableError.message}`
       );
 
@@ -272,7 +272,7 @@ export default class EditorsController {
       this._collectionDocumentsOperationsStore.operations[operationId]
         .isCurrentlyFetchingMoreDocuments
     ) {
-      vscode.window.showErrorMessage(
+      void vscode.window.showErrorMessage(
         'Already fetching more documents...'
       );
 
@@ -281,7 +281,7 @@ export default class EditorsController {
 
     // Ensure we're still connected to the correct connection.
     if (connectionId !== this._connectionController.getActiveConnectionId()) {
-      vscode.window.showErrorMessage(
+      void vscode.window.showErrorMessage(
         `Unable to view more documents: no longer connected to ${connectionId}`
       );
 
--- a/src/explorer/explorerTreeController.ts
+++ b/src/explorer/explorerTreeController.ts
@@ -38,7 +38,7 @@ implements vscode.TreeDataProvider<vscode.TreeItem> {
     this._connectionController.removeEventListener(
       DataServiceEventTypes.CONNECTIONS_DID_CHANGE,
       () => {
-        void this.refresh();
+        this.refresh();
       }
     );
   }
--- a/src/explorer/explorerTreeController.ts
+++ b/src/explorer/explorerTreeController.ts
@@ -27,7 +27,7 @@ implements vscode.TreeDataProvider<vscode.TreeItem> {
     this._connectionController.addEventListener(
       DataServiceEventTypes.CONNECTIONS_DID_CHANGE,
       () => {
-        this.refresh();
+        void this.refresh();
       }
     );
 
@@ -38,7 +38,7 @@ implements vscode.TreeDataProvider<vscode.TreeItem> {
     this._connectionController.removeEventListener(
       DataServiceEventTypes.CONNECTIONS_DID_CHANGE,
       () => {
-        this.refresh();
+        void this.refresh();
       }
     );
   }
@@ -124,10 +124,10 @@ implements vscode.TreeDataProvider<vscode.TreeItem> {
   _onDidChangeTreeData: vscode.EventEmitter<any>;
   readonly onDidChangeTreeData: vscode.Event<any>;
 
-  public refresh = (): Promise<boolean> => {
+  refresh = (): boolean => {
     this._onDidChangeTreeData.fire(null);
 
-    return Promise.resolve(true);
+    return true;
   };
 
   onTreeItemUpdate(): void {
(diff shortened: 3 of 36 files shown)


src/editors/editorsController.ts
```diff
@@ -196,17 +196,17 @@ export default class EditorsController {
       });
 
       // Save document changes to active editor.
-      activeEditor?.document.save();
+      await activeEditor?.document.save();
```
alenakhineika, line 199: Looks like here we could wait for saving before showing a successful message modal.

src/explorer/explorerTreeController.ts
```diff
@@ -124,10 +124,10 @@ implements vscode.TreeDataProvider<vscode.TreeItem> {
   _onDidChangeTreeData: vscode.EventEmitter<any>;
   readonly onDidChangeTreeData: vscode.Event<any>;
 
-  public refresh = (): Promise<boolean> => {
+  refresh = (): boolean => {
     this._onDidChangeTreeData.fire(null);
 
-    return Promise.resolve(true);
+    return true;
```
alenakhineika, line 130: Looks like the whole explorer refresh chain can be just sync.

src/extension.ts
```diff
@@ -38,7 +38,7 @@ export function activate(context: vscode.ExtensionContext): void {
   mdbExtension = new MDBExtensionController(context, {
     shouldTrackTelemetry: true
   });
-  mdbExtension.activate();
+  await mdbExtension.activate();
```
alenakhineika, line 41: We probably want to wait for activation steps to complete.

src/extension.ts
```diff
@@ -47,8 +47,8 @@ export function activate(context: vscode.ExtensionContext): void {
 }
 
 // Called when our extension is deactivated.
-export function deactivate(): void {
+export async function deactivate(): Promise<void> {
   if (mdbExtension) {
-    mdbExtension.deactivate();
+    await mdbExtension.deactivate();
```
alenakhineika, line 52: Same with deactivating.

src/mdbExtensionController.ts
```diff
@@ -122,15 +122,17 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+  async activate(): Promise<void> {
     this._explorerController.activateConnectionsTreeView();
     this._helpExplorer.activateHelpTreeView(this._telemetryService);
     this._playgroundsExplorer.activatePlaygroundsTreeView();
-    this._connectionController.loadSavedConnections();
     this._telemetryService.activateSegmentAnalytics();
-    this._languageServerController.startLanguageServer();
+
+    await this._connectionController.loadSavedConnections();
+    await this._languageServerController.startLanguageServer();
```
alenakhineika, line 132: These two events are async so we wait for them to complete.

src/mdbExtensionController.ts
```diff
@@ -520,24 +527,25 @@ export default class MDBExtensionController implements vscode.Disposable {
...
-  dispose(): void {
-    this.deactivate();
+  async dispose(): Promise<void> {
+    await this.deactivate();
   }
 
-  deactivate(): void {
-    this._connectionController.disconnect();
+  async deactivate(): Promise<void> {
+    await this._connectionController.disconnect();
```
alenakhineika, line 547: @Anemy do we actually need to wait for disconnecting before disposing of resources?

src/explorer/explorerTreeController.ts
```diff
@@ -27,7 +27,7 @@ implements vscode.TreeDataProvider<vscode.TreeItem> {
     this._connectionController.addEventListener(
       DataServiceEventTypes.CONNECTIONS_DID_CHANGE,
       () => {
-        this.refresh();
+        void this.refresh();
```
lerouxb, line 30: Isn't this now sync so no reason for void? Or maybe I'm misunderstanding and you're doing void when ignoring any return value and not just returned promises.

src/mdbExtensionController.ts
```diff
@@ -366,15 +368,16 @@ export default class MDBExtensionController implements vscode.Disposable {
       EXTENSION_COMMANDS.MDB_REFRESH_DATABASE,
       (databaseTreeItem: DatabaseTreeItem): Promise<boolean> => {
         databaseTreeItem.resetCache();
+        this._explorerController.refresh();
 
-        return this._explorerController.refresh();
+        return Promise.resolve(true);
```
lerouxb, line 373: There are a few of these registered commands that now return Promise.resolve(true) when they actually became sync. Is that just to not change the method signature for backwards compatibility? As far as I can tell some of the commands are sync and some are async so there seems to be precedent for making them sync if async is not necessary.

alenakhineika: The second argument for `this.registerCommand` should be a callback that returns a `Promise<boolean>`. For cases with async/await we can simply return true/false, but some of the functions are sync, so we force them being async by returning `Promise.resolve(true)`.

src/explorer/explorerTreeController.ts
alenakhineika: Thx for catching it! Yes, I changed the function to sync after some iterations and this is a leftover. Removed void.

src/mdbExtensionController.ts
lerouxb: Oh I see that here now: https://github.com/mongodb-js/vscode/blob/268cebe63758ae1f686d049786dec3857111a658/src/mdbExtensionController.ts#L215-L218

I first went looking for counter-examples in usage which is what confused me. I suppose for all those it happens to be implied.

src/connectionController.ts
```diff
@@ -476,7 +476,7 @@ export default class ConnectionController {
     );
 
     if (!this._activeDataService) {
-      vscode.window.showErrorMessage(
+      void vscode.window.showErrorMessage(
```
addaleax, line 479: A lot of these `void` operators are in async functions … is there a reason to avoid using `await`? What happens if this fails?

lerouxb: My 2 cents:

Personally I would default to awaiting every promise and only ignore (void) very specific ones for specific reasons. And then probably add a comment where it is ignored explaining the reasoning. Because if you await a promise it is clear where thrown errors go and bubble up and get caught. If not you end up with an unhandled rejection.

An example from my past of where code deliberately did _not_ await a promise was in code that kicked off some long-running "background" task inside a request handler and then responded with a 200 status code immediately. It still had mechanisms to catch and report errors, though - the returned promise was just handy in terms of testing so tests could wait for the thing to complete. But in that case the line of code had a comment above it explaining what was going on for future readers.

I suppose for these window object methods that return promises if the promises only resolved much later (like once the user closes a dialog or something), then that would be a good reason to ignore them. And if you have loads of calls to the same thing you wouldn't comment every time because it would be a common pattern in the project. I don't know if that's the case here, though.

alenakhineika: I had the same question. But I took a look at how vscode uses them and vscode extension samples and they do not await those 🤷🏻‍♀️

For example: https://github.com/microsoft/vscode-extension-samples/blob/main/terminal-sample/src/extension.ts#L231

the same with showInformationMessage: https://github.com/microsoft/vscode-extension-samples/blob/main/tree-view-sample/src/nodeDependencies.ts#L34

alenakhineika: The promise is resolved when modal is closed, so as we discussed in slack `then void is probably the right thing`.

src/explorer/explorerTreeController.ts
```diff
@@ -38,7 +38,7 @@ implements vscode.TreeDataProvider<vscode.TreeItem> {
     this._connectionController.removeEventListener(
       DataServiceEventTypes.CONNECTIONS_DID_CHANGE,
       () => {
-        this.refresh();
+        void this.refresh();
```
mcasimir: I guess this is as above right void not necessary? But in this case this was also ok here: 

```
async () => { await this.refresh() }
```

As far as i remember EventEmitter."method..." accepts any function and won't wait for the promise

alenakhineika: This code is outdated. The `this.refresh()` is sync now.

alenakhineika: This code is outdated. The `this.refresh()` is sync now.

src/explorer/schemaTreeItem.ts
```diff
@@ -212,7 +212,7 @@ export default class SchemaTreeItem extends vscode.TreeItem
     this.cacheIsUpToDate = true;
 
     if (!schema || !schema.fields || schema.fields.length < 1) {
-      vscode.window.showInformationMessage(
+      void vscode.window.showInformationMessage(
```
mcasimir, line 215: what would we wait with `await vscode.window.showInformationMessage`? Is it waiting for the info to be created or to be closed?

src/explorer/explorerTreeController.ts
alenakhineika: And in general wouldn't it be more descriptive to mark such function called as `void` to kind of let people who reads code know that emitter event does not wait for a response.

src/language/languageServerController.ts
```diff
@@ -138,7 +138,7 @@ export default class LanguageServerController {
     }
 
     // Stop the language server
-    this._client.stop();
+    void this._client.stop();
```
mcasimir, line 141: what happens if this throws an error?

src/explorer/schemaTreeItem.ts
alenakhineika: This is the same as we talked here above about `showErrorMessage`. It waits when the modal is closed.

src/language/mongoDBService.ts
```diff
@@ -211,7 +211,7 @@ export default class MongoDBService {
             );
           }
 
-          worker.terminate().then(() => {
+          void worker.terminate().then(() => {
```
mcasimir, line 214: what happens if this fails?

src/language/languageServerController.ts
alenakhineika: This is the same as `this._client.start()`, the error is being caught by the extension and shows the modal with the error that the language server can not start/stop.

src/language/mongoDBService.ts
```diff
@@ -283,7 +283,7 @@ export default class MongoDBService {
           );
         }
 
-        worker.terminate().then(() => {
+        void worker.terminate().then(() => {
```
mcasimir, line 286: Can't you use `async/await` also in the callback?

```
 worker.on('message', async (response: [Error, CompletionItem[] | []]) => { ... })
```

And then just `await` here .. i'm also not sure if this `terminate` is so fail safe, won't we have an unhandled promise rejection if that fails?

src/language/mongoDBService.ts
alenakhineika: It is inside try/catch, do you think some extra check should be done here?

src/language/mongoDBService.ts
```diff
@@ -381,7 +381,7 @@ export default class MongoDBService {
             this._connection.console.log(`SCHEMA error: ${util.inspect(error)}`);
           }
 
-          worker.terminate().then(() => {
+          void worker.terminate().then(() => {
```
mcasimir, line 384: Same here .. i believe you can just move to async await the handler

src/language/worker.ts
```diff
@@ -219,6 +219,6 @@ const handleMessageFromParentPort = async(message: string) => {
 parentPort?.once(
   'message',
   (message: string): void => {
-    handleMessageFromParentPort(message);
+    void handleMessageFromParentPort(message);
```
mcasimir, line 222: what happens if this promise rejects? .. also you can make the callback async

src/mdbExtensionController.ts
```diff
@@ -452,25 +456,27 @@ export default class MDBExtensionController implements vscode.Disposable {
...
     this.registerCommand(
       EXTENSION_COMMANDS.MDB_REFRESH_DOCUMENT_LIST,
-      (documentsListTreeItem: DocumentListTreeItem): Promise<boolean> => {
-        documentsListTreeItem.resetCache();
+      async (documentsListTreeItem: DocumentListTreeItem): Promise<boolean> => {
+        await documentsListTreeItem.resetCache();
+        this._explorerController.refresh();
 
-        return this._explorerController.refresh();
+        return Promise.resolve(true);
```
mcasimir, line 463: Why not just `return true`?

src/mdbExtensionController.ts
mcasimir: What is the issue with waiting?

src/storage/storageController.ts
```diff
@@ -58,11 +58,11 @@ export default class StorageController {
     storageScope?: StorageScope
   ): Thenable<void> {
     if (storageScope === StorageScope.WORKSPACE) {
-      this._workspaceState.update(variableName, value);
+      void this._workspaceState.update(variableName, value);
```
mcasimir, line 61: can't this thing fail? what is the issue with waiting?

src/language/mongoDBService.ts
alenakhineika: I think WorkerThreads expect a void function here and do not allow Promise.

src/mdbExtensionController.ts
alenakhineika: Missed this one! Will remove Promise.

src/language/mongoDBService.ts
mcasimir: that try/catch would not necessarily catch a rejection in terminate

src/language/mongoDBService.ts
mcasimir: That can be worked around, i don't think we should let that force us to use a particular style of promises.

src/storage/storageController.ts
alenakhineika: @Anemy probably you can answer this one?

src/language/worker.ts
alenakhineika: So as we discussed in slack, let's move this additional refactoring out of the scope of this PR. The current PR introduces a tool that helps you identify where is an intentional void function call and where is accidentally missed await. We can check all void cases later together with @Anemy since we share the code of the extension and update these cases. Please create a ticket for this! I also will create one to finish replacing promises with async/await syntax.

src/language/mongoDBService.ts
lerouxb: I see this is in a big promise, but a few lines below there is precedent for `await worker.terminate();` and then resolving the promise.

try/catch won't catch this because you only call .then() and not .catch(). it will only catch if you're using await. Without await you're on your own and have to catch, basically. So if terminate rejects you'll have an unhandled rejection.

src/language/mongoDBService.ts
lerouxb: Personally I would also lean towards what @mcasimir said because if it throws there will be an unhandled rejection. Alternatively you can .catch() and do something with the error.

Thinking about it some more that try/catch won't really catch much since all the action happens in a callback. Personally I'd remove it and maybe audit usage of try/catch when there are promises and callbacks inside them.

lerouxb: I just audited all the try/catch calls outside of tests, looking for .then() or callbacks that might be asynchronous and didn't spot any that stood out to me other than the four in this file.

src/language/languageServerController.ts
lerouxb: Ideally I think we'd have a comment here explaining that reasoning.

src/language/server.ts
```diff
@@ -68,7 +68,7 @@ connection.onInitialize((params: InitializeParams) => {
 connection.onInitialized(() => {
   if (hasConfigurationCapability) {
     // Register for all configuration changes.
-    connection.client.register(
+    void connection.client.register(
```
lerouxb, line 71: This one probably also deserves a comment explaining that connection.client.register deals with the errors. Assuming that's the case.

alenakhineika:
Heads up, this PR is focusing on adding void/await to existing code. Based on all conversations here that are outside of the scope of this PR I will create separate tickets later to address them and link to this PR for refernce:

- finish replacing promises with async/await
- do refactoring of some try/catch blocks that have promises inside and can potentially cause unhandled promise rejections errors
- review void cases and see maybe we can convert them to await or add meaningful comments why we prefer to have void here

src/storage/storageController.ts
Anemy: I remember it being flaky on ci, sometimes it would not resolve or error and would just stall. The value would still update though.

Issue #284, opened 2021-04-08
Bug Report


jhmoon2000:
#### Current Behavior
Cannot open documents with a slash in _id.

It generates error message like 'VIEW_DOCUMENT_SCHEME:/test.test:"abcd'

#### Environment

- node.js / npm versions:
- OS:

alenakhineika:
Hey, thanks for reporting it! I've opened a JIRA ticket so you can track progress here: https://jira.mongodb.org/projects/VSCODE/issues/VSCODE-276

PR #342, opened 2021-08-31
fix(data-explorer): opening _ids with slashes in them VSCODE-276

Anemy:
VSCODE-276

We were opening documents with the virtual file name with a slash in it which made VSCode think it was in a folder and couldn't find it. This PR updates it so that now we trim these special characters from the file name before displaying them.

Also added a playground useful for seeding a db with some docs with different kinds of _ids.

Addresses https://github.com/mongodb-js/vscode/issues/284

https://user-images.githubusercontent.com/1791149/131581509-2661138a-bc66-4db8-a7f4-5cbfcf133258.mp4


src/editors/editorsController.ts
```diff
@@ -32,6 +32,21 @@ import TelemetryService from '../telemetry/telemetryService';
 
 const log = createLogger('editors controller');
 
+export function getFileDisplayNameForDocumentId(documentId) {
+  let displayName = EJSON.stringify(documentId);
+
+  // Strip out all special characters to ensure VSCode handles
+  // it correctly in a uri.
+  const regularExpression = /[^a-z0-9{}_-\s":,$!*#@()]/gi;
+  displayName = displayName.replace(regularExpression, '');
```
Anemy, line 41: I think we could do a better implementation than this, going to update before merging.

PR #347, opened 2021-09-28
fix(language-server): remove loading ssl/tls files to just passing file names VSCODE-292

Anemy:
VSCODE-292

PR removes loading the files of tls/ssl files. Now we just pass the file names. I think we were doing this because of previous driver behavior, after we updated the driver in the latest release this probably broke.

Tested on mac w/ server+client and x509
Before releasing we should make sure we have this in the testing matrix and test it (I must have missed in it in the last release). 
We could probably add some nice e2e connectivity tests w/ playgrounds following the current Compass work.

Connecting + running a playground:
Screen Shot 2021-09-28 at 4 40 13 PM


alenakhineika (approved):
👍

PR #369, opened 2021-11-18
feat: trim selected text for export to language (VSCODE-296)

alenakhineika:
## Description

Trim selected text to make the extension smarter on when to show export to language code actions.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/test/suite/language/mongoDBService.test.ts
```diff
@@ -1246,4 +1247,160 @@ suite('MongoDBService Test Suite', () => {
       expect(result).to.deep.equal(expectedResult);
     });
   });
+
+  suite('getExportToLanguageMode', function () {
```
Anemy, line 1251: Should we also add a test for error case?
What happens if this fails?

Anemy (approved):
Changes look good, nice fix on only showing the command for playground files.
One suggestion on testing error case.
Not related to this pr, but maybe we want to address it here as we're also updating a few other things with export to language:
I ran into this error when clicking exclude driver syntax with Java:
Screen Shot 2021-11-22 at 11 04 54 AM
With C# we don't allow users to include driver syntax, maybe we can just hide the codelens if it's that language?

src/test/suite/language/mongoDBService.test.ts
alenakhineika: Pushed updates for both of your comments! 🔥

Anemy (approved):
👍

src/test/suite/editors/playgroundController.test.ts
```diff
@@ -193,41 +192,50 @@ suite('Playground Controller Test Suite', function () {
     testPlaygroundController._activeTextEditor = undefined;
 
     test('run all playground blocks should throw the playground not found error', async () => {
-      const errorMessage =
-        "Please open a '.mongodb' playground file before running it.";
-
-      fakeShowErrorMessage.resolves(errorMessage);
+      const expectedMessage = "Please open a '.mongodb' playground file before running it.";
+      const fakeShowErrorMessage: any = sinon.fake();
```
Anemy, line 196: We can probably use the `sinon.SinonFake` type here

alenakhineika: We took a look at this before and kept any because `Property 'firstArg' does not exist on type 'SinonFake'.`.

PR #371, opened 2021-11-25
refactor: apply tell-don’t-ask principle

alenakhineika:
## Description

Reduce complexity and improve code readability by applying Tell-Don’t-Ask refactoring principle: instead of asking an object about its state and then performing actions based on this, it’s much easier to simply tell the object what it needs to do and let it decide for itself how to do that.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


src/language/visitor.ts
```diff
@@ -371,113 +294,99 @@ export class Visitor {
...
-        node.callee.type === 'MemberExpression' &&
-        this._checkIsCollectionName(node.callee))
+      node.type === 'MemberExpression' &&
+      node.object.type === 'Identifier' &&
+      node.object.name === 'db' &&
+      node.property.type === 'Identifier' &&
+      node.property.name.includes(PLACEHOLDER)
     ) {
-      return true;
+      this._state.isCollectionName = true;
```
Anemy, line 374: Are these variables ever unset or set to `false`?

alenakhineika: Yeah, they all are being reset with each parse of the file:
https://github.com/mongodb-js/vscode/blob/apply-tell-do-not-ask-principle/src/language/visitor.ts#L141

Anemy (approved):

PR #372, opened 2021-12-01
chore: Update package lock

Anemy:
No behavior changes, just running npm install and these are the generated changes in the package-lock.json

Also added a timeout increase for a flaky test.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -125,7 +125,7 @@
       "engines": {
         "node": "^14.17.3",
         "npm": "^7.19.0",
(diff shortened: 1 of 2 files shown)


alenakhineika:
Yeah, you have the same issue that I mentioned earlier with `node-gyp`. I added a workaround here:
https://github.com/mongodb-js/vscode/pull/371/files#diff-07e18c3cd0cb1d22e44acffda4355e2d0c61d769b2f257679630a542dafac291R52

PR #377, opened 2021-12-16
VSCODE-311: new data service and connection model

alenakhineika:
## Description

Switch to a new connection info format. The new data service is using a new connection info object that should be adopted by the extension and used to connect and save connections.

VSCode should start using a new connection info object for the saved connections including migration of previously saved connections to a new format and using the `connection-secrets` module from Compass to protect all secrets.

- [VSCODE-312](https://jira.mongodb.org/browse/VSCODE-312) - Switch to a new data service and connection model
- [VSCODE-313](https://jira.mongodb.org/browse/VSCODE-313) - Use the connection-secrets module to protect all secrets
- [VSCODE-314](https://jira.mongodb.org/browse/VSCODE-314) - Transform previously saved connections to a new format

I also removed LSP inspector since the [webview extension](https://marketplace.visualstudio.com/items?itemName=octref.lsp-inspector-webview) is buggy and doesn't work almost all the time.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Dependents

Dependent on https://github.com/mongodb-js/compass/pull/2650

## Types of changes

- [ ] *Backport Needed*


src/language/mongoDBService.ts
```diff
@@ -8,13 +8,12 @@ import {
   MarkupContent,
   MarkupKind
 } from 'vscode-languageserver/node';
+import { MongoClientOptions } from 'mongodb';
```
addaleax, line 11:
```suggestion
import type { MongoClientOptions } from 'mongodb';
```

addaleax (commented):
Nice work! Took a first look at everything, but I think I’ll need to spend a bit more time before I can say that I fully follow the data flow everywhere :)

src/commands/launchMongoShell.ts
```diff
@@ -120,16 +119,11 @@ const openMongoDBShell = (connectionController: ConnectionController): Promise<b
...
+  const mdbConnectionString = activeDerivedConnectionModel.driverUrlWithSsh || '';
 
-  const mdbConnectionString = activeConnectionModel
-    ? activeConnectionModel.driverUrlWithSsh
-    : '';
-
-  if (activeConnectionModel && isSslConnection(activeConnectionModel)) {
-    mdbSslOptions = getSslOptions(activeConnectionModel.driverOptions);
+  if (isSslConnection(activeDerivedConnectionModel)) {
+    mdbSslOptions = getSslOptions(activeDerivedConnectionModel.driverOptions);
```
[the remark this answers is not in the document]
mcasimir: Just wanted to add that this approach won't work anymore already, we already chatted with @alenakhineika a bit, but generally using the ConnectionModel, converted from the connection string / connection info is not a good idea, and we probably don't want to do a release with that.

The conversion from ConnectionInfo to ConnectionModel only guarantees that `infoToModel(info).save()` and a subsequent `modelToInfo(model)` would work. 

Anything else, and `driverOptions` is one of those .. won't probably work correctly

src/connectionController.ts
```diff
@@ -1,23 +1,33 @@
...
-import SSH_TUNNEL_TYPES from './views/webview-app/connection-model/constants/ssh-tunnel-types';
-import { StatusView } from './views';
+import RawConnectionModel from './views/webview-app/connection-model/connection-model';
+import { StorageLocation, ConnectionsFromStorage } from './storage/storageController';
 import { StorageController, StorageVariables } from './storage';
+import { StatusView } from './views';
 import TelemetryService from './telemetry/telemetryService';
 
-const { name, version } = require('../package.json');
+// TODO: export these helpers from mongodb-data-service.
```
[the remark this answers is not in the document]
alenakhineika: Definitely in the scope of the unified connections epic, but probably as separate pr.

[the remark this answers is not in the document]
addaleax: Is there a ticket for this?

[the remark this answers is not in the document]
addaleax: Also, I opened https://github.com/mongodb-js/compass/pull/2647 for the relevant changes in Compass, maybe we can also already address them here (after that Compass PR is merged)? Then this becomes less urgent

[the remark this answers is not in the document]
alenakhineika: Here is the ticket: https://jira.mongodb.org/browse/COMPASS-5378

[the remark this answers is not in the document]
alenakhineika: And yes we can bump a new data service version here when https://github.com/mongodb-js/compass/pull/2647 is merged.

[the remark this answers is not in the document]
addaleax: No, I meant, since we’re deciding to duplicate the code here instead of exposing it in data-service right now, could we do those changes also to the duplicated code?

src/storage/storageController.ts
```diff
@@ -24,52 +26,50 @@ export enum DefaultSavingLocations {
...
 }
 
-// A saved connection does not contain the connection information,
-// just metadata about the connection and an id for referencing secure storage.
-export type SavedConnection = {
-  id: string; // uuidv4
-  name: string; // Possibly user given name, not unique.
-  storageLocation: StorageScope;
+export type ConnectionsFromStorage = {
+  [key: string]: StoreConnectionInfo
```
addaleax, line 30: (nit:) Maybe this could be

```suggestion
  [connectionId: string]: StoreConnectionInfo
```

for clarity, if I’m following everything correctly

src/connectionController.ts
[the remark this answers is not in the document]
alenakhineika: Ahh got it! let's see maybe I will do COMPASS-5378 tomorrow, so we just remove the duplicate completely, otherwise yeah I can add your change here as well.

src/storage/storageController.ts
```diff
@@ -24,52 +26,50 @@ export enum DefaultSavingLocations {
...
+  StorageVariables.GLOBAL_SAVED_CONNECTIONS |
+  StorageVariables.WORKSPACE_SAVED_CONNECTIONS |
+  StorageVariables.GLOBAL_HAS_BEEN_SHOWN_INITIAL_VIEW;
+
+type StoredItem<T> =
+    T extends StorageVariables.GLOBAL_USER_ID ? string :
+    T extends StorageVariables.GLOBAL_HAS_BEEN_SHOWN_INITIAL_VIEW ? boolean :
+    T extends StorageVariables.GLOBAL_SAVED_CONNECTIONS ? ConnectionsFromStorage :
+    T extends StorageVariables.WORKSPACE_SAVED_CONNECTIONS ? ConnectionsFromStorage :
+    never;
```
addaleax, line 43: (nit): How about this? It feels a bit more TypeScript-y, and it avoids a bit of duplication

```suggestion
interface StorageVariableContents {
  [StorageVariables.GLOBAL_USER_ID]: string;
  [StorageVariables.GLOBAL_HAS_BEEN_SHOWN_INITIAL_VIEW]: boolean;
  [StorageVariables.GLOBAL_SAVED_CONNECTIONS]: ConnectionsFromStorage;
  [StorageVariables.WORKSPACE_SAVED_CONNECTIONS]: ConnectionsFromStorage;
}

type StoredVariableName = keyof StorageVariableContents;
type StoredItem<T extends StoredVariableName> = StorageVariableContents[T];
```

(and as you might be able to imagine, I think it would be even simpler if instead of `StorageVariables.GLOBAL_USER_ID` we used `'GLOBAL_USER_ID'` as a constant string, but I don’t want to add too much stuff to this PR that isn’t already being touched anyway)

src/storage/storageController.ts
```diff
@@ -24,52 +26,50 @@ export enum DefaultSavingLocations {
...
   }
 
   // Update something in the storage. Defaults to global storage (not workspace).
   update(
     variableName: StorageVariables,
-    value: any,
-    storageScope?: StorageScope
+    value: boolean | string | ConnectionsFromStorage,
+    storageLocation: StorageLocation = StorageLocation.GLOBAL
   ): Thenable<void> {
```
addaleax, line 64: (nit:) This could also be

```js

  update<T extends StoredVariableName>(
    variableName: T,
    value: StoredItem<T>,
    storageLocation: StorageLocation = StorageLocation.GLOBAL
  ): Thenable<void>
```

to make sure that `variableName` and `value` match

src/storage/storageController.ts
```diff
@@ -103,103 +100,96 @@ export default class StorageController {
...
+      storeConnectionInfo.storageLocation = this.getPreferedStorageLocationFromConfiguration();
+    } else {
+      storeConnectionInfo.storageLocation = await this.getStorageLocationFromPrompt();
+    }
+
+    if (storeConnectionInfo.storageLocation === StorageLocation.WORKSPACE) {
+      await this.saveConnectionToWorkspaceStore(storeConnectionInfo);
+    } else if (storeConnectionInfo.storageLocation === StorageLocation.GLOBAL) {
+      await this.saveConnectionToGlobalStore(storeConnectionInfo);
+    }
```
addaleax, line 142: (non-blocking:) These two methods look very similar – would it make sense to share their code, and pass in the storage location/variable name as a parameter instead?

src/telemetry/connectionTelemetry.ts
```diff
@@ -78,65 +76,31 @@ async function getCloudInfoFromDataService(
...
-    derived: true
-  });
-
-  /* eslint-disable camelcase */
+  const connectionString = dataService.getConnectionString();
+  const firstHost = connectionString.hosts[0] || '';
+  const [hostname] = firstHost.split(':');
+  const authMechanism = connectionString.searchParams.get('authMechanism');
+  const username = connectionString.username ? 'DEFAULT' : 'NONE';
+  const authStrategy = authMechanism ? authMechanism : username;
```
addaleax, line 87: (nit:)

```suggestion
  const authStrategy = authMechanism ?? username;
```

src/connectionController.ts
```diff
@@ -31,30 +41,41 @@ export enum ConnectionTypes {
...
+}
 
 export default class ConnectionController {
   // This is a map of connection ids to their configurations.
   // These connections can be saved on the session (runtime),
   // on the workspace, or globally in vscode.
-  _connections: { [key: string]: LoadedConnection } = {};
-  _activeDataService: null | DataServiceType = null;
-  _activeConnectionModel: null | ConnectionModel = null;
+  _connections: { [key: string]: StoreConnectionInfo } = {};
```
addaleax, line 76: (nit:) also here?

```suggestion
  _connections: { [connectionId: string]: StoreConnectionInfo } = {};
```

src/connectionController.ts
```diff
@@ -77,106 +98,132 @@ export default class ConnectionController {
...
+      return {
+        ...savedConnectionInfo,
+        connectionOptions: connectionInfoWithSecrets.connectionOptions
       };
     } catch (error) {
-      // Here we're leniant when loading connections in case their
+      // Here we're lenient when loading connections in case their
       // connections have become corrupted.
+      const printableError = error as { message: string };
+      log.error(`Mergin connection with secrets failed: ${printableError.message}`);
```
addaleax, line 180:
```suggestion
      log.error(`Merging connection with secrets failed: ${printableError.message}`);
```

src/connectionController.ts
```diff
@@ -77,106 +98,132 @@ export default class ConnectionController {
...
+      StorageLocation.GLOBAL
+    );
+    await this._loadSavedConnectionsByStore(existingGlobalConnections);
+
+    // Try to pull in the connections previously saved in the workspace storage of vscode.
+    const existingWorkspaceConnections = this._storageController.get(
+      StorageVariables.WORKSPACE_SAVED_CONNECTIONS,
+      StorageLocation.WORKSPACE
+    );
+    await this._loadSavedConnectionsByStore(existingWorkspaceConnections);
```
addaleax, line 226: (nit:) Since we’re already parallelizing this above, could/should these also be done in parallel?

```suggestion
    await Promise.all([(async() => {
      // Try to pull in the connections previously saved in the global storage of vscode.
      const existingGlobalConnections = this._storageController.get(
        StorageVariables.GLOBAL_SAVED_CONNECTIONS,
        StorageLocation.GLOBAL
      );
      await this._loadSavedConnectionsByStore(existingGlobalConnections);
    })(), (async() => {
      // Try to pull in the connections previously saved in the workspace storage of vscode.
      const existingWorkspaceConnections = this._storageController.get(
        StorageVariables.WORKSPACE_SAVED_CONNECTIONS,
        StorageLocation.WORKSPACE
      );
      await this._loadSavedConnectionsByStore(existingWorkspaceConnections);
    })()]);
```

src/connectionController.ts
```diff
@@ -240,140 +297,92 @@ export default class ConnectionController {
...
   }
 
-  parseNewConnection(newConnectionModel: ConnectionModel): ConnectionModel {
-    // Here we re-parse the connection, as it can be loaded from storage or
-    // passed by the connection model without the class methods.
-    const connectionModel: ConnectionModel = new Connection(
-      newConnectionModel
-    );
+  parseNewConnection(rawConnectionModel: RawConnectionModel): ConnectionInfo {
+    const connectionModel = new ConnectionModel(rawConnectionModel);
```
addaleax, line 305: I guess this is the part where we’re not sure about whether it’s alright to keep going with `ConnectionModel`? I would think it’s okay, yeah. We should probably just leave a ticket about _eventually_ removing it. (But it might be nice if @Anemy could confirm this :slightly_smiling_face:)

src/views/webviewController.ts
```diff
@@ -83,9 +83,7 @@ export default class WebviewController {
     connectionAttemptId: string
   ): Promise<void> => {
     try {
-      const connectionModel = this._connectionController
-        .parseNewConnection(rawConnectionModel as any);
-
+      const connectionModel = this._connectionController.parseNewConnection(rawConnectionModel);
```
addaleax, line 86: Should this be called `connectionInfo`? Otherwise it’s confusing, because this is a `ConnectionInfo`, not a `ConnectionModel`

src/connectionController.ts
```diff
@@ -416,58 +422,55 @@ export default class ConnectionController {
...
 
-  async connectWithConnectionId(connectionId: string): Promise<boolean> {
-    if (!this._connections[connectionId]) {
-      throw new Error('Connection not found.');
-    }
+  private _endPrevConnectAttempt (attempt: {
+    connectionId: string,
+    connectingAttemptVersion: null | string,
+    newDataService: DataService
+  }): Boolean {
```
addaleax, line 445:
```suggestion
  }): boolean {
```

(surprised that TS didn’t error on this one, maybe it’s because we’re not using strict mode in vscode?)

addaleax (approved):
Looks good! I left a few more comments, but they are mostly non-blocking, and in any case, nothing larger.

One thing I’m curious about: Is there a specfic reason for keeping the connections in memory in `ConnectionController._connections`, instead of always accessing them through the storage API? I’m not saying we should change this, I’m just wondering :slightly_smiling_face:

src/connectionController.ts
Anemy: I think it's fine yup. It's probably not worth adding all the logic to build the connectionInfo from the connection configuration object the form in the webview creates. Would be nice to remove this usage of connection-model eventually, and this would definitely be something that sharing the connect form would help make go away as well.

src/connectionController.ts
```diff
@@ -714,32 +698,21 @@ export default class ConnectionController {
...
+    return this._activeDataService?.getMongoClientConnectionOptions();
   }
 
-  isCurrentlyConnected(): boolean {
-    return this._activeDataService !== null;
-  }
 
-  getActiveDataService(): null | DataServiceType {
-    return this._activeDataService;
+  getActiveDerivedConnectionModel(): ConnectionModel {
```
Anemy, line 710: Is this function used? If it's unused maybe we can remove it? I might just be missing where it's called.

src/connectionController.ts
```diff
@@ -416,58 +422,55 @@ export default class ConnectionController {
     void vscode.window.showInformationMessage('MongoDB connection successful.');
 
     this._activeDataService = newDataService;
-    this._activeConnectionModel = connectionModel;
+    this._activeConnectionModel = await convertConnectionInfoToModel({ connectionOptions });
```
Anemy, line 425: Do we still need this? Unless I'm missing something looks like we can pull everything from the connection options for the connection, maybe we can remove it.

src/connectionController.ts
```diff
@@ -192,8 +239,11 @@ export default class ConnectionController {
           'e.g. mongodb+srv://username:password@cluster0.mongodb.net/admin',
         prompt: 'Enter your connection string (SRV or standard)',
         validateInput: (uri: string) => {
-          if (!Connection.isURI(uri)) {
-            return 'MongoDB connection strings begin with "mongodb://" or "mongodb+srv://"';
+          try {
+            // eslint-disable-next-line no-new
+            new ConnectionString(uri);
```
Anemy, line 244: I like this change, but I'm wondering if the old one, although not as good of a validation check, might give a user unsure what to enter in something more information to help their case.
Mostly if they input something like a url but is invalid they'd get an `Invalid connection string` error message which doesn't hint at what might go there. https://github.com/mongodb-js/mongodb-connection-string-url/blob/main/src/index.ts#L106
Maybe something for that package instead of here. We do have a placeholder here that should give those users something to go off of so this may be a non-issue.

alenakhineika:
@addaleax Answering your question why we keep connections in memory: first of all, I didn't change this logic in this pr, this worked like this before. But was thinking about it while refactoring and kept like this because of two reasons:
- to pull all connections at once to migrate them to a new format:
- to fetch secret for all connections to allow users seamlessly to switch between connections without accessing a keychain, and allow copying URLs for such connections also without accessing the keychain.
We also discusses it with Maurizio and as I remember he was also up to pulling them all together. I am open here for a discussion if you have a strong opinion to do it differently!

src/storage/storageController.ts
alenakhineika: Love the suggestion with the interface format! ❤️ I felt like my solution could be optimized but didn't figure out your solution by myself, thx!

src/storage/storageController.ts
alenakhineika: Yeah, I improved it a bit already comparing to main, but you are right and can be done better...)

src/connectionController.ts
alenakhineika: Leftover :)

src/connectionController.ts
alenakhineika: @addaleax

src/connectionController.ts
alenakhineika: Also leftover :) thx for noticing it!

src/connectionController.ts
addaleax: I can definitely see the concern here – if you want to keep the previous check additionally (i.e. add 

```js
          if (!uri.startsWith('mongodb://') && !uri.startsWith('mongodb+srv://')) {
            return 'MongoDB connection strings begin with "mongodb://" or "mongodb+srv://"';
```

before the `new ConnectionString` check), I think that’s okay and won’t make anything worse, but I also honestly think that this isn’t something we need to worry about a ton, and I do agree with @Anemy that it would be a bit nicer to improve the error message inside the connection string url package as a long-term solution :slightly_smiling_face:

Anemy (approved):
nice work - big changes that touch almost all the extension. Solid cleanup on the language server.

src/explorer/connectionTreeItem.ts
```diff
@@ -106,44 +80,18 @@ export default class ConnectionTreeItem extends vscode.TreeItem
     return element;
   }
 
-  async listDatabasesUserHasAccessTo(
```
Anemy, line 109: Cool that we're sharing this with compass now 👍

alenakhineika: Yes! I was very excited to see this change in compass and know the fact that we can reuse it in vscode!

PR #381, opened 2022-01-11
test: add tests cases for connection info and secrets VSCODE-317

alenakhineika:
## Description

Write more tests to ensure a proper connection format is passed across the app and check data stored to the disc.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -68,11 +68,13 @@ interface ConnectionSecretsInfo {
   secrets: ConnectionSecrets
 }
 
+type StoreConnectionInfoWithConnectionOptions = StoreConnectionInfo & Required<Pick<StoreConnectionInfo, 'connectionOptions'>>;
+
 export default class ConnectionController {
   // This is a map of connection ids to their configurations.
   // These connections can be saved on the session (runtime),
   // on the workspace, or globally in vscode.
-  _connections: { [connectionId: string]: StoreConnectionInfo } = {};
+  _connections: { [connectionId: string]: StoreConnectionInfoWithConnectionOptions } = {};
   _activeDataService: DataService| null = null;
   _storageController: StorageController;
 
@@ -107,7 +109,7 @@ export default class ConnectionController {
 
   async _migratePreviouslySavedConnection(
     savedConnectionInfo: StoreConnectionInfo
-  ): Promise<StoreConnectionInfo> {
+  ): Promise<StoreConnectionInfoWithConnectionOptions> {
     // Transform a raw connection model from storage to an ampersand model.
     const newConnectionInfoWithSecrets = convertConnectionModelToInfo(savedConnectionInfo.connectionModel);
 
@@ -126,7 +128,7 @@ export default class ConnectionController {
 
   async _getConnectionInfoWithSecrets(
     savedConnectionInfo: StoreConnectionInfo
-  ): Promise<StoreConnectionInfo|undefined> {
+  ): Promise<StoreConnectionInfoWithConnectionOptions|undefined> {
     // Migrate previously saved connections to a new format.
     // Save only secrets to keychain.
     // Remove connectionModel and use connectionOptions instead.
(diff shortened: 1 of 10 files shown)


src/connectionController.ts
```diff
@@ -44,7 +44,7 @@ export interface StoreConnectionInfo {
   id: string; // Connection model id or a new uuid.
   name: string; // Possibly user given name, not unique.
   storageLocation: StorageLocation;
-  connectionOptions: ConnectionOptions;
+  connectionOptions?: ConnectionOptions;
```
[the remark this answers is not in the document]
alenakhineika: In old connection info object save to disc. We migrate those connections to a new format by removing connectionModel and adding connectionOptions.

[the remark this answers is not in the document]
addaleax: But in `_connections` the field is always there, correct? It might be nice to be explicit about that in the types and omit some of the type casts here, I left a suggestion about that below, feel free to check if it works

src/connectionController.ts
```diff
@@ -74,6 +74,7 @@ export default class ConnectionController {
   // on the workspace, or globally in vscode.
   _connections: { [connectionId: string]: StoreConnectionInfo } = {};
```
addaleax, line 75:
```suggestion
  _connections: { 
    [connectionId: string]: StoreConnectionInfo & Required<Pick<StoreConnectionInfo, 'connectionOptions'>> 
  } = {};
```

src/test/suite/connectionController.test.ts
```diff
@@ -915,4 +924,328 @@ suite('Connection Controller Test Suite', function () {
...
+      TEST_DATABASE_URI_USER
+    );
+
+    const workspaceStoreConnections = testConnectionController._storageController.get(
+      StorageVariables.GLOBAL_SAVED_CONNECTIONS
+    );
+
+    assert(
+      workspaceStoreConnections,
+      `Expected workspace store has onnections found ${workspaceStoreConnections}`
```
addaleax, line 1169:
```suggestion
      `Expected workspace store has connections found ${workspaceStoreConnections}`
```

or

```suggestion
      `Expected workspace store to have connections, found ${workspaceStoreConnections}`
```

(or just drop the message altogether, it doesn’t really provide much value here anyway)

src/test/suite/connectionController.test.ts
```diff
@@ -915,4 +924,328 @@ suite('Connection Controller Test Suite', function () {
...
+      `Expected a connection on disc has appname found ${connections[0].connectionOptions?.connectionString}`
+    );
+    assert(
+      testConnectionController._connections[connections[0].id].connectionOptions?.connectionString.includes(TEST_USER_PASSWORD) === true,
+      `Expected a connection in memory includes password found ${connections[0].connectionOptions?.connectionString}`
+    );
+    assert(
+      testConnectionController._connections[connections[0].id].name === 'localhost:27018',
+      `Expected a connection has the 'localhost:27018' name found ${testConnectionController._connections[connections[0].id].name}`
+    );
```
addaleax, line 1197: nit: It would be a bit less verbose if we only add messages for assertions when they provide extra context, e.g. here they don’t and you could just do:

```suggestion
    assert.strictEqual(
      testConnectionController._connections[connections[0].id].name,
      'localhost:27018'
    );
```

and get a similar (or better) error message

addaleax (approved):

src/test/suite/connectionController.test.ts
```diff
@@ -1,10 +1,11 @@
 import * as sinon from 'sinon';
+import * as util from 'util';
```
Anemy, line 2: Is there a reason we want to use this format for importing? 
`import { promisify } from 'util';` looks a little cleaner imo - I'm cool with either.

Anemy (approved):
lgtm! One comment on import styles, and a +1 on Anna's comment on using `strictEqual` where we can so we can get nice test failure messages without having to write them ourselves. Nice work

src/test/suite/connectionController.test.ts
alenakhineika: I transformed it to debug like this `console.log(`${util.inspect(error)}`)` and kept it this way so in the future I could do it again.

addaleax (approved):

addaleax (approved):

PR #390, opened 2022-02-07
feat(connect): use extracted connection logic from mongosh MONGOSH-1080

alenakhineika:
## Description

- Bump the `data-service` version to start using extracted connection logic from mongosh.
- Refactor connection controller to keep original stack trace while throwing errors.
- Improve formatting error messages.
- Remove `connection-model` from dependencies.
- Set setting `strictModuleExceptionHandling: true` because by default it false: https://webpack.js.org/configuration/output/#outputstrictmoduleerrorhandling. We need `devtools-connect` throw the error if can not resolve `os-dns-native`: https://github.com/mongodb-js/devtools-connect/blob/main/src/connect.ts#L91.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/test/suite/connectionController.test.ts
```diff
@@ -743,17 +739,20 @@ suite('Connection Controller Test Suite', function () {
...
         isSrvRecord: false,
         hostname: 'localhost',
         port: 27017,
         hosts: [{ host: 'localhost', port: 27017 }],
         extraOptions: {},
         connectionType: 'NODE_DRIVER',
-        authStrategy: 'NONE',
-        readPreference: 'primary',
+        authStrategy: AUTH_STRATEGY_VALUES.NONE,
+        readPreference: READ_PREFERENCES.PRIMARY,
```
addaleax, line 752: Why did these changes become necessary? I’d really like to go the other direction if we can with union types, unless there’s some real value in the enums here which I’m missing

src/utils/formatError.ts
```diff
@@ -0,0 +1,16 @@
+export default (err: any, message?: string): {
+  name?: string;
+  message: string;
+  stack?: string;
+} => {
+  if (err instanceof Error) {
```
addaleax, line 6: If you want to check whether something is a built-in `Error` object, this is better:

```suggestion
  if (Object.prototype.toString.call(err) === '[object Error]') {
```

since `instanceof` is not a reliable brand check in JS:

```js
> vm.runInNewContext('new Error("banana")')
Error: banana
    at evalmachine.<anonymous>:1:1
    [… rest of stack …]
> vm.runInNewContext('new Error("banana")') instanceof Error
false
```

or alternatively I’d be more specific and just test whether the actual properties you’re looking for are present:

```suggestion
  if (typeof err?.name === 'string' && typeof err?.message === 'string' && typeof err?.stack === 'string') {  
```

alenakhineika: nice!! thank you for the suggestion! ❤️

src/test/suite/connectionController.test.ts
alenakhineika: We use these enums in LegacyConnectionModel type, I removed ConnectionModel in this PR and replaced it with LegacyConnectionModel

addaleax: I’m not sure I understand fully – in `LegacyConnectionModel`, these are also string constant union types:

https://github.com/mongodb-js/compass/blob/0c41452ccc576bfbcc77f445a5a26b2ef289b4f2/packages/data-service/src/legacy/legacy-connection-model.ts#L37-L43

addaleax: Oh wait, I missed that the vscocde extension has its own `LegacyConnectionModel` … nvm then :)

addaleax (approved):

PR #398, opened 2022-03-01
fix(ci): add node-gyp workaround script

leorossi:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/.github/workflows/node-gyp-workaround.sh
+++ b/.github/workflows/node-gyp-workaround.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+# This is a workaround for a node-gyp bug that has not fully been investigated
+# due to problems reproducing it outside of CI environments (even though it
+# occurs both in evergreen and github actions).
+# Something seems to go wrong when node-gyp extracts the Node.js header tarball,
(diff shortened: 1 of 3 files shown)


.github/workflows/node-gyp-workaround.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
```
addaleax, line 1: maybe we could just do `curl -sSfLO https://raw.githubusercontent.com/mongodb-js/compass/42e6142ae08be6fec944b80ff6289e6bcd11badf/.evergreen/node-gyp-bug-workaround.sh && bash node-gyp-bug-workaround.sh` to use the script from Compass as a single source of truth? hopefully this isn’t something that we have to maintain in the long run, but … who knows 🥴

Anemy: There was a `node-gyp` release https://github.com/nodejs/node-gyp/releases 13 hours ago, might be getting solved soon. 🤞

PR #400, opened 2022-03-01
fix: update webpack to handle native dependencies correctly

Anemy:
Uses fix from https://github.com/mongodb-js/compass/pull/2793/files


alenakhineika (approved):
🚀

addaleax (approved):

PR #402, opened 2022-03-15
fix(shell): launch shell does not work with SSH VSCODE-327

alenakhineika:
## Description

Add proxy options to the current connection string to launch shell connected to a remote host, instead of localhost.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/telemetry/telemetryService.ts
```diff
@@ -162,13 +157,9 @@ export default class TelemetryService {
...
-      const identity: AnalyticsNodeIdentity = {
-        anonymousId: this._segmentAnonymousId
-      };
-      if (this._segmentUserId) {
-        identity.userId = this._segmentUserId;
-      }
-      this._segmentAnalytics.identify(identity);
+      const segmentProperties = this.getTelemetryUserIdentity();
+      this._segmentAnalytics.identify(segmentProperties);
+      log.info('TELEMETRY identify', segmentProperties);
```
Anemy, line 162: These changes seem unrelated, is it drive by cleanup?

src/connectionController.ts
```diff
@@ -713,6 +713,43 @@ export default class ConnectionController {
       : '';
   }
 
+  _getConnectionStringWithProxy(mongoClientConnectionOptions: { url: string; options: MongoClientOptions; }): string {
```
Anemy, line 716: Should we also use this method when we `Copy connection string`?
https://github.com/mongodb-js/vscode/blob/main/src/connectionController.ts#L725

src/telemetry/telemetryService.ts
alenakhineika: Yeah clean up to make it closed to the recent fix in mongosh: https://github.com/mongodb-js/mongosh/pull/1229/files#diff-ac2009df67ba6eafe8b274126601bb9491e31f6920c72bbd1b1c33d34864e237R78-R86

src/connectionController.ts
alenakhineika: We can copy a connection string for any connection on a sidebar, not only the active one. But proxy options are available only for the current connection, therefore we don't change the copy connection string behaviour.

Anemy (approved):
:shipit:

PR #403, opened 2022-03-22
chore: infer MongoClientOptions type instead of importing it

addaleax:
Rather than importing it, define it in a slightly more reliable way
by referring to the second argument of `CliServiceProvider.connect()`.

For example, this also enables usage of `tlsUseSystemCA`, which is
included in `DevtoolsConnectOptions` (which is now the actual type
accepted by `CliServiceProvider.connect()`) but not in the driver's
`MongoClientOptions`.

This is for fixing integration with https://github.com/mongodb-js/mongosh/pull/1241.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


Anemy (approved):
👍

PR #409, opened 2022-04-13
feat(playground): Add `clusteredIndex` option to createCollection playground template VSCODE-330

Anemy:
VSCODE-330

This pr adds the `clusteredIndex` option to the `createCollection` playground template. This template is opened when a user clicks to create a new database or collection. 

This pr also changes the behavior around where we would previously show the `time-series` collection creation configuration. Previously we had a separate template with these options more isolated, this pr removes the extra template and instead adds all of the options to one template. 

~Since there are some behavioral changes I'm going to sync with Max before we merge this. I think we wanted `time-series` collections to be hidden until world in the past, however I think this had the side effect that after world, users on earlier MongoDB versions would not see that `time-series` collections are a thing and they might want to upgrade/investigate.~ Synced w/ Max, he's cool with removing the version gate on showing this feature.

I opted to always show the two with a comment about the required version. I was aiming to bring it inline with the template provided in the documentation: https://www.mongodb.com/docs/v5.3/reference/method/db.createCollection/

We're also adding a link to the `createCollection` documentation in the template for `createCollection` in this pr.

https://user-images.githubusercontent.com/1791149/163245903-9fcfead4-1a0d-4785-9812-24340daf7332.mp4


src/templates/playgroundCreateCollectionTemplate.ts
```diff
@@ -25,9 +25,19 @@ db.createCollection(collection);
...
     collation: <document>,
-    writeConcern: <document>
+    writeConcern: <document>,
+    timeseries: { // Added in MongoDB 5.0
+      timeField: <string>, // required for time series collections
+      metaField: <string>,
+      granularity: <string>
+    },
+    expireAfterSeconds: <number>,
+    clusteredIndex: <document>, // Added in MongoDB 5.3
```
lerouxb, line 35: I see for timeseries we list some options but for clusteredIndex and a few others we just say `<document>`. Wondering if we should put more stuff in there? Maybe best like this so they have to read the documentation.

Anemy: We're following the docs there: https://www.mongodb.com/docs/v5.3/reference/method/db.createCollection/
Which also only expands the time-series document. I'm not totally sure on the reasoning there. I think it's alright though, since these are more complex configurations I think expanding all of them might make us have to update this more frequently, and could possibly lead some users astray on differing mongodb server versions.

Issue #415, opened 2022-05-23
Feature Request: Command to generate ObjectId


SethFalco:
## Feature Request

Could a command be added to generate an ObjectId and insert it into the editor? 🤔 

## Detailed Description

I'm proposing 2 new commands:
```json
{
  "command": "mdb.insertObjectIdToEditor",
  "title": "MongoDB: Insert ObjectId to Editor"
},
{
  "command": "mdb.generateObjectIdToClipboard",
  "title": "MongoDB: Generate ObjectId to Clipboard"
}
```

Where the first one will insert an ObjectId to the editor on all cursors (or replacing all selections if there are any).

Where there is more than 1 cursor or selection, it's unclear if the same or unique IDs should be generated for each position. We could either just handle 1 or the other, create 2 commands, or add an option in the settings to change behavior. Your choice what you think is best! 

## Context

Sometimes while working I need (or rather want) to generate an ObjectId and insert it into the code. Usually, this is while creating fixtures for testing or sharing examples of models.

## Possible Implementation

Note this solution does not handle multiple cursors, and doesn't replace selections. It was just a quick thing I threw together to suggest it. I'm happy to implement this myself though if you wouldn't mind having it as part of the extension.

```ts
import { ObjectId } from 'bson';

this.registerCommand(
  EXTENSION_COMMANDS.MDB_INSERT_OBJECTID_TO_EDITOR,
  async (): Promise<boolean> => {
    const editor = vscode.window.activeTextEditor;

    if (!editor) {
      void vscode.window.showInformationMessage('No active editor to insert to.');
      return false;
    }

    await editor.edit((editBuilder) => {
      editBuilder.insert(editor.selection.active, new ObjectId().toHexString());
    });
    return true;
  }
);
this.registerCommand(
  EXTENSION_COMMANDS.MDB_GENERATE_OBJECTID_TO_CLIPBOARD,
  async (): Promise<boolean> => {
    await vscode.env.clipboard.writeText(new ObjectId().toHexString());
    void vscode.window.showInformationMessage('Copied to clipboard.');

    return true;
  }
);
```

PR #416, opened 2022-05-24
feat: add command to generate objectid

SethFalco:
## Related
* Closes https://github.com/mongodb-js/vscode/issues/415

## Description

Adds 2 commands and 1 configuration setting.

`mdb.insertObjectIdToEditor` - Generates an ObjectId and inserts it to the editor.

`mdb.generateObjectIdToClipboard` - Generates an ObjectId and saves it to the clipboard.

`mdb.uniqueObjectIdPerCursor` - If we should use the same or unique ObjectId when `mdb.insertObjectIdToEditor` is called with multiple cursors/selections.

### Checklist

## Motivation and Context

Not really a problem, but I'd just find it pretty convenient if I could generate and insert ObjectIds into the editor through a command.

This is usually while creating fixtures for testing or sharing examples of models.

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

Peek 2022-05-24 22-31

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,8 @@
     "onCommand:mdb.removeConnection",
     "onCommand:mdb.openMongoDBShell",
     "onCommand:mdb.saveMongoDBDocument",
+    "onCommand:mdb.insertObjectIdToEditor",
+    "onCommand:mdb.generateObjectIdToClipboard",
     "onView:mongoDB",
     "onView:mongoDBConnectionExplorer",
     "onView:mongoDBPlaygroundsExplorer",
@@ -393,6 +395,14 @@
           "light": "images/light/plus-circle.svg",
           "dark": "images/dark/plus-circle.svg"
         }
+      },
+      {
+        "command": "mdb.insertObjectIdToEditor",
+        "title": "MongoDB: Insert ObjectId to Editor"
+      },
+      {
+        "command": "mdb.generateObjectIdToClipboard",
+        "title": "MongoDB: Generate ObjectId to Clipboard"
       }
     ],
     "menus": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -896,6 +906,11 @@
           "type": "boolean",
           "default": true,
           "description": "Use default template for playground files."
+        },
+        "mdb.uniqueObjectIdPerCursor": {
```
Anemy, line 910: It may be a bit redundent, but we've been keeping a list of the extension settings in the README as well:
https://github.com/mongodb-js/vscode/blob/6357efaabaf57a287f042087cec0f9598d789e18/README.md?plain=1#L67
Can we add this there to keep it consistent?

SethFalco:
Honestly, the multi-cursors usage is not something I **had** plans to use, so I wasn't sure what would be the best default: 
* Only implemented it because I thought a new user would expect it to work. Otherwise, you might get bug reports about it later.
* Added the setting because users might have different ideas on how it should work. And then you might get bug reports due on what they think the expected behavior should be. Figured it's easier to just support both so if that happens you could redirect them to the setting.

I just used `false` by default because it was lighter (i.e. only need to generate 1 ObjectId). In practice even with 10,000 cursors, both take a negligible amount of time, though.

I don't really see a situation where I'd want to replace multiple instances of an arbitrary ID with another arbitrary ID. IDs generated by this extension aren't intended to match a document in the database, so they should never really require updating in bulk.

However, the use-case I can see myself having after giving it some thought. When creating static fixtures, so not using a factory or a library like faker, usually one creates one object, then copy and pastes it however many times it's needed. Those who know that this command exists would probably favor replacing all IDs at once so each fixture gets a unique one.

Happy to update the default value to false based on this.

Anemy (approved):
Sounds fine to leave it as it is, like you said, it is a bit lighter.  Folks can always change it. Thanks for the PR!

SethFalco:
README updated and rebased with `main`. 👍🏾 
Hopefully should be cool now!

PR #417, opened 2022-05-25
feat: support getCollection for names and shell methods

forivall:
Currently this implementation is unpolished, but it works fine in my local install.

## Description
Parsed out calls to `getCollection` so that they will also be given completion. Also, detect when the collection is being typed out as a string so that completion works properly. 

### Checklist

I'll do these if the feature is desired.

## Motivation and Context

The default "Search For Documents" option on a collection creates a playground document with `db.getCollection('collection_name')`, but the language service only provides completion when accessing the collection like `db.collection_name`.

- [x] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions
I'm not sure about the names of the props in the `CompletionState`, so some feedback there would be appreciated.

The tests to filter on the correct AST node generate high complexity in the eslint complexity check, so I did a bit of refactoring. Methods could be cleaned up in general. Not sure how far i should go, or the style i should use in appeasing eslint there.

The commits could also be split up a bit more. Let me know if you want that.

In the longer term, I'd suggest moving towards using, or at least supporting a typescript plugin, but that's out of scope for this simpler change. I'd also like editor suggestions for query parameters, but that might be more maintainable in typescript definitions. At least, I'll create a feature request for it. I saw that there's definitions for those completions in `mongodb-ace-autocompleter`, but the operators dont have the same links to documentation that the aggregation pipeline stages do.

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/language/mongoDBService.ts
+++ b/src/language/mongoDBService.ts
@@ -344,7 +344,7 @@ export default class MongoDBService {
         worker.postMessage(ServerCommands.GET_FIELDS_FROM_SCHEMA);
 
         worker.on('message', (response: [Error, CompletionItem[] | []]) => {
-          const [error, result] = response;
+          const [error, result = []] = response;
 
           if (error) {
             this._connection.console.log(`SCHEMA error: ${util.inspect(error)}`);
@@ -403,7 +403,8 @@ export default class MongoDBService {
   _prepareCollectionsItems(
     textFromEditor: string,
     collections: Array<CollectionItem>,
-    position: { line: number; character: number }
+    position: { line: number; character: number },
+    isProperty = true,
   ): CompletionItem[] {
     if (!collections) {
       return [];
@@ -411,8 +412,8 @@ export default class MongoDBService {
 
     this._connection.console.log(`collections: ${util.inspect(collections)}`);
 
-    return collections.map((item) => {
-      if (this._isValidPropertyName(item.name)) {
+    return collections.map((item): CompletionItem => {
+      if (!isProperty || this._isValidPropertyName(item.name)) {
         return {
           label: item.name,
           kind: CompletionItemKind.Folder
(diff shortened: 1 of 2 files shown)


alenakhineika:
Hey @forivall, thank you for opening the PR! This solution can work, but as you mentioned, chosen names and the visiting flow are a bit confusing. For example, you introduce the `isProperty` argument that you check later inside the `if` statement:

```ts
if (!isProperty || this._isValidPropertyName(item.name)) {}
``` 

So people could read as "if this is not a property but a valid property name". Hard to guess what to expect here, right?

In the visitor itself, you introduce an extra complexity of identifying your node. You want to check if you are inside a string literal (by the way you should also check a string template here), and this is a call expressing with the `getCollection ` property name. But to get this information, you change other not related visitors, like shell methods and db call (what actually breaks a shell completion case and is caught by one of the tests).

You also add the `isCollectionString` state, again the name is misleading and we don't really need to export this info from the visitor.

You could consider taking the `_checkIsUseCall` as an example and build the `_checkIsGetCollectionCall` visitor. To find your node you can check if it has the `getCollection` property name and the string argument includes a trigger character (see the `_checkIsGetCollectionCall` function as an example).

The completion tests live in the `./src/test/suite/language/mongoDBService.test.ts` file. The broken test I have mentioned above is one of those. When you finish your implementation, please add additional tests to this file to check a new completion use case.

Feel free to reach out with questions! Sorry for such a later response this time, we will be faster next time :)

mmarcon:
Hi @forivall! First of all, thank you again for your contribution!

I would love to get this merged: were you planning to rework this PR based on @alenakhineika's recommendations above?

Also, if you do intend to get this PR merged, could you please sign our [contributor's agreement](https://www.mongodb.com/legal/contributor-agreement) if you have not done so already?

Thank you!

forivall:
Sure, I do still want to rework the PR, I think i just forgot about this one as my day-to-day shifted off of working directly with the mongo shell.

alenakhineika:
Hey @forivall, thank you for putting in the effort to bring this feature to life. However, due to timing constraints, we had to take over and implement it ourselves as part of the other visitor module improvements. We will be happy to see more PRs from you in the future!

forivall:
no worries; yeah, i just didnt have the time to get to it on any of my weekends.

PR #419, opened 2022-05-28
feat: copy id or document to clipboard

SethFalco:
## Description

Adds 2 new context menu actions.

* `mdb.copyDocumentIdFromTree` - To copy the ID of a document to the clipboard.
* `mdb.copyDocumentFromTree` - To copy the value of a document to the clipboard.

Sometimes I can recognize an ID, or especially for small collections (i.e. during development) I know the collection at the bottom is the one I need to interest with. It'd be nice to be able to just copy the ID directly from the UI to be able to paste/use in the platground editor.

### Checklist
  * Not sure if we need tests for this one?

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/mdbExtensionController.ts
```diff
@@ -487,6 +487,38 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+      EXTENSION_COMMANDS.MDB_COPY_DOCUMENT_ID_FROM_TREE,
+      async (element: DocumentTreeItem): Promise<boolean> => {
+        const { documentId } = element;
+
+        if (!documentId) {
+          void vscode.window.showWarningMessage('Can\'t copy ID from document that has no ID.');
+          return false;
+        }
+
+        let body = JSON.stringify(documentId);
```
addaleax, line 500: What happens here if `documentId` contains non-JS bson datatypes, e.g. `Binary()` or similar? Should this be EJSON instead of JSON?

src/mdbExtensionController.ts
```diff
@@ -487,6 +487,38 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+
+        if (body.startsWith('"')) {
+          body = body.slice(1, body.length - 1);
+        }
```
addaleax, line 504: If the idea here is to leave string document ids as-is, it might be better to check e.g. `if (typeof body !== 'string') { body = JSON.stringify(documentId); }` instead, since this currently still leaves JSON escapes in `body`

src/mdbExtensionController.ts
```diff
@@ -487,6 +487,38 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+      }
+    );
+    this.registerCommand(
+      EXTENSION_COMMANDS.MDB_COPY_DOCUMENT_FROM_TREE,
+      async (element: DocumentTreeItem): Promise<boolean> => {
+        const { document } = element;
+        const tabSize = vscode.workspace
+          .getConfiguration('editor')
+          .get('tabSize', 2);
+        const body = JSON.stringify(document, undefined, tabSize);
```
addaleax, line 517: Similar question here :)

addaleax (commented):
Thank you for the PR! I think we’d need at least two things more, both for this one and for #420:
- They need tests :)
- And you need to sign the [MongoDB Contributor’s Agreement](https://www.mongodb.com/contributor) if you haven’t already

Anemy:
Added view document and copy document in this pr: https://github.com/mongodb-js/vscode/pull/445 - it'll be in the next release. No copy id, yet, though.

PR #420, opened 2022-05-28
fix: send 1 notification switch connection

SethFalco:
## Description

Previously when one had 2 connections, switching creates 2 notifications.

image

This isn't a major problem, but is a little annoying.

This instead adds a `quiet` argument to `#disconnect`, and if there was a previous connection uses a single message to convey both actions.

image

### Checklist

Not sure if this PR requires them as this only really updates the use of `showInformationMessage` which is mocked in the tests anyway.

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -393,8 +393,10 @@ export default class ConnectionController {
     this._connectingConnectionId = connectionId;
     this.eventEmitter.emit(DataServiceEventTypes.CONNECTIONS_DID_CHANGE);
 
-    if (this._activeDataService) {
-      await this.disconnect();
+    const oldConnectionId = this._currentConnectionId;
+
+    if (oldConnectionId) {
+      await this.disconnect(true);
     }
 
     this._statusView.showMessage('Connecting to MongoDB...');
@@ -433,7 +435,17 @@ export default class ConnectionController {
     }
 
     log.info('Successfully connected');
-    void vscode.window.showInformationMessage('MongoDB connection successful.');
+
+    if (oldConnectionId) {
+      const currentConnectionName = oldConnectionId
+        ? this.getSavedConnectionName(oldConnectionId)
+        : 'previous server';
+      const nextConnectionName = this.getSavedConnectionName(connectionId);
+
+      void vscode.window.showInformationMessage(`Disconnected from ${currentConnectionName} and connected to ${nextConnectionName}.`);
+    } else {
+      void vscode.window.showInformationMessage('MongoDB connection successful.');
+    }
 
     this._activeDataService = newDataService;
     this._currentConnectionId = connectionId;
@@ -491,7 +503,11 @@ export default class ConnectionController {
     }
   }
 
-  async disconnect(): Promise<boolean> {
+  /**
+   * @param quiet Don't display non-error notifications to the user.
+   * @returns If we disconnected from MongoDB succesfully.
+   */
+  async disconnect(quiet = false): Promise<boolean> {
     log.info(
       'Disconnect called, currently connected to:',
       this._currentConnectionId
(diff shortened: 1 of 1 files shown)


src/connectionController.ts
```diff
@@ -433,7 +435,17 @@ export default class ConnectionController {
...
+
+    if (oldConnectionId) {
+      const currentConnectionName = oldConnectionId
+        ? this.getSavedConnectionName(oldConnectionId)
+        : 'previous server';
+      const nextConnectionName = this.getSavedConnectionName(connectionId);
+
+      void vscode.window.showInformationMessage(`Disconnected from ${currentConnectionName} and connected to ${nextConnectionName}.`);
+    } else {
+      void vscode.window.showInformationMessage('MongoDB connection successful.');
```
Anemy, line 447: I like the change in the new messages to use the connection name. Would it be possible to do the same in the existing message?
```suggestion
      const connectionName = this.getSavedConnectionName(connectionId);

      void vscode.window.showInformationMessage(connectionName ? `Connected to ${connectionName}.` : 'MongoDB connection successful.');
```

Anemy (approved):
lgtm! One suggestion on a possible improvement.

SethFalco:
I noticed you merged one of my PRs!
Just wanted to apologize for putting off these 2 that are still pending some work before they're ready to be reviewed. I'll try to get back to them tomorrow if I have time!

alenakhineika:
Hey @SethFalco, do you feel like picking up these PRs? They have been open for quite a while :) Or can we safely close them at this point?

SethFalco:
@alenakhineika Ahh, sorry about that! I actually forgot about this. A bit too much on my plate. 😓

If you don't mind, I would still like to address these PRs as I still use the extension. I can rebase and address it over the weekend.

And since last time I left maintainers hanging… if I don't update the PR for whatever reason by next week, feel free to close it. When I have time, I'm happy to open a new PR and reference this one.

SethFalco:
I've amended the commit to mention the connection name on connect, disconnect, and both (switching between two connections).

Any other feedback is welcome!

src/connectionController.ts
```diff
@@ -327,11 +327,14 @@ export default class ConnectionController {
...
 
+    const prevConnectionId = this._currentConnectionId;
+    const nextConnectionName = this.getSavedConnectionName(connectionId);
+
     if (this._activeDataService) {
       log.info('Disconnecting from the previous connection...', {
         connectionId: this._currentConnectionId,
       });
-      await this.disconnect();
+      await this.disconnect(true);
```
alenakhineika, line 337:
```suggestion
      await this.disconnect({ quiet: true });
```
I would prefer having an object as an argument here, so the intention is easier readable from the code.

src/connectionController.ts
```diff
@@ -418,7 +421,20 @@ export default class ConnectionController {
...
     log.info('Successfully connected', { connectionId });
-    void vscode.window.showInformationMessage('MongoDB connection successful.');
+
+    if (prevConnectionId) {
+      const prevConnectionName = prevConnectionId
+        ? this.getSavedConnectionName(prevConnectionId)
+        : 'MongoDB server';
+
+      void vscode.window.showInformationMessage(
+        `Disconnected from ${prevConnectionName} and connected to ${nextConnectionName}.`
```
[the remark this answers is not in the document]
SethFalco: Rather than the word `Changed`, how about `Set`?

The word `Changed` implies the presence of a connection before, but if you'd like a single message for both an initial connection and changing connection, I think `Set` is more fitting.

PR #426, opened 2022-07-07
Chores 2

SethFalco:
### Checklist
- N/A New tests and/or benchmarks are included
- N/A Documentation is changed or added

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

---

* The docs are the wrong way around right now, so we need to either switch the items for `Workspace` and `Global` or switch the docs. In this PR I swapped the items since `Global` was default anyway, so I figured it could go first.
* Just merged some imports.

Here is what we currently see:
image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -880,8 +880,8 @@
         "mdb.connectionSaving.defaultConnectionSavingLocation": {
           "type": "string",
           "enum": [
-            "Workspace",
             "Global",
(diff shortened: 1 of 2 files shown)


Anemy (approved):
👍

PR #434, opened 2022-08-31
fix: export to language actions appear for regular playground results VSCODE-334

alenakhineika:
## Description

- Remove export to language code lenses for regular playground results.
- Seems that setting configurations in the `./src/test/suite/index.ts` causes some kind of race condition and vscode rejetcs operation with the `rejected promise not handled within 1 second: Error: Unable to write to Workspace Settings because mdb.connectionSaving.hideOptionToChooseWhereToSaveNewConnections is not a registered configuration.` error for recent versions of vscode. Moved the following part to the `./src/test/suite/connectionController.test.ts` before hook and looks like the error has gone. Let's keep an eye on it.

```ts
const connectionSaving = vscode.workspace.getConfiguration('mdb.connectionSaving');
await connectionSaving.update(
  'hideOptionToChooseWhereToSaveNewConnections',
  true,
  vscode.ConfigurationTarget.Global
);
``` 

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


package.json
```diff
@@ -1,1031 +1,1037 @@
...
+	},
+	"precommit": [
+		"check"
+	],
+	"__metadata": {
+		"id": "4b6540da-bfa3-4b88-b9e3-df8f2d161124",
+		"publisherDisplayName": "MongoDB",
+		"publisherId": "96be5892-f133-40ba-88dd-d450cff649f0",
+		"isPreReleaseVersion": false
+	}
```
Anemy, line 1036: Looks like this file caught some diff from going from 2 space tabs to 4, can we keep it 2 spaces?

alenakhineika: This is super weird because the package.json shows 2 spaces for me.

Screenshot 2022-08-31 at 17 47 59

Anemy (approved):
lgtm! Nice. Left one suggestion on the test setting we can probably remove.

src/test/suite/connectionController.test.ts
```diff
@@ -51,6 +51,18 @@ suite('Connection Controller Test Suite', function () {
...
+    const connectionSaving = vscode.workspace.getConfiguration(
+      'mdb.connectionSaving'
+    );
+    await connectionSaving.update(
+      'hideOptionToChooseWhereToSaveNewConnections',
+      true,
+      vscode.ConfigurationTarget.Global
+    );
+  });
+
```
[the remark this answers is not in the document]
alenakhineika: Niceee! Removed this!

Anemy (approved):

PR #439, opened 2022-09-22
feat: make mongodb-client-encryption optional VSCODE-342

alenakhineika:
## Description

- Bump the `mongodb-data-service` version that makes the `mongodb-client-encryption` module optional. That will ensure that VSCode extension builds are OS-independent.
- Add the `compass-preferences-model` module to webpack externals because it is now an optional dependency of `@mongodb-js/compass-logging` and otherwise causes the `Can't resolve 'compass-preferences-model'` error.
- Replace [output.libraryTarget](https://webpack.js.org/configuration/output/#outputlibrarytarget) with `output.library.type` since webpack might drop support for `output.libraryTarget` in the future.
- Use `commonjs` output type since `commonjs2` is not a thing anymore and probably anyway aliases to `commonjs`.
- Use the proper `os-dns-native` module name instead of underscores.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [x] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


addaleax:
In the ticket, I made the suggestion to use `npm install --omit=optional` – did you look into that? I think that might be a good way to help catch these issues ahead of time in the future.

Anemy (approved):
🔥

addaleax (approved):

PR #452, opened 2022-11-30
feat(tree-explorer): add delete document context menu item VSCODE-349

Anemy:
VSCODE-349

This adds a setting `confirmDeleteDocument` which defaults `true` to show a confirmation when deleting a document. Can add more tests if folks would like that.

https://user-images.githubusercontent.com/1791149/204882698-a93218c1-4933-4769-b38c-84c62715f501.mp4


alenakhineika:
Would it make sense to have this setting not only for deleting documents but for any deleting operation? I am wondering how often users expect a different confirmation behaviour for different entities.

PR #453, opened 2022-12-01
chore: improve disconnected action error messages

Anemy:
Improves the error messages by showing the connection name in a few error messages that occur when the connection has changed, previously we were showing the connection id which doesn't mean much to a user. In the edge case where the old connection was removed and no longer exists we default to showing `the connection` as the name.

| before | after | 
| - | - |
| Screen Shot 2022-12-01 at 1 24 30 PM | Screen Shot 2022-12-01 at 1 26 34 PM |

### DIFF
--- a/src/editors/collectionDocumentsProvider.ts
+++ b/src/editors/collectionDocumentsProvider.ts
@@ -60,12 +60,15 @@ export default class CollectionViewProvider
     // Ensure we're still connected to the correct connection.
     if (connectionId !== this._connectionController.getActiveConnectionId()) {
       operation.isCurrentlyFetchingMoreDocuments = false;
+      const oldConnectionName =
+        this._connectionController.getSavedConnectionName(connectionId || '') ||
+        'the connection';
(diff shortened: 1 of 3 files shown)


src/editors/collectionDocumentsProvider.ts
```diff
@@ -60,12 +60,15 @@ export default class CollectionViewProvider
     // Ensure we're still connected to the correct connection.
     if (connectionId !== this._connectionController.getActiveConnectionId()) {
       operation.isCurrentlyFetchingMoreDocuments = false;
+      const oldConnectionName =
+        this._connectionController.getSavedConnectionName(connectionId || '') ||
+        'the connection';
```
addaleax, line 65: “no longer connected to the connection” sounds a bit odd, right? Maybe this could say “no longer connected to the database”?

Anemy: That reads much better yup. Updated

PR #470, opened 2023-02-01
feat(playgrounds): update playground template VSCODE-337

Anemy:
VSCODE-337
Opening to open a discussion with the team on how we can best supply a playground template for folks.

The changes in this one are:
- Removed the drop collection call. I'm thinking it's something that's a bit risky to have in a default script. It's a useful command, but I don't think it's something folks want to have a possibility of accidentally running.
- Removed the _id number usage (so we default the _id to ObjectId).
- Changed how we use collection methods collection methods from db.sales.find  to db['sales'].find to reduce possible rough spots with folks that have spaces and dashes in their collection names. EDIT: Based on Anna's reply: https://github.com/mongodb-js/vscode/pull/470#discussion_r1097199338 I'm thinking let's update it to db.getCollection('sales'). instead. I don't think our autocomplete for collection names works there but it's something we can add.
- Explicitly mentioned console.log usage and added a logged message. I think folks will want to debug their playgrounds often, and if they're not familiar with shell scripting or javascript environments this might not be intuitive without a callout.
- Explicitly mentioned cursor 20 document results higher up in the template. I'm under the impression people who don't use the shell often might not expect cursor results to be shown.

### DIFF
--- a/src/templates/playgroundTemplate.ts
+++ b/src/templates/playgroundTemplate.ts
@@ -5,6 +5,8 @@ const template = `// MongoDB Playground
 // The result of the last command run in a playground is shown on the results panel.
 // By default the first 20 documents will be returned with a cursor.
 // Use 'console.log()' to print to the debug output.
+// For more documentation on playgrounds please refer to
+// https://www.mongodb.com/docs/mongodb-vscode/playgrounds/
 
 // Select the database to use.
 use('mongodbVSCodePlaygroundDB');
@@ -24,10 +26,10 @@ db.getCollection('sales').insertMany([
 // Run a find command to view items sold on April 4th, 2014.
 const salesOnApril4th = db.getCollection('sales').find({
   date: { $gte: new Date('2014-04-04'), $lt: new Date('2014-04-05') }
-}).toArray();
+}).count();
 
 // Print a message to the output window.
-console.log(\`\${salesOnApril4th.length} sales occurred in 2014.\`);
+console.log(\`\${salesOnApril4th} sales occurred in 2014.\`);
 
 // Here we run an aggregation and open a cursor to the results.
 // Use '.toArray()' to exhaust the cursor to return the whole result set.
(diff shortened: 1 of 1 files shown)


src/templates/playgroundTemplate.ts
```diff
@@ -2,39 +2,42 @@ const template = `// MongoDB Playground
...
-db.sales.insertMany([
-  { '_id': 1, 'item': 'abc', 'price': 10, 'quantity': 2, 'date': new Date('2014-03-01T08:00:00Z') },
-  { '_id': 2, 'item': 'jkl', 'price': 20, 'quantity': 1, 'date': new Date('2014-03-01T09:00:00Z') },
-  { '_id': 3, 'item': 'xyz', 'price': 5, 'quantity': 10, 'date': new Date('2014-03-15T09:00:00Z') },
-  { '_id': 4, 'item': 'xyz', 'price': 5, 'quantity':  20, 'date': new Date('2014-04-04T11:21:39.736Z') },
-  { '_id': 5, 'item': 'abc', 'price': 10, 'quantity': 10, 'date': new Date('2014-04-04T21:23:13.331Z') },
-  { '_id': 6, 'item': 'def', 'price': 7.5, 'quantity': 5, 'date': new Date('2015-06-04T05:08:13Z') },
-  { '_id': 7, 'item': 'def', 'price': 7.5, 'quantity': 10, 'date': new Date('2015-09-10T08:43:00Z') },
-  { '_id': 8, 'item': 'abc', 'price': 10, 'quantity': 5, 'date': new Date('2016-02-06T20:20:13Z') },
+db['sales'].insertMany([
```
addaleax, line 13: If we want “proper” programmatic usage, this is probably the right way to go, since it handles leading underscores well + could potentially be used with TS (if there were public TS definitions for mongosh):

```suggestion
db.getCollection('sales').insertMany([
```

mcasimir: What is the reason for this change?

addaleax: From Slack:

> Changed how we use collection methods collection methods from `db.sales.find` to `db['sales'].find` to reduce possible rough spots with folks that have spaces and dashes in their collection names.

alenakhineika: I like Anna's suggestion!

Anemy: Just added it to the description, sorry only had it in the slack.
One thing to note is that `db.getCollection('...')` doesn't suggest collection names currently. It won't be too hard to add though. I'll create a ticket to add it once we decide if we add it here.

Anemy:
Got the 👌 from Max, so I'm going to merge, feel free to make or suggest any other changes folks want down the line!

PR #479, opened 2023-02-14
feat(playground): makes creation of columnstore indexes more visible in index creation template VSCODE-364

himanshusinghs:
## Description
VSCODE-364

This PR adds makes creation of `columnstore` indexes more visible in index creation template. Additionally we also surface `columnstoreProjection` as available option, supported alongside `columnstore` indexes.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/templates/playgroundCreateIndexTemplate.ts
+++ b/src/templates/playgroundCreateIndexTemplate.ts
@@ -19,7 +19,7 @@ db.getCollection('CURRENT_COLLECTION')
        * '$**': 1, // wildcard index on all fields and subfields in a document
        * 'path.to.field.$**': 1 // wildcard index on a specific field and its subpaths
        *
-       * Columnstore index // added in MongoDB 6.3
+       * Columnstore index
        * '$**': 'columnstore', // columnstore index on multiple specific field
        * 'path.to.field.$**': 'columnstore', // columnstore index on one field and all the subfields
        *
--- a/src/templates/playgroundCreateIndexTemplate.ts
+++ b/src/templates/playgroundCreateIndexTemplate.ts
@@ -12,16 +12,16 @@ db.getCollection('CURRENT_COLLECTION')
        * Keys
        *
        * Normal index
-       * fieldA:  1, //ascending
-       * fieldB: -1  //descending
+       * fieldA:  1, // ascending
+       * fieldB: -1  // descending
        *
        * Wildcard index
-       * '$**': 1, //wildcard index on all fields and subfields in a document
-       * 'path.to.field.$**': 1 //wildcard index on a specific field and its subpaths
+       * '$**': 1, // wildcard index on all fields and subfields in a document
+       * 'path.to.field.$**': 1 // wildcard index on a specific field and its subpaths
        *
-       * Columnstore index //added in MongoDB 6.3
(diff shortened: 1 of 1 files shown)


src/templates/playgroundCreateIndexTemplate.ts
```diff
@@ -19,6 +19,10 @@ db.getCollection('CURRENT_COLLECTION')
        * '$**': 1, //wildcard index on all fields and subfields in a document
        * 'path.to.field.$**': 1 //wildcard index on a specific field and its subpaths
        *
+       * Columnstore index //added in MongoDB 6.3
+       * '$**': "columnstore", //columnstore index on multiple specific field
+       * 'path.to.field.$**': "columnstore", //columnstore index on one field and all the subfields
```
Anemy, line 24: Definitely a small thing/nit, how do you feel about updating the comments in this template to have a space after the `//`, before the first word in the comments?
Like: 
```suggestion
       * 'path.to.field.$**': "columnstore", // columnstore index on one field and all the subfields
```

Anemy (approved):

src/templates/playgroundCreateIndexTemplate.ts
alenakhineika: @Anemy the whole templated misses spaces after `//`. @himanshusinghs could you please fix them everywhere? :)

alenakhineika: I also think maybe we could write something like `Columnstore index (added in MongoDB 6.3)`. This `//` in the section title looks a bit weird, even tho the rest of the comment block is also not a valid js 🙈

himanshusinghs: About the space before `//`, I was thinking why it was different compared to other files but then I just decided to follow how it was in the file. Definitely will update as per suggestions:
- space after comments
- single quotes

himanshusinghs: Ohh I just saw the recent comments. I have cleaned up the entire file and verified that there no other places with comment problem.

About the `added in MongoDB 6.3` - I added that because the release itself is still not available. For that reason maybe it makes sense to keep it but totally agreed with the remark about "valid JS" - that is something I can fix 😄

himanshusinghs: Ok looks like we do the invalid JS style in one other template as well. I would maybe raise another PR fixing that across all templates. Let me know if that's alright.

alenakhineika: I would still replace `Columnstore index // added in MongoDB 6.3` with just `Columnstore index` to reduce the number of comments and make it more readable. It feels like `columnstoreProjection: {} //added in MongoDB 6.3` below is enough to empathize that this is added in 6.3. @mmarcon, what would be your preference here? We had a similar conversation recently regarding these comments in templates.

himanshusinghs: Yea that make sense - I missed the part last time about `columnstoreProjection` already mentioning 6.3
Will do the needful.

alenakhineika (approved):

Issue #485, opened 2023-02-23
Sort databases by name


miqmago:
## Feature Request

Sort databases from a connection by name.

## Detailed Description

When listing databases of a connection they seem not to be sorted by any alphabetic order.

## Context

This is specially important when connecting to local machine where sometimes there are many databases from different projects, it is difficult to find the collection by name.

I've also tried to press the first letter of the database but it does not seem to work very well, as it mixes databases with collections and it is difficult to find the database.

Anemy:
@miqmago can you tell us a little about what you're seeing or perhaps a screenshot? Mine appear to be sorted - which MongoDB version are you using?
<img width="288" alt="Screenshot 2023-02-24 at 1 00 38 AM" src="https://user-images.githubusercontent.com/1791149/221103862-cf1f4c82-eb36-4c72-bce2-b4510e98daa7.png">

We can add [a fix](https://github.com/mongodb-js/vscode/compare/sort-dbs?expand=1), I'm mostly curious on why there's a discrepancy with the naming between our systems.

stoberov:
@Anemy - I too can report the same issue. I just installed the extension and connected to one of our clusters at work (with 100s of databases) => the list in VSCode extension is then absolutely non-sorted.

One difference between you and me would be that you're connecting to a local instance of mongo, if that makes a difference.

Another possibility could be this `.map` over here messing up the order when there are more databases to make a difference - https://github.com/mongodb-js/vscode/blob/main/src/explorer/connectionTreeItem.ts#L104

I believe your suggested fix would be **great**.

Anemy:
Merged in the fix and released in the latest version of the extension, thanks again for opening the issue! @miqmago  @stoberov

PR #488, opened 2023-02-28
feat(tree-explorer): sort dbs in the tree by name

Anemy:
Looks like we were defaulting to the server's returned ordering, this pr makes the databases in the list sorted. While most of the databases returned from MongoDB are already sorted, `admin`, `local`, and `config` appear before others alphabetically. These changes make those databases appear alphabetically as well as ensures databases that might return databases differently, (maybe Atlas Data Federation or Cosmos), order the same.

Reported in https://github.com/mongodb-js/vscode/issues/485 

Before merging I'll add a test.

### DIFF
--- a/src/explorer/connectionTreeItem.ts
+++ b/src/explorer/connectionTreeItem.ts
@@ -150,6 +150,9 @@ export default class ConnectionTreeItem
     }
 
     const databases = await this.listDatabases();
+    databases.sort((a: string, b: string) => {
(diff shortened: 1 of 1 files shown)


lerouxb:
Do we want this? Feels right to me that the "system" dbs are listed first or at least separately without being mixed in with the others. Might be worth bringing up with product/design.

alenakhineika (approved):

PR #489, opened 2023-03-01
refactor: tests and commands clean up VSCODE-383 Tests and command palette clean up

alenakhineika:
## Description

- Mock VSCode messages to fix flaky tests.
- Get rid of some `any` types in tests.
- Clean up command palette.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


package.json
```diff
@@ -644,23 +636,23 @@
         },
         {
           "command": "mdb.exportToRuby",
-          "when": "mdb.isPlayground == true"
+          "when": "mdb.isPlayground == true && connectedToMongoDB == true"
```
addaleax, line 639: Maybe a dumb question, I’m not sure how `when` contents are being evaluated, but since this _looks_ like a JS-like language, should this be the following?

```suggestion
          "when": "mdb.isPlayground == true && mdb.connectedToMongoDB == true"
```

src/test/suite/commands/launchMongoShell.test.ts
```diff
@@ -1,131 +1,128 @@
...
+  let createTerminalStub: SinonStub;
+  let sendTextStub: SinonSpy;
 
   beforeEach(() => {
-    sandbox.stub(vscode.window, 'showInformationMessage');
-
-    fakeShowErrorMessage = sandbox.stub(vscode.window, 'showErrorMessage');
-    fakeGetMongoClientConnectionOptions = sandbox.stub(
-      mockConnectionController,
+    sinon.stub(vscode.window, 'showInformationMessage');
```
addaleax, line 21: Why are we moving away from using `sandbox` at the top level here? If the goal here is for other tests outside this file to also be affected by this line, then this stubbing should be added to the relevant tests (or, if necessary, as a global `beforeEach`/`afterEach` hook pair)

alenakhineika: Even tho in this particular case, the clean up is handled in afterEach hook, I decided to move everything to sandbox as we discussed. I find tests now cleaner :) thx for sharing your point of view in DM!

package.json
alenakhineika: Thx! copy pasted without a prefix. This is what `when` is if you are curious: https://code.visualstudio.com/api/references/when-clause-contexts

addaleax (approved):

PR #490, opened 2023-03-02
feat: show MongoDB completion items before other JS completion VSCODE-382, VSCODE-385

alenakhineika:
## Description

- Inside aggregations and queries show fields completion before other JS things completion.
- e2e test setup.
- Tests clean up (do not multiply language services and providers).

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/language/mongoDBService.ts
```diff
@@ -496,10 +492,6 @@ export default class MongoDBService {
   ): ExportToLanguageMode {
     const state = this._visitor.parseAST(params);
 
-    this._connection.console.log(
-      `EXPORT TO LANGUAGE state: ${util.inspect(state)}`
-    );
-
```
alenakhineika, line 502: Too noisy log.

src/language/worker.ts
```diff
@@ -92,98 +96,85 @@ const executeAll = async (
...
-) => {
+  connectionOptions: ConnectionOptions
+): Promise<[unknown, Document[]?]> => {
   try {
-    const serviceProvider: CliServiceProvider =
-      await CliServiceProvider.connect(connectionString, connectionOptions);
+    const serviceProvider = await CliServiceProvider.connect(
+      connectionString,
+      connectionOptions
+    );
```
addaleax, line 151: I know this is a draft PR and pre-existing code, but I’m a bit surprised here … we are connecting a new `CliServiceProvider` for every autocompletion request? And never (explicitly) closing them? Those seem like bugs to me, might be good to fix that here and/or create follow-up tickets

src/language/worker.ts
```diff
@@ -92,98 +96,85 @@ const executeAll = async (
...
+  let result: string[] = [];
 
-    const result = await findAndParse(
-      serviceProvider,
-      databaseName,
-      collectionName
+  try {
+    serviceProvider = await CliServiceProvider.connect(
+      connectionString,
+      connectionOptions
```
addaleax, line 115: Does this work? The `ConnectionOptions` type from the driver is not something that an application should normally use, and it’s also not what `CliServiceProvider.connect` expects

src/language/worker.ts
alenakhineika: Yeah, this is a bug. Pushed changes, let me know if this is the right way.

src/language/worker.ts
alenakhineika: Replaced with MongoClientOptions defined here in the begging of the file. But answering your question if it worked, weirdly yes)

src/language/worker.ts
addaleax: It answers the second part here, yeah :+1:

alenakhineika: I will probably refactor it in the current pr (not a separate one) to not create it each time. Makes total sense to change it.

src/test/suite/stubs.ts
```diff
@@ -317,7 +308,7 @@ class LanguageServerControllerStub {
     return;
   }
 
-  executeAll(/* codeToEvaluate: string */): Promise<ShellExecuteAllResult> {
+  evalauate(/* codeToEvaluate: string */): Promise<ShellEvaluateResult> {
```
addaleax, line 311:
```suggestion
  evaluate(/* codeToEvaluate: string */): Promise<ShellEvaluateResult> {
```

(similarly elsewhere)

src/telemetry/connectionTelemetry.ts
```diff
@@ -38,75 +36,78 @@ type CloudInfo = {
...
 
+  try {
+    const connectionString = dataService.getConnectionString();
+    const firstHost = connectionString.hosts[0] || '';
+    const [hostname] = firstHost.split(':');
+    const authMechanism = connectionString.searchParams.get('authMechanism');
+    const username = connectionString.username ? 'DEFAULT' : 'NONE';
+    const authStrategy = authMechanism ?? username;
+    const instance = await dataService.instance();
+    const cloudInfo = await getCloudInfoFromDataService(hostname);
```
addaleax, line 90: I know this is existing code, but there's probably no reason not to run this in parallel, right?

```suggestion
    const [instance, cloudInfo] = await Promise.all([
      dataService.instance(), getCloudInfoFromDataService(hostname)
    ]);
```

src/language/mongoDBService.ts
```diff
@@ -153,47 +170,39 @@ export default class MongoDBService {
...
-                printableError.message
-              );
-            }
-
-            void worker.terminate().then(() => {
-              resolve(result);
-            });
-          }
-        );
+        worker?.on('error', (error) => {
```
addaleax, line 179: When can `worker` be false-y here?

.depcheckrc
```diff
@@ -14,3 +14,4 @@ ignores:
  - "vscode-languageserver-textdocument"
  - "vscode"
  - "webpack-cli"
+ - "@mongosh/browser-runtime-electron"
```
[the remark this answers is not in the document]
alenakhineika: This thing is super weird and confuses me. if you delete it from depcheck ignore, then `npm run check fails` with:

```
Unused dependencies
* @mongosh/browser-runtime-electron
```

But is is used in the worker.ts

src/language/mongoDBService.ts
```diff
@@ -203,200 +212,98 @@ export default class MongoDBService {
...
-            `MONGOSH found ${result.length} databases`
-          );
-          this._updateCurrentSessionDatabases(result);
-        });
-      });
+      // TODO: There is a mistake in the service provider interface
+      // Use `admin` as arguments to get list of dbs
+      // and remove it later when `mongosh` will merge a fix.
+      const documents = await this._serviceProvider.listDatabases('admin');
+      result = documents.databases ? documents.databases : [];
```
addaleax, line 244:
```suggestion
      result = documents.databases ?? [];
```

src/language/mongoDBService.ts
```diff
@@ -1,145 +1,162 @@
...
+    // If already connected close the previous connection.
+    await this.disconnectFromServiceProvider();
+
+    this._connectionId = connectionId;
+    this._connectionString = connectionString;
+    this._connectionOptions = connectionOptions;
+    this._serviceProvider = await CliServiceProvider.connect(
+      connectionString,
+      connectionOptions
+    );
```
addaleax, line 97: So, a question: This runs in the main thread, right, not inside a worker? Is there anything preventing us from using the `DataService` created by the `ConnectionController` for the operations for which we currently use the `CliServiceProvider` here? That would mean that we only need to connect once to a server, not twice, from the main thread, which seems reasonable to me

src/language/mongoDBService.ts
```diff
@@ -153,47 +170,39 @@ export default class MongoDBService {
...
+          );
+          void worker.terminate().then(() => {
+            resolve(undefined);
+          });
+        });
+
+        worker?.on('message', (data) => {
+          void worker.terminate().then(() => {
+            resolve(data);
+          });
```
addaleax, line 195: We create 1 Worker + 1 CliServiceProvider per `evaluate()` call here, right?

src/language/mongoDBService.ts
```diff
@@ -203,200 +212,98 @@ export default class MongoDBService {
...
+    }
 
-            return resolve(true);
-          });
-        });
-      } catch (error) {
-        this._connection.console.log(`SCHEMA error: ${util.inspect(error)}`);
+    try {
+      const documents = await this._serviceProvider
+        .find(databaseName, collectionName, {}, { limit: 1 })
```
addaleax, line 289: Unrelated note, should this really be `1`? `parseSchema()` is generally meant for input with more than one document, since otherwise it just gives you the (potentially nested) keys of the document (and it actually accepts a cursor, which means that the `.toArray()` call here might be unnecessary).

src/language/mongoDBService.ts
```diff
@@ -544,146 +446,183 @@ export default class MongoDBService {
...
 
-  _clearCurrentSessionConnection(): void {
+  async _clearCurrentConnection(): Promise<void> {
+    this._connectionId = undefined;
     this._connectionString = undefined;
     this._connectionOptions = undefined;
+
+    if (this._serviceProvider) {
+      await this._serviceProvider.close(true);
+      this._serviceProvider = undefined;
```
addaleax, line 625: Is there any chance that something could call `connectToServiceProvider()` while the `close()` call is active? If so, you’ll probably need to do this to avoid a race condition where `connectToServiceProvider()` finishes before `close()` and you’d accidentally be overriding a _new_ value of `this._serviceProvider` here:

```suggestion
      const serviceProvider = this._serviceProvider;
      this._serviceProvider = undefined;
      await serviceProvider.close(true);
```

addaleax (approved):

src/language/mongoDBService.ts
alenakhineika: As I remember we wanted to make the parsing process faster by parsing one document only. Maybe I miss something, but isn't Compass does the same? It uses just one document for this:
[compass
](https://github.com/mongodb-js/compass/blob/main/packages/compass-field-store/src/stores/store.js#L193)

And then passes it to the `parseSchema`:
https://github.com/mongodb-js/compass/blob/main/packages/compass-field-store/src/stores/store.js#L180

About `toArray` you are right, we can remove it.

src/language/mongoDBService.ts
alenakhineika: Refactored as we discussed during 1-on-1 to always send back a message and not throw the error.

src/language/mongoDBService.ts
addaleax: I see … that should be fixable somehow since we don’t actually make use of the storage-mixin functionality :slightly_smiling_face: If it doesn’t work right now, the Browser-Compatible Data Service project should make this usable here. But yeah, definitely something for a separate PR in any case!

src/language/mongoDBService.ts
addaleax: Yeah, definitely not a blocker for this PR. I think ideally, we’d persist the connection while the playground is opened or something, but that’s not something we need to optimize right now

src/language/mongoDBService.ts
addaleax: Interesting, I’m also surprised that Compass does this then (it even has multiple documents available in the refresh case… it makes sense that when inserting a document, it only processes that document, but otherwise I’m not sure why we use `parseSchema` then… :slightly_smiling_face:)

alenakhineika: I think we use parseSchema with one document to guess how the whole collection can look like and not parse a big amount of data, and also get types and other info that parseSchema results. Is there another way to get a field type?

addaleax: > and also get types and other info that parseSchema results

Yeah, that might change the equation a bit – I’m just surprised because below we do `schema.fields.map((item) => item.name)` to get the result of this function, so it seems like we currently only use the names?

To be clear, I don’t mind using `parseSchema`, I’m mostly just surprised that we’re only passing a single document here

src/language/worker.ts
```diff
@@ -74,168 +67,46 @@ const executeAll = async (
...
-    // Use `admin` as arguments to get list of dbs
-    // and remove it later when `mongosh` will merge a fix.
-    const result = await serviceProvider.listDatabases('admin');
-    const databases = prepareCompletionItems(result);
-
-    return [null, databases];
+    return { data: { outputLines, result } };
   } catch (error) {
-    return [error];
+    return { error };
```
addaleax, line 88: I think depcheck was failing because this was previously `throw new Error((<any>error).message);`, and the parser it uses didn’t understand the `<any>error` cast – you should be able to remove the addition to `.depcheckrc` now that that’s no longer the case

alenakhineika: You are right!! 🔥

src/language/mongoDBService.ts
alenakhineika: I see your point! Do you think it is worth to get rid of parseSchema and just get deep object keys?

addaleax: I’m not sure? Just to reiterate, it’s surprising to me that we don’t use more documents (not a lot – but maybe 5 or 10 or so?) than 1. And if we want to keep the flexibility to do so, then keeping `parseSchema` seems like the right thing here

alenakhineika: Left TODO comment in code and added the same question to the `Technical Alignment Sync` backlog)

src/language/mongoDBService.ts
alenakhineika: And btw this is not a main thread, this is a LS separate thread already. I just tried a new DataService and it does not throw the error anymore there, we could replace CliServiceProvider with DataService in the mongoDBService.ts, but since it is a separate thread we can not just pass DataService, we need to pass the connection options and do a separate connect there. And I am thinking if we really want to pass two sets of connection options in a bit different format, one for CliServiceProvider and one for DataService 🤔

alenakhineika: Or we actually could pass only those that are meant for DataService and then do getMongoClientConnectionOptions.

addaleax: No, it’s not if it’s all on the same thread anyway :)

PR #493, opened 2023-03-15
feat: add diagnostics feature to the language server VSCODE-375

alenakhineika:
## Description

Add the VSCode Language Server [Diagnostics](https://code.visualstudio.com/api/language-extensions/programmatic-language-features#provide-diagnostics) feature to highlight the usage of commands that only works inside interactive sessions. We've seen folks try to use Shell syntax that doesn't work in playgrounds like `show dbs;` or instead of using the correct `use('database_name')` people type `use database_name` which only works in the Shell and not playgrounds.

The diagnostic not only highlights the incorrect syntax but also offers quick fixes for the current problem and for all similar problems in the file.

Currently, the following fixes are available (the complete list is [here](https://www.mongodb.com/docs/mongodb-shell/reference/access-mdb-shell-help/)):

```js
[
  { issue: 'use', fix: "use('database')" },
  { issue: 'show databases', fix: 'db.getMongo().getDBs()' },
  { issue: 'show dbs', fix: 'db.getMongo().getDBs()' },
  { issue: 'show collections', fix: 'db.getCollectionNames()' },
  { issue: 'show tables', fix: 'db.getCollectionNames()' },
  { issue: 'show profile', fix: "db.getCollection('system.profile').find()" },
  { issue: 'show users', fix: 'db.getUsers()' },
  { issue: 'show roles', fix: 'db.getRoles({ showBuiltinRoles: true })' },
  { issue: 'show logs', fix: "db.adminCommand({ getLog: '*' })" },
  { issue: 'show log', fix: "db.adminCommand({ getLog: 'global' })" },
]
```

https://user-images.githubusercontent.com/16307679/225434031-9a293c9e-e3d8-4946-92ae-ee5ac9510d9e.mov

---

Additionally, I tried _again_ to fix the telemetry flaky test.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/language/mongoDBService.ts
+++ b/src/language/mongoDBService.ts
@@ -118,8 +118,10 @@ export default class MongoDBService {
     }
 
     try {
-      // Get and cache log names for diagnostic fixes.
-      this._cachedLogs = await this._getLogs();
+      // Get log names for the admin database.
+      const logNames = await this._getLogs();
+      // Cache log names for diagnostic fixes.
+      this._cacheLogNames(logNames);
     } catch (error) {
       this._connection.console.error(
         `LS get logs error: ${util.inspect(error)}`
@@ -628,10 +630,10 @@ export default class MongoDBService {
                 endCharacter += item.label.length;
               } else if (line.indexOf(`use '${item.label}'`) >= 0) {
                 databaseName = item.label;
-                endCharacter += item.label.length;
+                endCharacter += item.label.length + 2;
               } else if (line.indexOf(`use "${item.label}"`) >= 0) {
                 databaseName = item.label;
-                endCharacter += item.label.length;
+                endCharacter += item.label.length + 2;
               }
             });
             fix = `use('${databaseName}')`;
@@ -646,10 +648,10 @@ export default class MongoDBService {
                 endCharacter += name.length;
               } else if (line.indexOf(`show log '${name}'`) >= 0) {
                 logName = name;
-                endCharacter += name.length;
+                endCharacter += name.length + 2;
               } else if (line.indexOf(`show log "${name}"`) >= 0) {
                 logName = name;
-                endCharacter += name.length;
+                endCharacter += name.length + 2;
               }
             });
             fix = `db.adminCommand({ getLog: '${logName}' })`;
@@ -686,6 +688,13 @@ export default class MongoDBService {
 
       this._cachedFields[namespace] = fields ? fields : [];
     }
+  }
+
+  /**
+   * Cache log names for diagnostic fixes.
+   */
+  _cacheLogNames(logs: string[]): void {
+    this._cachedLogs = logs;
   }
 
   /**
--- a/src/language/mongoDBService.ts
+++ b/src/language/mongoDBService.ts
@@ -561,6 +561,7 @@ export default class MongoDBService {
   }
 
   // Highlight the usage of commands that only works inside interactive session.
+  // eslint-disable-next-line complexity
   provideDiagnostics(textFromEditor: string) {
     const lines = textFromEditor.split(/\r?\n/g);
     const diagnostics: Diagnostic[] = [];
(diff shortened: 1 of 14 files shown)


src/language/mongoDBService.ts
```diff
@@ -101,15 +109,23 @@ export default class MongoDBService {
...
+
+    try {
+      // Get log names for the admin database.
+      const logNames = await this._getLogs();
+      // Cache log names for diagnostic fixes.
+      this._cacheLogNames(logNames);
+    } catch (error) {
+      this._connection.console.error(
+        `LS get logs error: ${util.inspect(error)}`
+      );
```
addaleax, line 128: Fwiw, I don’t really expect people to try to run `show log ...` in the VSCode extension code, I don’t think we need to put in a lot of effort to support it. I know the code’s here already written, so feel free to just resolve this, but yeah, for 99.9 % of users the slowdown from the extra `getLog` command that’s being run for every connection here will not be useful to them.

src/language/mongoDBService.ts
```diff
@@ -557,6 +595,86 @@ export default class MongoDBService {
...
+              if (line.indexOf(`use ${item.label}`) >= 0) {
+                databaseName = item.label;
+                endCharacter += item.label.length;
+              } else if (line.indexOf(`use '${item.label}'`) >= 0) {
+                databaseName = item.label;
+                endCharacter += item.label.length + 2;
+              } else if (line.indexOf(`use "${item.label}"`) >= 0) {
+                databaseName = item.label;
+                endCharacter += item.label.length + 2;
+              }
```
addaleax, line 637: This only accounts for databases that exist, right? It’s also valid to create a new database by doing `use newdatabase` and then (explicitly or implicitly) creating a collection there

src/language/mongoDBService.ts
```diff
@@ -557,6 +595,86 @@ export default class MongoDBService {
...
+      { issue: 'show logs', fix: "db.adminCommand({ getLog: '*' })" },
+      { issue: 'show log ', fix: "db.adminCommand({ getLog: 'global' })" },
+    ];
+
+    lines.forEach((line, i) => {
+      invalidInteractiveSyntaxes.forEach(({ issue, fix }) => {
+        const startCharacter = line.indexOf(issue);
+        let endCharacter = startCharacter + issue.length;
+
+        if (startCharacter >= 0) {
```
addaleax, line 623: Since this is a long conditional that wraps the rest of the function here, maybe it’s easier to do the inverse?

```suggestion
        if (startCharacter === -1) return;
```

(Also, totally random side note, but this PR adds both `for ... of` loops and `.forEach()` loops; might be nicer to use `for ... of` loops consistently? For the `lines.forEach((line, i) => {` one, you can use e.g. `for (const [ i, line ] of lines.entries()) {`)

src/language/mongoDBService.ts
```diff
@@ -557,6 +595,86 @@ export default class MongoDBService {
...
+          if (issue === 'use ') {
+            let databaseName = 'database';
+
+            this._cachedDatabases.forEach((item: CompletionItem) => {
+              if (line.indexOf(`use ${item.label}`) >= 0) {
```
addaleax, line 628: I’m wondering if it makes more sense to validate these items more strictly. For example, right now, I think this would also catch cases such as `print("I want a new mouse in my house")`, right?

I think it probably makes sense to at least check that there’s nothing but whitespace before the `use` keyword, and it probably also makes sense to ensure that the check here returns the same `use` keywords as the one above, e.g.

```suggestion
              if (line.substring(startCharacter).startsWith(`use ${item.label}`)) {
```

src/language/mongoDBService.ts
alenakhineika: I kept log, but considering the fact that dbs also can have new names, I've rewritten the whole block, so I parse the string to find a database name and log type, or whatever can come later that has a replacement value and use it when a build a fix string.

src/language/mongoDBService.ts
```diff
@@ -557,6 +560,87 @@ export default class MongoDBService {
...
+        fix: "db.adminCommand({ getLog: 'VALUE_TO_REPLACE' })",
+        default: 'global',
+      },
+    ];
+
+    for (const [i, line] of lines.entries()) {
+      for (const item of invalidInteractiveSyntaxes) {
+        const issue = item.issue;
+        const startCharacter = line.indexOf(issue);
+        if (!line.trim().startsWith(issue)) {
```
addaleax, line 591: This would catch something like `user.authenticate();`, right?

src/language/mongoDBService.ts
```diff
@@ -557,6 +560,87 @@ export default class MongoDBService {
...
+        // instead of VALUE_TO_REPLACE placeholder of the `fix` string.
+        if (valueToReplaceWith) {
+          const words = line.substring(startCharacter).split(' ');
+
+          // The index of the value for `use <value>` and `show log <value>`.
+          const valueIndex = issue.split(' ').length;
+
+          if (words[valueIndex]) {
+            // The `use ('database')` is a valid JS.
+            if (issue === 'use' && words[valueIndex].startsWith('(')) {
```
addaleax, line 613: `show ('logs')` is also valid JS, not just `use(...)`

alenakhineika: But we don't have just `show` in rules, so it won't be validated. But I see your point, we can safely remove this `issue === 'use'` validation and it won't break anything and may catch  other similar issues if we add any.

addaleax (approved):

PR #495, opened 2023-03-23
chore(lint): remove require imports

Anemy:
Standardizing imports and removing the ignore on the `@typescript-eslint/no-var-requires` rule.


alenakhineika (approved):
🔥

src/views/webview-app/components/app.tsx
```diff
@@ -15,7 +15,8 @@ import {
   MESSAGE_TYPES,
 } from '../extension-app-message-constants';
 
-const styles = require('../connect.module.less');
+// TODO:
```
alenakhineika, line 18: We don't need it here anymore probably?

Anemy, line 18:
```suggestion
```

PR #504, opened 2023-03-30
fix: save documents from tree view and do not reopen them for playground runs VSCODE-399, VSCODE-400

alenakhineika:
## Description

Bug fixes for the release:
- [VSCODE-399](https://jira.mongodb.org/browse/VSCODE-399) Can not save a mongodb document from a tree view
- [VSCODE-400](https://jira.mongodb.org/browse/VSCODE-400) Run playground sets focus to a previously open document

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/package.json
+++ b/package.json
@@ -619,7 +619,7 @@
         },
         {
           "command": "mdb.saveMongoDBDocument",
-          "when": "editorLangId == json"
+          "when": "resourceScheme == VIEW_DOCUMENT_SCHEME"
         },
         {
           "command": "mdb.runSelectedPlaygroundBlocks",
@@ -812,7 +812,7 @@
         "command": "mdb.saveMongoDBDocument",
         "key": "ctrl+s",
         "mac": "cmd+s",
-        "when": "editorLangId == json"
+        "when": "resourceScheme == VIEW_DOCUMENT_SCHEME"
       }
     ],
     "capabilities": {
--- a/package.json
+++ b/package.json
@@ -812,7 +812,7 @@
         "command": "mdb.saveMongoDBDocument",
         "key": "ctrl+s",
         "mac": "cmd+s",
-        "when": "mdb.isPlayground == true"
+        "when": "editorLangId == json"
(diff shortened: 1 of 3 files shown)


src/editors/playgroundController.ts
```diff
@@ -581,8 +581,6 @@ export default class PlaygroundController {
 
     this._playgroundResult = evaluateResponse.result;
 
-    this._explorerController.refresh();
```
alenakhineika, line 584: This triggers the click event on the selected in the tree view document. We do not refresh lists on the fly on other explorers, so let's also use the existing refresh button in this case.

package.json
```diff
@@ -812,7 +812,7 @@
         "command": "mdb.saveMongoDBDocument",
         "key": "ctrl+s",
         "mac": "cmd+s",
-        "when": "mdb.isPlayground == true"
+        "when": "editorLangId == json"
```
Anemy, line 815: Will this run on all json files? Or is this hot key only used when we have the extension open?

Anemy: Maybe not for this pr, unless you want to, would it make sense to not handle every save event ourselves there?
Instead we could update this `when` clause to check for `resourceScheme` which indicates a mongodb file. 
https://code.visualstudio.com/api/references/when-clause-contexts
```suggestion
        "when": "resourceScheme == VIEW_DOCUMENT_SCHEME"
```
Scheme from https://github.com/mongodb-js/vscode/blob/main/src/editors/mongoDBDocumentService.ts#L20

It's a bit obscure, but I'm thinking if another extension is used for formatting json or doing something when json is saved this current approach we have would conflict with that. There's one other place where we use the `editorLangId == json` `when` clause which we could update as well.

Anemy (approved):

Anemy (approved):

PR #509, opened 2023-04-05
feat: extend typescript help signatures with mongodb items VSCODE-403 

alenakhineika:
## Description

The JS contextual signature help for `find` thinks that any `find` keyword is the `Array.find()` method. This is confusing for users of the extension, because MongoDB has its own usage and syntax the [find](https://www.mongodb.com/docs/manual/reference/method/db.collection.find/) method.

VSCode does not allow extensions to change the results of other extensions, therefor to overcome this obstacle we [use the TypeScript Language Service API](https://github.com/microsoft/TypeScript/wiki/Using-the-Language-Service-API) to extend default JavaScript [Method Signatures](https://code.visualstudio.com/api/language-extensions/programmatic-language-features#help-with-function-and-method-signatures) in playgrounds.

If the cursor is positioned in the query node we show the query method help:

```text
Collection.find(query, projection, options) : Cursor
---
Selects documents in a collection or view.
```

And similarly for the [aggregate](https://www.mongodb.com/docs/manual/reference/method/db.collection.aggregate/) method, we show the aggregate method help:

```text
Collection.aggregate(pipeline, options) : Cursor
---
Calculates aggregate values for the data in a collection or a view.
```

If the current node is neither `find` nor `aggregation`, we return the TypeScript language server help signatures.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/language/server.ts
+++ b/src/language/server.ts
@@ -9,10 +9,13 @@ import {
   RequestType,
   TextDocumentSyncKind,
   Connection,
+  SignatureHelpParams,
 } from 'vscode-languageserver/node';
 import { TextDocument } from 'vscode-languageserver-textdocument';
 
 import MongoDBService from './mongoDBService';
+import TypeScriptService from './tsLanguageService';
+
 import { ServerCommands } from './serverCommands';
 import {
   PlaygroundEvaluateParams,
@@ -29,6 +32,9 @@ const documents: TextDocuments<TextDocument> = new TextDocuments(TextDocument);
 
 // MongoDB Playground Service Manager.
 const mongoDBService = new MongoDBService(connection);
+
+// TypeScript language service.
+const typeScriptService = new TypeScriptService(connection);
 
 let hasConfigurationCapability = false;
 // let hasWorkspaceFolderCapability = false;
@@ -66,6 +72,10 @@ connection.onInitialize((params: InitializeParams) => {
         resolveProvider: true,
         triggerCharacters: ['.'],
       },
+      signatureHelpProvider: {
+        resolveProvider: true,
+        triggerCharacters: [ ',' ]
+      }
       // documentFormattingProvider: true,
       // documentRangeFormattingProvider: true,
       // codeLensProvider: {
@@ -221,6 +231,16 @@ connection.onCompletionResolve((item: CompletionItem): CompletionItem => {
   // }
 
   return item;
+});
+
+connection.onSignatureHelp((params: SignatureHelpParams) => {
+  const document = documents.get(params.textDocument.uri);
+
+  if (!document) {
+    return;
+  }
+
+  return typeScriptService.doSignatureHelp(document, params.position);
 });
 
 connection.onRequest('textDocument/rangeFormatting', (event) => {
--- a/src/language/server.ts
+++ b/src/language/server.ts
@@ -74,8 +74,8 @@ connection.onInitialize((params: InitializeParams) => {
       },
       signatureHelpProvider: {
         resolveProvider: true,
-        triggerCharacters: [ ',' ]
-      }
+        triggerCharacters: [','],
+      },
       // documentFormattingProvider: true,
       // documentRangeFormattingProvider: true,
       // codeLensProvider: {
@@ -240,7 +240,21 @@ connection.onSignatureHelp((params: SignatureHelpParams) => {
     return;
   }
 
-  return typeScriptService.doSignatureHelp(document, params.position);
+  const textFromEditor = document.getText() || '';
+  const mongodbSignatures = mongoDBService.provideSignatureHelp(
+    textFromEditor,
+    params.position,
+    params.context
+  );
+
+  if (mongodbSignatures) {
+    return mongodbSignatures;
+  }
+
+  return typeScriptService.doSignatureHelp(
+    document,
+    params.position
+  );
 });
 
 connection.onRequest('textDocument/rangeFormatting', (event) => {
--- a/src/language/tsLanguageService.ts
+++ b/src/language/tsLanguageService.ts
@@ -1,6 +1,10 @@
 import { Connection } from 'vscode-languageserver/node';
-import type { SignatureHelp, SignatureInformation, ParameterInformation } from 'vscode-languageserver/node';
-import * as ts from 'typescript';
+import type {
+  SignatureHelp,
+  SignatureInformation,
+  ParameterInformation,
+} from 'vscode-languageserver/node';
+import ts from 'typescript';
 import { TextDocument, Position } from 'vscode-languageserver-textdocument';
 import { readFileSync } from 'fs';
 import { join, basename, dirname } from 'path';
@@ -12,9 +16,13 @@ type TypeScriptServiceHost = {
 };
 
 // Server folder.
-const serverPath = basename(__dirname) === 'dist' ? dirname(__dirname) : dirname(dirname(__dirname));
+const serverPath =
+  basename(__dirname) === 'dist'
+    ? dirname(__dirname)
+    : dirname(dirname(__dirname));
+
 // TypeScript library folder.
-const librarPath = join(serverPath, '../node_modules/typescript/lib');
+const librarPath = join(serverPath, 'node_modules/typescript/lib');
 const contents: { [name: string]: string } = {};
 
 export default class TypeScriptService {
@@ -47,10 +55,10 @@ export default class TypeScriptService {
     const compilerOptions = {
       allowNonTsExtensions: true,
       allowJs: true,
-      lib: ['lib.es2020.full.d.ts'],
+      lib: ['lib.es2020.full.d.ts'], // Should match to lib from tsconfig.json.
       target: ts.ScriptTarget.Latest,
       moduleResolution: ts.ModuleResolutionKind.Classic,
-      experimentalDecorators: false
+      experimentalDecorators: false,
     };
     let currentTextDocument = TextDocument.create('init', 'javascript', 1, '');
 
@@ -77,11 +85,11 @@ export default class TypeScriptService {
         return {
           getText: (start, end) => text.substring(start, end),
           getLength: () => text.length,
-          getChangeRange: () => undefined
+          getChangeRange: () => undefined,
         };
       },
       getCurrentDirectory: () => '',
-      getDefaultLibFileName: () => 'es2020.full',
+      getDefaultLibFileName: () => 'es2020.full', // Should match to lib from tsconfig.json.
       readFile: (path: string): string | undefined => {
         if (path === currentTextDocument.uri) {
           return currentTextDocument.getText();
@@ -100,7 +108,7 @@ export default class TypeScriptService {
           return false;
         }
         return true;
-      }
+      },
     };
 
     // Create the language service files.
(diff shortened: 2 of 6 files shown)


src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,184 @@
...
+          text = this._loadLibrary(fileName);
+        }
+        return {
+          getText: (start, end) => text.substring(start, end),
+          getLength: () => text.length,
+          getChangeRange: () => undefined,
+        };
+      },
+      getCurrentDirectory: () => '',
+      getDefaultLibFileName: () => 'es2020.full', // Should match to lib from tsconfig.json.
```
addaleax, line 92: This should probably always match the filename in line 58 – does it make sense to extract this to a constant that’s used in both places?

src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,184 @@
...
+      contents[name] = content;
+    }
+    return content;
+  }
+
+  _getTypeScriptServiceHost() {
+    const compilerOptions = {
+      allowNonTsExtensions: true,
+      allowJs: true,
+      lib: ['lib.es2020.full.d.ts'], // Should match to lib from tsconfig.json.
```
addaleax, line 58: Which `tsconfig.json` are we referring to here? This is about playgrounds (… right), so there wouldn’t normally be a `tsconfig.json` for those, correct?

src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,184 @@
...
+        return {
+          getText: (start, end) => text.substring(start, end),
+          getLength: () => text.length,
+          getChangeRange: () => undefined,
+        };
+      },
+      getCurrentDirectory: () => '',
+      getDefaultLibFileName: () => 'es2020.full', // Should match to lib from tsconfig.json.
+      readFile: (path: string): string | undefined => {
+        if (path === currentTextDocument.uri) {
```
addaleax, line 94: For my own understanding – are `path` and `currentTextDocument.uri` filesystem paths or are they URIs? I’m guessing it’s the former, it’s just confusing that a property named `.uri` would not be a URI…

src/language/visitor.ts
```diff
@@ -201,13 +243,35 @@ export class Visitor {
...
   }
 
+  _getDefaultsForSignatureHelp() {
+    return {
+      isFind: false,
+      isAggregation: false,
+    };
+  }
+
+  _getDefaultsForExportToLanguagep() {
```
addaleax, line 253: Might be a typo (`p` at the end)?

src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,184 @@
...
+          return true;
+        }
+        return !!this._loadLibrary(path);
+      },
+      directoryExists: (path: string): boolean => {
+        // Typescript tries to first find libraries in node_modules/@types and node_modules/@typescript.
+        if (path.startsWith('node_modules')) {
+          return false;
+        }
+        return true;
```
addaleax, line 110: Should this return `true` unconditionally?

This is very interesting though! We might be able to avoid having to implement all the stuff we are adding here in `visitor.ts` if we just manage to provide Typescript with an extra definition file, e.g. something like

```js
/// <reference types="es2020.full" />
declare interface Collection {
  find(query: Record<string, unknown>, project: Record<string, unknown>, options: Record<String, unknown>): any;
  aggregate(pipeline: Record<string, unknown>[], options: Record<String, unknown>): any;
}

declare global {
  let db: {
    [key: string]: Collection;
  };

  let use: (dbName: string) => void;
}
```

this would hopefully make it a lot easier to add definitions for other shell commands as well, so that we don’t have to inspect the AST manually every time

src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,184 @@
...
+};
+
+// Server folder.
+const serverPath =
+  basename(__dirname) === 'dist'
+    ? dirname(__dirname)
+    : dirname(dirname(__dirname));
+
+// TypeScript library folder.
+const librarPath = join(serverPath, 'node_modules/typescript/lib');
```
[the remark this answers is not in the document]
alenakhineika: Yeah..) 👀

src/language/tsLanguageService.ts
alenakhineika: The documentation says:
_URI - The associated URI for this document. Most documents have the file-scheme, indicating that they represent files on disk. However, some documents may have other schemes indicating that they are not available on disk._

If URI starts with `file:` scheme it indicates that this is a filesystem path, and for the `untitled:` scheme, this is URI.

src/language/tsLanguageService.ts
alenakhineika: Hah I just saw this comment and this is exactly what I left as a comment about above when I added mongodb.d.ts to the lib array, nice 👍

src/language/server.ts
```diff
@@ -61,11 +66,16 @@ connection.onInitialize((params: InitializeParams) => {
...
+      // Tell the client that the server supports code completion.
       completionProvider: {
         resolveProvider: true,
         triggerCharacters: ['.'],
       },
+      // Tell the client that the server supports help signatures.
+      signatureHelpProvider: {
+        resolveProvider: true,
+        triggerCharacters: [',', '('],
+      },
```
addaleax, line 78: Just checking, is this a leftover from before the .d.ts change or is it still having an effect?

src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,215 @@
...
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
+type TypeScriptServiceHost = {
+  getLanguageService(jsDocument: TextDocument): ts.LanguageService;
+  getCompilationSettings(): ts.CompilerOptions;
+  dispose(): void;
+};
+
+const contents: { [name: string]: string } = {};
```
addaleax, line 18: Should this be global or should it be a property on the `TypeScriptService` instance?

(also: you’ll probably want to use a Map here or `Object.create(null)` instead of `{}`, otherwise weird things happen if something accesses e.g. `contents['__proto__']`)

src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    if (typeof content !== 'string') {
+      let libPath;
+
+      /* if (name === `lib.${TS_CONFIG_LIBRARY_NAME}.d.ts`) {
+        libPath = join(this._extensionPath, 'node_modules/typescript/lib', name);
+      } else */
+
+      if (name === `${MDB_CONFIG_LIBRARY_NAME}.d.ts`) {
+        libPath = join(this._extensionPath, 'node_modules/mongodb', name);
```
addaleax, line 59: Are we loading the driver’s type definitions here? Out of curiosity – why?

src/language/tsLanguageService.ts
addaleax: Okay, that answers the question for `.uri`, but what about `path`? We pass `path` to `_loadLibrary` below, which doesn’t seem to expect its argument to be an URI (but rather a relative file path)? What am I missing here? :)

src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    // Create the language service files.
+    const jsLanguageService = ts.createLanguageService(host);
+
+    return {
+      // Return a new instance of the language service.
+      getLanguageService(jsDocument: TextDocument): ts.LanguageService {
+        currentTextDocument = jsDocument;
+        return jsLanguageService;
+      },
```
addaleax, line 145: The comment above this function and its name are a bit misleading – the function always returns the same instance, not a new one. Would it make more sense to either a) create a new language service instance each time or b) split this function into two functions, `getLanguageService(): ts.LanguageService` and `setCurrentTextDocument(jsDocument: TextDocument): void`?

src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,215 @@
...
+        content = '';
+      }
+
+      contents[name] = content;
+    }
+
+    return content;
+  }
+
+  _getTypeScriptServiceHost() {
```
addaleax, line 85:
```suggestion
  _getTypeScriptServiceHost(): TypeScriptServiceHost {
```

src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,215 @@
...
+      const ret: SignatureHelp = {
+        activeSignature: signHelp.selectedItemIndex,
+        activeParameter: signHelp.argumentIndex,
+        signatures: [],
+      };
+      signHelp.items
+        .map((item) => {
+          const hasInt8Array = item.prefixDisplayParts.filter(
+            (prefix) => prefix.text !== 'Int8Array'
+          );
```
addaleax, line 182: I think there might be room for a bit of readability improvement here – `hasInt8Array` sounds like it would be a boolean, but it seems like it’s an array of items that are specifically not referring to `Int8Array`, so maybe there’s a better variable name? And finally, I think it’s definitely worth leaving a comment about why `Int8Array` is excluded (and why only it and none of the other typed arrays)

src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,215 @@
...
+            signature.parameters?.push(parameter);
+            if (i < a.length - 1) {
+              signature.label += ts.displayPartsToString(
+                item.separatorDisplayParts
+              );
+            }
+          });
+          signature.label += ts.displayPartsToString(item.suffixDisplayParts);
+          ret.signatures.push(signature);
+        });
```
addaleax, line 210: Maybe this is from editing leftovers, but it’s a bit odd that this code is using `.map()` with the semantics of `.forEach()` and `.forEach()` with the semantics of `.map()` … :slightly_smiling_face: This could probably be a `ret.signatures = signHelp.items.map((item) => { ...; return signature; })` call instead?

src/language/tsLanguageService.ts
```diff
@@ -0,0 +1,215 @@
...
+          return currentTextDocument.getText();
+        }
+        return this._loadLibrary(path);
+      },
+      fileExists: (): boolean => false,
+      directoryExists: (): boolean => false,
+    };
+
+    // Create the language service files.
+    const jsLanguageService = ts.createLanguageService(host);
```
addaleax, line 138: [Question only:] Correct me if I’m wrong, but the general structure here is that `host` serves as our way to connect the TS logic to our VSCode documents, but other than that, the `LanguageService` instance is unaware of the fact that it’s running against a VSCode environment/as part of a VSCode extension. Is that right? Could we potentially use this, say, inside of mongosh to provide autocompletions there?

src/types/global.d.ts
```diff
@@ -0,0 +1,5 @@
+export {};
+
+declare global {
+  let use: (dbName: string) => void;
+}
```
addaleax, line 5: I’m guessing the next step here is to figure out how to create a .d.ts file for mongosh environments? Probably not something for this PR, though :slightly_smiling_face:

src/language/server.ts
alenakhineika: Actually, we still need it, otherwise the `connection.onSignatureHelp` won't be called and help signatures won't be returned from the ts service.

src/language/tsLanguageService.ts
alenakhineika: Will return just undefined here, does not affect anything.

PR #536, opened 2023-06-14
electron external

lerouxb:
https://mongodb.slack.com/archives/GJ2CBBY86/p1686239462406709?thread_ts=1686237929.639499&cid=GJ2CBBY86

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -51,6 +51,7 @@ module.exports = (env, argv) => {
 
       // Electron:
       '@electron/remote': '@electron/remote',
+      electron: 'commonjs2 electron',
 
       // MongoDB node driver externals:
       snappy: 'snappy',


webpack.config.js
```diff
@@ -51,6 +51,7 @@ module.exports = (env, argv) => {
 
       // Electron:
       '@electron/remote': '@electron/remote',
+      electron: 'commonjs2 electron',
```
mcasimir, line 54: We already remove `electron` from the bundle with `alias: { electron: false }`

lerouxb: Then I don't think we understand that mongosh failure 😞

lerouxb: I filed a ticket. Will close this https://jira.mongodb.org/browse/VSCODE-427

PR #543, opened 2023-06-15
test: run nightly tests against latest vscode VSCODE-340

alenakhineika:
## Description

Run test suites on the latest stable release.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/test/runTest.ts
+++ b/src/test/runTest.ts
@@ -17,6 +17,7 @@ async function main(): Promise<any> {
 
     // Download VS Code, unzip it and run the integration test
     await runTests({
+      version: 'stable',
       extensionDevelopmentPath,
       extensionTestsPath,
       launchArgs: [testWorkspace, '--disable-extensions'],
--- a/src/test/runTest.ts
+++ b/src/test/runTest.ts
@@ -17,9 +17,7 @@ async function main(): Promise<any> {
 
     // Download VS Code, unzip it and run the integration test
     await runTests({
-      // Download latest stable release.
-      // Might be different from the vscode engine specified in package.json
-      version: 'stable',
+      version: 'insiders', // Download latest insiders.
       extensionDevelopmentPath,
       extensionTestsPath,
       launchArgs: [testWorkspace, '--disable-extensions'],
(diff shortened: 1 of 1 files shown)


src/test/runTest.ts
```diff
@@ -17,6 +17,9 @@ async function main(): Promise<any> {
 
     // Download VS Code, unzip it and run the integration test
     await runTests({
+      // Download latest stable release.
+      // Might be different from the vscode engine specified in package.json
+      version: 'stable',
```
Anemy, line 22: This would make all tests run against the `stable` version of VSCode. Should we instead only have this on the `nightly` tests? The ones that are run on the schedule: https://github.com/mongodb-js/vscode/blob/975bfde1f0c9d1335ea2fc95308a00c950db297d/.github/workflows/test-and-build.yaml#LL11C7-L11C11

alenakhineika: Aren't those also using runTest.ts? I thought this is the same test setup that runs `npm run test` that goes through runTest.ts.

alenakhineika: Or do you want only nightly tests to run again stable and others against the version in package.json? I think we do not always remember to update the engine in package.sjon so even for a local run this is fine to run against stable.

Anemy: I think it would be good for us to run it against `insiders` if the goal is to catch things early. 
https://github.com/microsoft/vscode-test/blob/0ec222ef170e102244569064a12898fb203e5bb7/lib/download.ts#L170
I think running on stable vscode, while it is good to do, won't help us catch things until it's too late.

alenakhineika: I thought about this and don't mind using insiders if you prefer this, but even with stable it will help us to catch more bugs compared to what we have now, forgetting to update package.json :)

Anemy: Trueee that 👍

Anemy (approved):
If we find `insiders` to be unreliable for ci testing then we can move back to `stable`.

PR #550, opened 2023-06-29
chore: added migration step to migrate keytar secrets to vscode SecretStorage - VSCODE-435

himanshusinghs:
…t storage

## Description
In this PR, we added a one-off migration step to migrate secrets stored using keytar to VsCode's SecretStorage. The idea is to run this migration only once when the extension activates and loads the connections, unless there was some error other than keytar unavailable.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -26,7 +26,9 @@ import {
 } from './storage/storageController';
 import { StorageController, StorageVariables } from './storage';
 import { StatusView } from './views';
-import TelemetryService from './telemetry/telemetryService';
+import TelemetryService, {
+  TelemetryEventTypes,
+} from './telemetry/telemetryService';
 import LINKS from './utils/links';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const packageJSON = require('../package.json');
@@ -76,6 +78,35 @@ interface ConnectionSecretsInfo {
 
 type StoreConnectionInfoWithConnectionOptions = StoreConnectionInfo &
   Required<Pick<StoreConnectionInfo, 'connectionOptions'>>;
+
+export class KeytarModuleUnavailableError extends Error {
+  name = 'KeytarModuleUnavailableError';
+  message = 'Keytar module not found';
+}
+
+export class ConnectionMigrationFailedError extends Error {
+  name = 'ConnectionMigrationFailedError';
+  failedConnectionIdsAndNames: {
+    connectionId: string;
+    connectionName: string;
+  }[];
+  constructor(
+    failedConnectionIdsAndNames: {
+      connectionId: string;
+      connectionName: string;
+    }[]
+  ) {
+    super(
+      [
+        'Could not migration the following connections:',
+        failedConnectionIdsAndNames
+          .map(({ connectionName }) => connectionName)
+          .join(', '),
+      ].join(' ')
+    );
+    this.failedConnectionIdsAndNames = failedConnectionIdsAndNames;
+  }
+}
 
 export default class ConnectionController {
   // This is a map of connection ids to their configurations.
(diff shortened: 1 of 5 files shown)


src/connectionController.ts
```diff
@@ -77,6 +79,35 @@ interface ConnectionSecretsInfo {
 type StoreConnectionInfoWithConnectionOptions = StoreConnectionInfo &
   Required<Pick<StoreConnectionInfo, 'connectionOptions'>>;
 
+export class KeytarModuleUnavailableError extends Error {
+  name = 'KeytarModuleUnavailableError';
```
alenakhineika, line 83: Do you know if we need to set a name explicitly here? I think `err.constructor.name` should work based on the class name. And maybe I missing it somewhere, but seem you don't read it anywhere and only check `instanceof KeytarModuleUnavailableError`.

himanshusinghs: yea that's right. I thought I would use it but did not. Will remove it, thank you :D

addaleax: Fwiw, not strong feelings either way I think it’s a good practice to set a `name` property for `Error` subclasses and use that for instance checking rather than `instanceof` or `err.constructor.name` (because it works across contexts and `err.name` is [already standard built-in JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name))

src/connectionController.ts
```diff
@@ -204,19 +225,233 @@ export default class ConnectionController {
     }
   }
 
+  _notifyKeytarMigrationError(migrationError: Error) {
+    const notificationAction = {
+      title: "Don't show again",
```
alenakhineika, line 230: How do we know that it won't be shown again? The closing information message just dismisses it. Users often ask for a feature to disable some notifications and this message can be misleading that we start to support closing messages forever.

src/connectionController.ts
```diff
@@ -204,19 +225,233 @@ export default class ConnectionController {
...
+      migrationError instanceof KeytarModuleUnavailableError &&
+      !keytarUnavailableErrAlreadyShown
+    ) {
+      void vscode.window
+        .showInformationMessage(
+          `Failed to update connection storage - ${migrationError.message}`,
+          notificationAction
+        )
+        .then((action) => {
+          if (action === notificationAction) {
```
alenakhineika, line 246: What is the purpose here to check the type of action? Then handles only the response of showInformationMessage it is attached to.

src/test/suite/connectionController.test.ts
```diff
@@ -2,13 +2,16 @@ import sinon from 'sinon';
 import type { SinonStub } from 'sinon';
 import util from 'util';
 import * as vscode from 'vscode';
-import { afterEach, beforeEach } from 'mocha';
+import { afterEach, beforeEach, describe } from 'mocha';
```
alenakhineika, line 5: We use `suite` instead of `describe`. You don't need to import `suite` to use it.

src/connectionController.ts
himanshusinghs: yea that's right. We probably are fine without that check as well, given that we only have one action. In the original write up I had two actions here and I just did not remove this check after I got rid of the other action.

src/connectionController.ts
alenakhineika: But why it will appear again and again? Shouldn't the migration happen only once?

himanshusinghs: There is one (more like theoretical) case which is if the migration failed for an error which is not `KeytarModuleUnavailableError`. I referred that [here](https://github.com/mongodb-js/vscode/pull/550/files#diff-c2dbb92c81a472b4bc81088731816dd9f4676cdb1b4a8ade3fdb49c6a3b20d82R411-R421)

alenakhineika: I have a feeling that this is the kind of project that needs a design doc.

src/connectionController.ts
```diff
@@ -204,19 +225,233 @@ export default class ConnectionController {
...
+          connection.connectionId,
+          secretsFromKeyTar
+        );
+
+        log.debug('_migrateKeytarSecrets: removing secrets from keytar');
+        await ext.keytarModule.deletePassword(
+          this._serviceName,
+          connection.connectionId
+        );
+      } catch (error) {
```
addaleax, line 380: If there is an error here, should it end up being logged somehow?

Issue #476, opened 2023-02-10
Bug Report: Connections not storing on linux with credential store


M-Gonzalo:
#### Current Behavior

Connections are not saved, either locally or globally. A reload or restart of the ID deletes everything

#### Environment

- node.js / npm versions:
	- Linux 6.1.8-arch1-1 
	- Plasma 5.26.5 
	- node v16.18.1

Anemy:
@M-Gonzalo It might be that VSCode is not able to access the credential storage on your system.

To fix this I think you can try to make the credential storage usable by VSCode/the extension. To make the credential storage work without erroring, installing [gnome-keyring](http://manpages.ubuntu.com/manpages/cosmic/man1/gnome-keyring-daemon.1.html#:~:text=The%20gnome%2Dkeyring%2Ddaemon%20is,tool%20to%20interact%20with%20it.) might fix the issue.

Closing this, feel free to reopen if you do have a credential storage on your system that the extension has access to.

M-Gonzalo:
> @M-Gonzalo It might be that VSCode is not able to access the credential storage on your system.
> 
> To fix this I think you can try to make the credential storage usable by VSCode/the extension. To make the credential storage work without erroring, installing [gnome-keyring](http://manpages.ubuntu.com/manpages/cosmic/man1/gnome-keyring-daemon.1.html#:~:text=The%20gnome%2Dkeyring%2Ddaemon%20is,tool%20to%20interact%20with%20it.) might fix the issue.
> 
> Closing this, feel free to reopen if you do have a credential storage on your system that the extension has access to.

I'm on KDE, it has its own credential storage system. I can use any other extensions that might need to store something, including some with API keys or those that need to login somewhere first

alenakhineika:
Hey @M-Gonzalo, we release a new 1.1.0 version of the extension, that includes migration secrets from keytar to SecretStorage. See the [release notes](https://github.com/mongodb-js/vscode/releases/tag/v1.1.0) for more details. Please try the new version and let us know if this solves the issue of storing connections for you.

M-Gonzalo:
> 

Thanks. I found out that the problem only happens when I configure the extension to ask where to store the credentials. Not if I leave the default settings. That's how I've been managing for now. Might help to reproduce the issue somewhere else

alenakhineika:
[...]

PR #552, opened 2023-07-04
chore: added migration step to migrate keytar secrets to vscode SecretStorage - VSCODE-435

himanshusinghs:
## Description
In this PR, we added a migration step to migrate secrets one by one, earlier stored using keytar to VsCode's SecretStorage. The pseudo-code followed here is the following:

```
## Migration process
- Extension activates
- Initiate a call to fetch saved connections
  - Get all the connections from our connection storage
  - For each connection
    - perform the old connection modal to new connection info migration
    - return connection info merged with secrets
      - Check if we have `secretStorageLocation` key set in root of connection info object
      - Yes? (We attempted a migration already for this connection)
        - Is `secretStorageLocation` === `vscode.Keytar`? (We tried migrating but failed, keytar was unavailable)
          - Return the connection info without merging the secrets
        - Is `secretStorageLocation` === `vscode.SecretStorage`?
          - Fetch the secret from `vscode.SecretStorage`
          - Merge it with connection info and return the merged connection info
      - No? (The connection is yet to be migrated)
        - Check if keytar is available
        - No?
          - Set `secretStorageLocation` key to `vscode.Keytar`
          - Log as migration error that keytar is unavailable
          - Return the connection info as it is
        - Yes?
          - Check if credentials are in keytar
          - Yes?
            - Copy the credentials over to `vscode.SecretStorage`
            - Set `secretStorageLocation` key on connection info to `vscode.SecretStorage`
            - Return connection info merged with retrieved secrets
          - No?
            - Set `secretStorageLocation` key to `vscode.SecretStorageLocation`
            - Save '{}' to vscode.SecretStorageLocation for this connection
            - Return connection info merged with empty object for secret
  - Check if there were any connections that we could not migrate and notify the user of them and also track the telemetry event
```

If we fail to migrate for any connection, we notify the user using vscode's information message API. An example here:

image (2)

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Dependents

Caused by: https://github.com/mongodb-js/vscode/issues/546
Resolves: https://github.com/mongodb-js/vscode/issues/476

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -166,7 +166,7 @@ export default class ConnectionController {
 
     if (connectionsWithSecretsInKeytar.length) {
       log.error(
-        'Could not migrate secrets for a few connections',
+        `Could not migrate secrets for ${connectionsWithSecretsInKeytar.length} connections`,
         connectionsWithSecretsInKeytar
       );
       this._telemetryService.track(
@@ -178,7 +178,7 @@ export default class ConnectionController {
       );
       void vscode.window.showInformationMessage(
         [
-          'Could not migrate secrets for a few connections. Please review the following connections:',
+          `Could not migrate secrets for ${connectionsWithSecretsInKeytar.length} connections. Please review the following connections:`,
           connectionsWithSecretsInKeytar
             .map(({ connectionName }) => connectionName)
             .join(', '),
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -4,7 +4,6 @@ import {
   ConnectionInfo,
   ConnectionOptions,
   getConnectionTitle,
-  ConnectionSecrets,
   extractSecrets,
   mergeSecrets,
   connect,
@@ -22,11 +21,13 @@ import formatError from './utils/formatError';
 import LegacyConnectionModel from './views/webview-app/connection-model/legacy-connection-model';
 import {
   StorageLocation,
-  ConnectionsFromStorage,
+  SecretStorageLocation,
 } from './storage/storageController';
 import { StorageController, StorageVariables } from './storage';
 import { StatusView } from './views';
-import TelemetryService from './telemetry/telemetryService';
+import TelemetryService, {
+  TelemetryEventTypes,
+} from './telemetry/telemetryService';
 import LINKS from './utils/links';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const packageJSON = require('../package.json');
@@ -50,6 +51,7 @@ export interface StoreConnectionInfo {
   id: string; // Connection model id or a new uuid.
   name: string; // Possibly user given name, not unique.
   storageLocation: StorageLocation;
+  secretStorageLocation?: SecretStorageLocation;
   connectionOptions?: ConnectionOptions;
   connectionModel?: LegacyConnectionModel;
 }
@@ -69,20 +71,24 @@ interface ConnectionQuickPicks {
   data: { type: NewConnectionType; connectionId?: string };
 }
 
-interface ConnectionSecretsInfo {
-  connectionId: string;
-  secrets: ConnectionSecrets;
-}
+type StoreConnectionInfoWithConnectionModel = StoreConnectionInfo &
+  Required<Pick<StoreConnectionInfo, 'connectionModel'>>;
 
 type StoreConnectionInfoWithConnectionOptions = StoreConnectionInfo &
   Required<Pick<StoreConnectionInfo, 'connectionOptions'>>;
+
+type MigratedStoreConnectionInfo = StoreConnectionInfo &
+  Required<Pick<StoreConnectionInfo, 'secretStorageLocation'>>;
+
+type MigratedStoreConnectionInfoWithConnectionOptions =
+  StoreConnectionInfoWithConnectionOptions & MigratedStoreConnectionInfo;
 
 export default class ConnectionController {
   // This is a map of connection ids to their configurations.
   // These connections can be saved on the session (runtime),
   // on the workspace, or globally in vscode.
   _connections: {
-    [connectionId: string]: StoreConnectionInfoWithConnectionOptions;
+    [connectionId: string]: MigratedStoreConnectionInfoWithConnectionOptions;
   } = {};
   _activeDataService: DataService | null = null;
   _storageController: StorageController;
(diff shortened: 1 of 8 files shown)


src/test/suite/connectionController.test.ts
```diff
@@ -1150,4 +1165,352 @@ suite('Connection Controller Test Suite', function () {
...
+
+            // Assert that we have our secrets now in SecretStorage
+            assert.strictEqual(
+              await testStorageController.getSecret(savedConnection.id),
+              expectedSecret
+            );
+            assert(
+              updatedConnection.secretStorageLocation ===
+                SecretStorageLocation.SecretStorage
+            );
```
addaleax, line 1302:
```suggestion
            assert.strictEqual(
              updatedConnection.secretStorageLocation,
              SecretStorageLocation.SecretStorage
            );
```

(ditto elsewhere)

src/storage/storageController.ts
```diff
@@ -32,6 +32,11 @@ export type ConnectionsFromStorage = {
   [connectionId: string]: StoreConnectionInfo;
 };
 
+export enum SecretStorageLocation {
+  Keytar = 'vscode.Keytar',
+  SecretStorage = 'vscode.SecretStorage',
+}
```
addaleax, line 38: Side note, the main advantage of TS `enum`s over string unions is when used from vanilla JS code, we basically don’t do this anymore so I’d generally recommend using string unions in new code (definitely not a big deal or a blocker)

src/connectionController.ts
```diff
@@ -596,9 +633,15 @@ export default class ConnectionController {
...
   private async _removeSecretsFromKeychain(connectionId: string) {
+    // Even though we migrated to SecretStorage from keytar, we are still removing
+    // secrets from keytar to make sure that we don't leave any left-overs when a
+    // connection is removed. This block can safely be removed after our migration
+    // has been out for some time.
     if (ext.keytarModule) {
       await ext.keytarModule.deletePassword(this._serviceName, connectionId);
     }
+
+    await this._storageController.deleteSecret(connectionId);
```
addaleax, line 644: Does it make sense to do this operation first, in case the `keytar` one fails?

src/connectionController.ts
```diff
@@ -69,20 +71,24 @@ interface ConnectionQuickPicks {
...
+  StoreConnectionInfoWithConnectionOptions & MigratedStoreConnectionInfo;
+
 export default class ConnectionController {
   // This is a map of connection ids to their configurations.
   // These connections can be saved on the session (runtime),
   // on the workspace, or globally in vscode.
   _connections: {
-    [connectionId: string]: StoreConnectionInfoWithConnectionOptions;
+    [connectionId: string]: MigratedStoreConnectionInfoWithConnectionOptions;
   } = {};
```
addaleax, line 92: Just something I noticed while reviewing:

```suggestion
  } = Object.create(null);
```

Shouldn’t be an issue in practice, but this is just good practice for objects with generic keys

src/connectionController.ts
```diff
@@ -116,142 +122,188 @@ export default class ConnectionController {
...
+          connectionInfoWithSecret.secretStorageLocation ===
+          SecretStorageLocation.Keytar
+        ) {
+          connectionsWithSecretsInKeytar.push({
+            connectionId,
+            connectionName: connectionInfo.name,
+          });
+        }
+      }
+    }
```
addaleax, line 164: The previous code here was parallelizing the loop with `.map()/Promise.all()`, is there a reason to not do that anymore?
```suggestion
    connectionsWithSecretsInKeytar = (await Promise.all(totalConnectionEntries.map(async ([connectionId, connectionInfo]) => {
      const connectionInfoWithSecret = await this._getConnectionInfoWithSecrets(
        connectionInfo
      );

      if (connectionInfoWithSecret) {
        connectionsDidChange = true;
        this._connections[connectionId] = connectionInfoWithSecret;

        if (
          connectionInfoWithSecret.secretStorageLocation ===
          SecretStorageLocation.Keytar
        ) {
          return {
            connectionId,
            connectionName: connectionInfo.name,
          });
        }
      }
    })).filter(Boolean);
```

src/connectionController.ts
```diff
@@ -116,142 +122,188 @@ export default class ConnectionController {
...
       );
     }
+  }
+
+  async _getConnectionInfoWithSecrets(
+    connectionInfo: StoreConnectionInfo
+  ): Promise<MigratedStoreConnectionInfoWithConnectionOptions | undefined> {
+    try {
+      if (connectionInfo.connectionModel) {
+        return this._migratePreviouslySavedConnection(
```
addaleax, line 198: Maybe we should rename this given that there are multiple different kinds of migrations now?

src/connectionController.ts
```diff
@@ -116,142 +122,188 @@ export default class ConnectionController {
...
+        StorageLocation.WORKSPACE
+      ),
+    };
+
+    let connectionsDidChange = false;
+    const connectionsWithSecretsInKeytar: Array<{
+      connectionId: string;
+      connectionName: string;
+    }> = [];
+    const totalConnectionEntries = Object.entries(
```
alenakhineika, line 142: Maybe `allConnectionEntries` would be a better naming. `Total` sounds like this is a total number of entries.

src/connectionController.ts
```diff
@@ -116,142 +122,188 @@ export default class ConnectionController {
...
+    if (!unparsedSecrets) {
+      return connectionInfo;
+    }
+
+    const secrets = JSON.parse(unparsedSecrets);
+    const connectionInfoWithSecrets = mergeSecrets(
+      {
+        id: connectionInfo.id,
+        connectionOptions: connectionInfo.connectionOptions,
+      } as ConnectionInfo,
```
addaleax, line 299: I realize this is existing code, but 
```suggestion
      },
```

compiles for me, so unless I’m missing something, we should remove the cast here :slightly_smiling_face:

src/connectionController.ts
```diff
@@ -116,142 +122,188 @@ export default class ConnectionController {
...
+          connectionsWithSecretsInKeytar: connectionsWithSecretsInKeytar.length,
+        }
+      );
+      void vscode.window.showInformationMessage(
+        [
+          'Could not migrate secrets for a few connections. Please review the following connections:',
+          connectionsWithSecretsInKeytar
+            .map(({ connectionName }) => connectionName)
+            .join(', '),
+        ].join('\n')
```
addaleax, line 188: Do we have any intuition about what the cause of this error would typically be in practice? What do we expect the user to do about this? Is there some kind of error message that we could bubble up to the user or a log that we could point them to?

src/connectionController.ts
```diff
@@ -116,142 +122,188 @@ export default class ConnectionController {
...
+      }
+    }
+
+    if (connectionsDidChange) {
+      this.eventEmitter.emit(DataServiceEventTypes.CONNECTIONS_DID_CHANGE);
+    }
+
+    if (connectionsWithSecretsInKeytar.length) {
+      log.error(
+        'Could not migrate secrets for a few connections',
```
addaleax, line 172:
```suggestion
        `Could not migrate secrets for ${connectionsWithSecretsInKeytar.length} connections`,
```

src/connectionController.ts
```diff
@@ -116,142 +122,188 @@ export default class ConnectionController {
...
+      }
+
+      if (!connectionInfo.secretStorageLocation) {
+        return this._migrateKeytarSecrets(
+          connectionInfo as StoreConnectionInfoWithConnectionOptions
+        );
+      }
 
+      // We tried migrating this connection earlier but failed because Keytar was not
+      // available. So we return simply the connection without secrets.
```
addaleax, line 210: Could this happen when the user clicks away a request to unlock their OS keyring because they are repetitive/annoying if they have multiple connections stored? In that case it would be reasonable to assume that a subsequent attempt to get credentials would succeed, right?

src/connectionController.ts
```diff
@@ -116,142 +122,188 @@ export default class ConnectionController {
...
+        StorageVariables.GLOBAL_SAVED_CONNECTIONS,
+        StorageLocation.GLOBAL
+      ),
+      ...this._storageController.get(
+        StorageVariables.WORKSPACE_SAVED_CONNECTIONS,
+        StorageLocation.WORKSPACE
+      ),
+    };
+
+    let connectionsDidChange = false;
```
alenakhineika, line 137: We might also just check `Object.keys(this._connections).length`.

src/connectionController.ts
```diff
@@ -116,142 +122,188 @@ export default class ConnectionController {
...
+          });
+        }
+      }
+    }
+
+    if (connectionsDidChange) {
+      this.eventEmitter.emit(DataServiceEventTypes.CONNECTIONS_DID_CHANGE);
+    }
+
+    if (connectionsWithSecretsInKeytar.length) {
```
alenakhineika, line 170: With this check, we will show the 'Could not migrate secrets for a few connections' message and send telemetry all the time a user loads the extension and they have previously failed migrations from keytar. We should check not for `connectionInfoWithSecret.secretStorageLocation === SecretStorageLocation.Keytar` but how many times it was nothing in secretStorageLocation but we set the SecretStorageLocation.Keytar value. And if we marked something new then we show the message and send telemetry.

src/connectionController.ts
alenakhineika: Wanted to suggest the same. It can be something like `_migrateConnectionsWithConnectionModel` and `_migrateConnectionsWithKeytarSecrets` or similar to indicate that in both cases we update connections based on different attributes.

src/connectionController.ts
himanshusinghs: TIL and thank you for this 😄
I always used to do `for in + if hasOwnProperty` or `Object.keys`, not anymore.

addaleax: `Object.keys()` is still a good thing to use, I think ;)

src/connectionController.ts
himanshusinghs: Not really. This might have also been a mistake as I was too focused in getting that emitted event out of the parallel promises. I will move back to the parallel logic again. Thank you.

src/connectionController.ts
himanshusinghs: Yea this should be good as well. I was probably trying to avoid a situation when `this._connections` has entries already but just checked again and it is very unlikely because it is called only once, when the extension activates.

Will be updating with the suggestion.

src/connectionController.ts
himanshusinghs: This one in particular will happen when the keytar shim will no longer be available for us to use(in the next public release of vscode). This is not actually an error but an error like condition which we plan to notify the users about using the `vscode.showInformationMessage`. Unfortunately the only action users can take about it is to remove the connection from the explorer view and add it again manually

We are not explicitly stating the same in the error message because I was not sure about it but I do think that it might be ok to call that out specifically in the message that we show. Iff we decide to show the message to remove the connection, then we might even just provide users a small button to purge connections that we could not migrate (because of keytar not available anymore).

src/connectionController.ts
himanshusinghs: Yea, for such connections it would be safe to assume that they can be re-tried for migration.

There is one thing not so much clear to me - how / when would a user get a prompt to unlock the keyring? Asking this because it did not happen for me? Could be because I have been using the extension for sometime and might have allowed it at some point or could it be because `keytar` module does not prompt?

addaleax: > could it be because `keytar` module does not prompt?

`keytar` doesn’t do the prompting as far as I know, it integrates with the OS keychains instead, so when exactly the prompt occurs will depend on the OS and its configuration.

(I think we get prompts on macOS when upgrading Electron in Compass, for example. I’m not sure if something similar happens with VSCode at some point.)

src/connectionController.ts
alenakhineika: The message here could be more descriptive saying that keytar is being removed from vscode and we tried to migrate connections, but some failed. @GaurabAryal, maybe the product team could help with better phrasing. I think this is important to explain what happened and the alternatives users have. I am not sure if listing connections by names will work well if users have many saved connections. Some of those can have duplicate default names like `localhost`. This is why my suggestion was to show a broken leaf icon for such connections to easier identify those for users and let them take corresponding action. For this PR the improved message would be enough, but maybe it would be also nice to have s follow-up PR with an improved UX.

src/connectionController.ts
alenakhineika: It happened before for vscode as well but does not happen anymore for a year or so. Not sure if we did something or vscode handles it differently. @Anemy do you have any context about it?

src/connectionController.ts
himanshusinghs: Yea, this makes sense. Sorry for the bad understanding, I have updated it in my recent commit.

src/connectionController.ts
himanshusinghs: Yea @GaurabAryal is in touch with designers to have the broken leaf prepared but until then maybe I will sit with him for a better copy.

src/connectionController.ts
himanshusinghs: In that case - I would assume that the `keytar` module will still be available (unless it was removed) but the `getPassword` call should fail which should throw an error. For this reason we will never mark such connection as migrated and thus they should be good for re-trying. But I should to test that.

src/connectionController.ts
alenakhineika: You could make it simpler and just fetch globalAndWorkspaceConnections once again after migration and compare how many connections with 'vscode.Keytar' was before migration and after.

src/connectionController.ts
Anemy: There's only a prompt for the first time the credential store is to be used for an app (only for some os's that are more stringent with permissions). If something with how the app is built (maybe bundle name? don't remember off the top of my head) changes then this also happens. When we update the electron version in compass-dev we see the prompts as the identifier changes then. As Alena mentioned it happened once around two years ago for VSCode which caused users to see this prompt again. I don't think we need to worry much about that, I don't see it as a bad thing to account for though.
As Anna said, the prompt is from the OS's credential store. _Maybe_ the user could have configured it in a way where they get a prompt each time it's requested.

https://github.com/mongodb-js/compass/blob/main/packages/hadron-build/lib/target.js

src/test/suite/connectionController.test.ts
```diff
@@ -1475,43 +1475,68 @@ suite('Connection Controller Test Suite', function () {
...
+            ...connectionInfo,
             secretStorageLocation: SecretStorageLocation.Keytar,
           } as any)
       );
+      const trackStub = testSandbox.stub(testTelemetryService, 'track');
 
-      const fakeTrack = sinon.stub();
-      testSandbox.replace(testTelemetryService, 'track', fakeTrack);
-
+      // Load the connections
```
alenakhineika, line 1520: The comment is misleading since below with clearAllConnections we clean connections.

src/test/suite/connectionController.test.ts
```diff
@@ -1475,43 +1475,68 @@ suite('Connection Controller Test Suite', function () {
...
     });
 
-    test('should track and also notify the users of failed keytar secrets migration', async () => {
-      await testConnectionController.addNewConnectionStringAndConnect(
-        TEST_DATABASE_URI
-      );
-      const [savedConnection] = testConnectionController.getSavedConnections();
-
-      testConnectionController.clearAllConnections();
+    test('should track and also notify the users of unique failed keytar secrets migration (in the current load of extension)', async () => {
```
alenakhineika, line 1478: We could add one more test to verify that the message is not being shown if all connections are already marked with 'vscode.Keytar'

src/connectionController.ts
himanshusinghs: Understood and thank you for the explanation here @Anemy 
In that case, we can still safely assume that keytar will be available to our extension, it is just that the calls to get secret would fail (if the user did not unlock or took too long to unlock), right?
If so then I would particularly account for that failure (originating from `keytar.getPassword`).

As of now, these errors would just bubble up and eventually get caught in `_getConnectionInfoWithSecret` and the connection will not be marked as migrated (ready to be re-attempted for migration).

alenakhineika: SGTM

src/test/suite/connectionController.test.ts
himanshusinghs: Sounds good, I will add this.

src/test/suite/connectionController.test.ts
himanshusinghs: Sorry for missing that, updating it.

src/storage/storageController.ts
```diff
@@ -32,6 +32,15 @@ export type ConnectionsFromStorage = {
...
 
+export const SecretStorageLocation = {
+  Keytar: 'vscode.Keytar',
+  SecretStorage: 'vscode.SecretStorage',
+} as const;
+
+export type SecretStorageLocationType =
+  | typeof SecretStorageLocation.Keytar
+  | typeof SecretStorageLocation.SecretStorage;
+
```
alenakhineika, line 43: Could it be just `export type SecretStorageLocationType = 'vscode.Keytar' | 'vscode.SecretStorage';`? Do we need at all SecretStorageLocation? cc @addaleax

src/connectionController.ts
```diff
@@ -596,6 +657,11 @@ export default class ConnectionController {
   }
 
   private async _removeSecretsFromKeychain(connectionId: string) {
+    await this._storageController.deleteSecret(connectionId);
+    // Even though we migrated to SecretStorage from keytar, we are still removing
+    // secrets from keytar to make sure that we don't leave any left-overs when a
+    // connection is removed. This block can safely be removed after our migration
+    // has been out for some time.
     if (ext.keytarModule) {
       await ext.keytarModule.deletePassword(this._serviceName, connectionId);
```
Anemy, line 666: Does this error if the password doesn't exist? If so maybe it's worth `try` `catch`ing as we could have already deleted it in the migration.

src/connectionController.ts
```diff
@@ -116,142 +128,206 @@ export default class ConnectionController {
...
-        const connectionInfoWithSecrets =
-          await this._getConnectionInfoWithSecrets(
-            savedConnections[connectionId]
-          );
-
-        // Save connection info with secrets to extension memory.
-        if (connectionInfoWithSecrets) {
-          this._connections[connectionId] = connectionInfoWithSecrets;
-        }
+    await this._saveConnectionWithSecrets(migratedConnectionInfoWithSecrets);
```
Anemy, line 306: When this succeeds, should we (or do we already?) also remove the keytar password?

Anemy (commented):
Looking good, left a question on keytar handling. Also +1 on updating the text on the info message we show for those cases where things can't be loaded.

src/connectionController.ts
```diff
@@ -69,21 +72,30 @@ interface ConnectionQuickPicks {
...
-  connectionId: string;
-  secrets: ConnectionSecrets;
-}
+type StoreConnectionInfoWithConnectionModel = StoreConnectionInfo &
+  Required<Pick<StoreConnectionInfo, 'connectionModel'>>;
 
 type StoreConnectionInfoWithConnectionOptions = StoreConnectionInfo &
   Required<Pick<StoreConnectionInfo, 'connectionOptions'>>;
 
+type MigratedStoreConnectionInfo = StoreConnectionInfo &
```
alenakhineika, line 81: Maybe `StoreConnectionInfoWithSecretStorageLocation` would match closer to other used variables.

src/connectionController.ts
alenakhineika: We talked about this and agreed to keep keytar password in case users try to use an older version of vscode, that is very unlikely, but still might be the case (and was useful for debugging this PR :)). We clean those secrets when a user deletes a connection, then secrets are deleted both from vscode secret storage and keytar.

src/storage/storageController.ts
himanshusinghs: Yea I think it could be. I probably did it this way just so I also have a string const to use in places where the types are not available from the context (for example tests and a couple other places in the code as well where we declare a connection info like object with secret storage).

alenakhineika:
I feel like the PR is good to be merged when the text message we show to users in case of migration failure is updated 👍 great job! 👏

src/connectionController.ts
himanshusinghs: Yea I think that is worth adding. On it.

src/connectionController.ts
himanshusinghs: Yea it definitely would. Thank you :)

src/connectionController.ts
himanshusinghs: Earlier I was doing that but in this iteration I am not because I just thought that it would then not be an explicit user action which I tried avoiding. An argument against it could be to not leave orphan data behind. While writing this I started leaning a little more towards removing the secret from keytar as well. Let me know if we all think that is what we should do.

Anemy: I'm cool with either 👍 
Mostly wanted to check it was something we considered.

src/connectionController.ts
himanshusinghs: Addressed this in the [following commit](https://github.com/mongodb-js/vscode/pull/552/commits/808e60e8b6d1b60ee5d71103c2ff858bb8baaa7d)

src/connectionController.ts
himanshusinghs: Addressed this in the [following commit](https://github.com/mongodb-js/vscode/pull/552/commits/808e60e8b6d1b60ee5d71103c2ff858bb8baaa7d)

src/connectionController.ts
himanshusinghs: I just addressed this in my [commit here](https://github.com/mongodb-js/vscode/pull/552/commits/4250dd6a6a984fff425c50701aa8ea67c4c4286e). The connections which will fail migration for anything other than keytar not available are not marked as migrated and hence can be re-attempted again in the next load.

There is a kinda bad side effect of this which is - such connections (for which our `_migrateConnectionWithKeytarSecrets` rejected) will not appear in our Connection tree when the extension loads. I was also able to confirm from the [following code](https://github.com/mongodb-js/vscode/blob/5a9a6b4ac305ddb977e1371b33e13a822b6d16dc/src/connectionController.ts#L178-L208) that this is an existing behaviour.

I have created [this ticket](https://jira.mongodb.org/browse/VSCODE-443) to define and agree upon a behaviour of "loading of the connections" when we are not able to fetch the secrets (for whichever reason) for some of the connections. Please let me know if this is something we should absolutely fix within this PR.

src/connectionController.ts
himanshusinghs: PR updated with the unproved copy

alenakhineika (approved):

PR #563, opened 2023-07-19
chore: add language server logs VSCODE-447, VSCODE-448, VSCODE-449

alenakhineika:
## Description

Split from [VSCODE-431](https://jira.mongodb.org/browse/VSCODE-431) (Add additional logging for log file) to shed some light on [VSCODE-448](https://jira.mongodb.org/browse/VSCODE-448) (The playground’s active connection does not match the extension’s active connection).
- Add the initial extension log
- Log reconnecting events
- Log the extension path
- Log deactivating extension

Additionally stub the 'trackNewConnection' telemetry event and skip the telemetry live connection test to fix CI because of the [issue with the Insiders](https://jira.mongodb.org/browse/VSCODE-449) 1.81.0.

The new logs have shown, that some unexpected errors may cause the language server to shut down, which restarts automatically but then misses required for its functioning parameters. We will handle those cases when the MongoDBService re-initializes itself and use the previously used extensionPath and connection details from LanguageServerController. We will also inform users that `'An internal error has occurred. The playground services have been restored.'`.

The current PR does not solve the cause of the VSCODE-448 problem (the ticket will remain open for further investigation), but it improves the error messaging and prevents the extension from crashing.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


src/editors/playgroundController.ts
```diff
@@ -245,35 +249,23 @@ export default class PlaygroundController {
...
+    this._activeConnectionCodeLensProvider.refresh();
 
-      return;
+    if (dataService && connectionId) {
+      mongoClientOption =
+        this._connectionController.getMongoClientConnectionOptions();
     }
 
-    await this._languageServerController.connectToServiceProvider({
+    await this._languageServerController.activeConnectionChanged({
```
alenakhineika, line 264: Remove the hassle of disconnecting and connecting in the playground and language controller, when we can do it only in the MongoDB service.

src/explorer/explorerController.ts
```diff
@@ -38,14 +34,12 @@ export default class ExplorerController {
   };
 
   activateConnectionsTreeView(): void {
-    log.info('Activating explorer controller...');
```
alenakhineika, line 41: Such logs turned out to be distracting and not helpful for debugging, so removing them.

src/language/languageServerController.ts
```diff
@@ -143,15 +160,22 @@ export default class LanguageServerController {
...
       this._source.token
     );
 
     this._isExecutingInProgress = false;
 
-    return result;
+    log.info('Evaluate response', {
+      namespace: res?.result?.namespace,
+      type: res?.result?.type,
+      outputLength: JSON.stringify(res?.result?.content || '').length,
```
alenakhineika, line 174: Same with output. Maybe we receive something big that crashes a rendering of the response virtual editor.

src/editors/playgroundController.ts
```diff
@@ -420,36 +413,21 @@ export default class PlaygroundController {
...
-        result,
-        this._isPartialRun,
-        result ? false : true
-      );
-
-      return result;
-    } catch (err: any) {
-      // We re-initialize the language server when we encounter an error.
-      // This happens when the language server worker runs out of memory, can't be revitalized, and restarts.
-      if (err?.code === -32097) {
```
alenakhineika, line 442: I _think_ we don't need this part anymore, since the MongoDBService reinitialises itself after LS automatically restarting, which will handle all error codes, that can crash the server, not only this particular one.

src/logging.ts
```diff
@@ -4,7 +4,7 @@ import util from 'util';
 
 class Logger implements ILogger {
   static channel: vscode.OutputChannel =
-    vscode.window.createOutputChannel('mongodb');
+    vscode.window.createOutputChannel('MongoDB');
```
Anemy, line 7: How do you feel about
```suggestion
    vscode.window.createOutputChannel('MongoDB Extension');
```

src/extension.ts
```diff
@@ -27,28 +29,50 @@ let mdbExtension: MDBExtensionController;
...
+    workspaceStoragePath: context.storageUri?.path,
+    globalStoragePath: context.globalStorageUri.path,
+    defaultConnectionSavingLocation,
+    hasKeytar,
+    buildInfo: {
+      nodeVersion: process.version,
+      runtimePlatform: process.platform,
+      runtimeArch: process.arch,
+    },
+  });
```
Anemy, line 62: Niceeee

src/explorer/playgroundsTree.ts
```diff
@@ -50,7 +50,9 @@ export default class PlaygroundsTree
     });
 
     treeView.onDidExpandElement(async (event: any): Promise<void> => {
-      log.info('Tree item was expanded', event.element.label);
+      log.info('Playground tree item was expanded', {
+        laygroundName: event.element.label,
```
Anemy, line 54:
```suggestion
        playgroundName: event.element.label,
```

src/editors/playgroundController.ts
Anemy: Nice, looks like it is yup. Do we have logging there?
Also, just mentioning, we have a playground in the project called `out-of-memory-test.mongodb` that can be used for testing this.

The error we now show here is:
Screenshot 2023-07-25 at 12 30 12 PM

Instead of
```
An error occurred when running the playground. This can occur when the playground runner runs out of memory.
```
Is that okay? Should we improve the messaging somehow?

Anemy (commented):
Niceee, I really like the logging improvements. Left a couple questions/suggestions.

src/editors/playgroundController.ts
alenakhineika: Good point. A put back try catch to log out of memory and similar errors and added international test for this scenario.

src/extension.ts
```diff
@@ -27,28 +29,50 @@ let mdbExtension: MDBExtensionController;
...
+    hasKeytar = true;
   } catch (err) {
     // Couldn't load keytar, proceed without storing & loading connections.
   }
 
+  const defaultConnectionSavingLocation = vscode.workspace
+    .getConfiguration('mdb.connectionSaving')
+    .get('defaultConnectionSavingLocation');
+
+  log.info('Activating extension...', {
```
Anemy, line 46: I'm thinking a test for this log might be a very useful one for debugging the environment. Not needed for this pr if we don't want to put the time in now, just a thought.

Anemy (approved):
lgtm :shipit:

src/connectionController.ts
```diff
@@ -551,7 +559,7 @@ export default class ConnectionController {
       throw connectError;
     }
 
-    log.info('Successfully connected');
+    log.info('Successfully connected', { connectionId });
```
Anemy, line 562: Adding the connection name as well might help debugging as it can be hard to infer which connection the connectionId belongs to.

src/language/languageServerController.ts
```diff
@@ -101,29 +110,61 @@ export default class LanguageServerController {
...
+        connectionString: this._currentConnectionString,
+        connectionOptions: this._currentConnectionOptions,
+      });
+      // The _currentConnectionId is null when the extension initially activates.
+      // When an unexpected error occurs and the language server restarts after failure,
+      // it loses all default configurations.
+      // But the LanguageServerController still stores the previous _currentConnectionId,
+      // based on what we can tell that the services have been restored.
+      if (this._currentConnectionId) {
+        void vscode.window.showErrorMessage(
```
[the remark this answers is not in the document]
alenakhineika: Since I returned try/catch to the playground controller it should also go there, agree!

src/extension.ts
alenakhineika: What would you test here? Do we have tests for logs already somewhere in compass?

src/connectionController.ts
alenakhineika: We don't pass the connection name anywhere outside the connection controller, not sure we can use it to match connectivity issues here and for example in LS. But I added a name to the explorer logs, so we can find the info what was expanded:
https://github.com/mongodb-js/vscode/pull/563/files#diff-06a415e2a54193ec0955688552bf1cb4813ac5868fc9392083f3c37f8f65a8c4R67-R69

Anemy (approved):

himanshusinghs:
🎉

PR #567, opened 2023-07-21
feat: added Export to Go support VSCODE-411

GaurabAryal:
## Description

Exposes "Export to GO" support.
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


package.json
```diff
@@ -243,6 +243,10 @@
         "command": "mdb.exportToRuby",
         "title": "MongoDB: Export To Ruby"
       },
+      {
+        "command": "mdb.exportToGo",
```
alenakhineika, line 247: You should also specify that this command is available only when a user is connected and for playground only and not other files opened in vscode.

himanshusinghs (commented):
Other than what Alena mentioned, I would only request a small change in PR title aligning with the guideline [here](https://github.com/10gen/devtools-internal-docs/blob/0b37352260d7d5d08a8ca493ca4c7e435de6cd17/onboarding/code-review-guidelines.md?plain=1#L31). Helps creating better, semantically readable release notes towards the end when we do the releases.

package.json
GaurabAryal: Done!

himanshusinghs (approved):

PR #572, opened 2023-08-02
chore: only load keytar during the migration process VSCODE-450

kmruiz:
Only load keytar during the migration of stored connections to process to the new SecretStorageAPI, maintained by the VSCode team. This is a requirement specified in VSCODE-450. This is related to #546.

## Description

Keytar is removed from the activation of the plugin, and only loaded when the migration process starts. 

### Checklist

## Motivation and Context

Please, see #546 for more information.

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -13,7 +13,7 @@ import ConnectionString from 'mongodb-connection-string-url';
 import { EventEmitter } from 'events';
 import type { MongoClientOptions } from 'mongodb';
 import { v4 as uuidv4 } from 'uuid';
-
+import { createKeytar } from './utils/keytar';
 import { CONNECTION_STATUS } from './views/webview-app/extension-app-message-constants';
 import { createLogger } from './logging';
 import { ext } from './extensionConstants';
@@ -296,6 +296,13 @@ export default class ConnectionController {
   async _migrateConnectionWithKeytarSecrets(
     savedConnectionInfo: StoreConnectionInfoWithConnectionOptions
   ): Promise<MigratedStoreConnectionInfoWithConnectionOptions | undefined> {
+    try {
+      ext.keytarModule =
+        ext.keytarModule === undefined ? createKeytar() : ext.keytarModule;
+    } catch (err) {
+      // Couldn't load keytar, proceed without storing & loading connections.
+    }
+
     // If the Keytar module is not available, we simply mark the connections
     // storage as Keytar and return
     if (!ext.keytarModule) {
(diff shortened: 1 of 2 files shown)


himanshusinghs (approved):
Looks good! 🎉 
I was thinking, we probably won't need to assign the created keytar module to `ext` but that would also entail a bunch of refactoring in our tests. Maybe we can leave it as you have it right now, since it all will anyways be removed at some point.

src/connectionController.ts
```diff
@@ -13,7 +13,7 @@ import ConnectionString from 'mongodb-connection-string-url';
 import { EventEmitter } from 'events';
 import type { MongoClientOptions } from 'mongodb';
 import { v4 as uuidv4 } from 'uuid';
-
+import { createKeytar } from './utils/keytar';
```
alenakhineika, line 16: Should we maybe update this model, so it remembers that keytar require failed once, and we don't try to resolve it again for each next connection that goes through the migration? With our webpack configuration, the failed require won't be cached: https://github.com/mongodb-js/vscode/blob/711587b9755127056d70b0ff91a482cb318957f1/webpack.config.js#L75-L76

alenakhineika: Another way to handle it is to move the createKeytar call to the loadSavedConnections function, where can see if there are connections without secretStorageLocation and then try to require it once for the whole migration attempt.

alenakhineika: I would require it once in loadSavedConnections, but you can decide what works better implementation-wise.

kmruiz: Moving it to loadSavedConnections needs some changes in the tests, as some of them fail. 

I'll keep the current version and merge.

Thanks a lot!

alenakhineika: What will happen when vscode removes keytar from vscode and there are migrations that require a migration? Are we ok with resolving the keytar module for each connection? Will this make the launch of the extension slower than it could be?

alenakhineika: Moving createKeytar to loadSavedConnections should note brake tests. Have you tried something like this?

```typescript
const hasConnectionsThatDidNotMigrateEarlier =
      !!globalAndWorkspaceConnections.some(([, connectionInfo]) => !connectionInfo.storageLocation);

if (hasConnectionsThatDidNotMigrateEarlier) {
  try {
    ext.keytarModule =
    ext.keytarModule === undefined ? createKeytar() : ext.keytarModule;
  } catch (err) {
  }
}
```

PR #574, opened 2023-08-08
chore: move keytar to loadSavedConnections VSCODE-450

kmruiz:
## Description
Try to move keytar loading into loadSavedConnections so it's only loaded once per migration.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


src/connectionController.ts
```diff
@@ -143,6 +143,13 @@ export default class ConnectionController {
   }
 
   async loadSavedConnections(): Promise<void> {
+    try {
```
alenakhineika, line 146: You can check if we have something to migrate and only then attempt to requite keytar: https://github.com/mongodb-js/vscode/pull/572#discussion_r1286274579

src/connectionController.ts
```diff
@@ -170,6 +170,15 @@ export default class ConnectionController {
         []
       );
 
+    if (connectionIdsThatDidNotMigrateEarlier.length > 0) {
+      try {
+        ext.keytarModule =
+          ext.keytarModule === undefined ? createKeytar() : ext.keytarModule;
```
addaleax, line 176: I know this is pre-existing code, but this could just be

```suggestion
        ext.keytarModule ??= createKeytar();
```

right?

src/connectionController.ts
kmruiz: Moved to a check after connectionIdsThatDidNotMigrateEarlier.

src/connectionController.ts
kmruiz: Hmmm, it doesn't have the same behaviour (sadly). `??=` will also overwrite keytarModule on null, and we only want to do it on undefined (due existing test behaviour).

kmruiz:
@alenakhineika that was the tests that were failing also on my local, they don't seem to be flaky, but some change of behaviour when we moved keytar to loadSavedConnections:

https://github.com/mongodb-js/vscode/actions/runs/5798050950/job/15714958344?pr=574#step:9:1290

Any clue? I personally won't put much more effort on this if the fix of the tests are too complex, as this is going to be run theoretically once during the migration and later we can just remove these 3 lines of code. 

I won't be too concerned about the performance impact of this, as people don't have thousands of connections to migrate, probably just less than 5/10 and the storages are going to be far slower than failing loading the module.

src/connectionController.ts
alenakhineika: It is not the same :) I would rename the existing `connectionIdsThatDidNotMigrateEarlier` to `connectionIdsThatFailedMigrationEarlier`, because this is actually what happened: we tried to migrate, but it failed.

```typescript
connectionInfo.secretStorageLocation === SecretStorageLocation.Keytar
```

 You need a different check for keytar: check that we have never tried to migrate some connections.

```typescript
const hasConnectionsThatDidNotMigrateEarlier =
      !!globalAndWorkspaceConnections.some(([, connectionInfo]) => !connectionInfo.storageLocation);
```

kmruiz: Pushed the change 😄

alenakhineika:
We do want to ensure that we require keytar once per extension activation to [provide an accurate telemetry for VSCode team](https://github.com/mongodb-js/vscode/issues/546#issuecomment-1649900937). The CI failure that you experienced is not related to the code moving you did. This started to happen because of a new VSCode release between your PRs, that introduced DNS result order changes. [I have merged a fix for this issue](https://github.com/mongodb-js/vscode/pull/575), so rebase from main to unblock your PR.

alenakhineika (approved):

PR #590, opened 2023-09-01
fix(playgrounds): stringify non-string types for playground output VSCODE-466

Anemy:
[VSCODE-466](https://jira.mongodb.org/browse/VSCODE-466)

Happens with both `console.log` and `print`.
Using the following example:
```
console.log('test');
console.log({a: 2});
console.log(['a', {a: 3}])
```
| before | after |
| - | - |
| Screenshot 2023-09-01 at 12 35 06 PM | Screenshot 2023-09-01 at 12 33 07 PM |

### DIFF
--- a/src/editors/playgroundController.ts
+++ b/src/editors/playgroundController.ts
@@ -5,6 +5,7 @@ import { ProgressLocation } from 'vscode';
 import vm from 'vm';
 import os from 'os';
 import transpiler from 'bson-transpilers';
+import util from 'util';
 
 import type ActiveConnectionCodeLensProvider from './activeConnectionCodeLensProvider';
 import type PlaygroundSelectedCodeActionProvider from './playgroundSelectedCodeActionProvider';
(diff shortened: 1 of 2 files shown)


alenakhineika (approved):
👍 Yeah having a test would be nice!

src/editors/playgroundController.ts
```diff
@@ -573,7 +573,11 @@ export default class PlaygroundController {
 
     if (evaluateResponse?.outputLines?.length) {
       for (const line of evaluateResponse.outputLines) {
-        this._outputChannel.appendLine(line.content);
+        this._outputChannel.appendLine(
+          typeof line.content === 'string'
+            ? line.content
+            : JSON.stringify(line.content)
```
addaleax, line 579: Is `JSON.stringify()` the right stringifier here? We’re showing this in the output to humans, maybe something like `util.inspect()` would make more sense, if a user really wants JSON output they can always use `print(EJSON.stringify(...))` (or [hopefully soon](https://jira.mongodb.org/browse/MONGOSH-1361) `printejson(...)`)?

alenakhineika: I like the idea of using util.inspect() here 👍

Anemy: Yes it definitely would, thanks for the suggestion.
Updated and added a test.

addaleax (approved):

PR #592, opened 2023-09-12
feat: update mongosh to 2.0.0 and driver to 6.0.0 VSCODE-453

addaleax:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [x] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


package.json
```diff
@@ -1089,6 +1090,11 @@
     "xvfb-maybe": "^0.2.1",
     "yargs-parser": "^20.2.9"
   },
+  "overrides": {
+    "mongodb-connection-model": {
+      "@mongodb-js/compass-utils": "0.3.4"
```
addaleax, line 1095: `@mongodb-js/compass-utils` doesn’t export `getStoragePaths()` in the most recent version anymore which makes `require()`ing `mongodb-connection-model` fail, technically a breaking change and maybe we should have actually labelled https://github.com/mongodb-js/compass/commit/2c91060412b423097939120dd759492f497d31cb as `feat!:` rather than `chore(fs):` …

src/connectionController.ts
```diff
@@ -31,6 +24,24 @@ import { StorageVariables } from './storage';
...
 import type TelemetryService from './telemetry/telemetryService';
 import LINKS from './utils/links';
+import type {
+  ConnectionInfo as ConnectionInfoFromLegacyDS,
+  ConnectionOptions as ConnectionOptionsFromLegacyDS,
+} from 'mongodb-data-service-legacy';
+import {
+  getConnectionTitle,
+  extractSecrets,
+  mergeSecrets,
```
addaleax, line 34: These are now exported by `@mongodb-js/connection-storage`, and we may be able to use it, but that has a hard, non-optional/non-lazy dependency on `keytar`, so I don’t think we can use it here. Together with the item below, that made it seem very reasonable to just use a prior version of `mongodb-data-service` and take care of this issue when we actually do the VSCode connection improvements project.

src/connectionController.ts
```diff
@@ -31,6 +24,24 @@ import { StorageVariables } from './storage';
...
+  convertConnectionModelToInfo,
```
addaleax, line 35: We could copy the source code for this one, but then we’d still have a dependency on `mongodb-connection-model` and would need to use the outdated version of that (or, we could, recursively, also add the source code for *that* entire package).

src/connectionController.ts
```diff
@@ -31,6 +24,24 @@ import { StorageVariables } from './storage';
...
+} from 'mongodb-data-service-legacy';
+
+export function launderConnectionOptionTypeFromLegacyToCurrent(
+  opts: ConnectionOptionsFromLegacyDS
+): ConnectionOptionsFromCurrentDS {
+  // Ensure that, at most, the types for OIDC mismatch here.
+  return opts as Omit<typeof opts, 'oidc'>;
```
addaleax, line 42: We don’t have proper OIDC support yet anyway

src/language/mongoDBService.ts
```diff
@@ -352,17 +352,19 @@ export default class MongoDBService {
...
+          collectionName,
+          {},
+          // TODO: figure out if we need parseSchema for one field at all.
+          // For one document we can simply get deep object keys,
+          // or we could analyze the schema for at least 5-10 documents.
+          // The current behaviour came from Compass when we do the same:
+          // https://github.com/mongodb-js/compass/blob/main/packages/compass-field-store/src/stores/store.js#L193
+          { limit: 1 }
+        )
+        .toArray();
```
addaleax, line 367: `parseSchema` would want a cursor from a 5.x driver right now, should probably update this as well at some point but since we only use a single document anyway `.toArray()` is a nice way to decouple the types here

kmruiz (approved):

package.json
mabaasit: I didn't know that vscode is so coupled with compass. Actually I didnt know that we are using legacy compass connections in vscode at all.

PR #602, opened 2023-10-19
fix: snyk-js-babeltraverse-5962462 VSCODE-475

alenakhineika:


package.json
```diff
@@ -973,7 +973,7 @@
   },
   "dependencies": {
     "@babel/parser": "^7.22.6",
-    "@babel/traverse": "^7.22.6",
+    "@babel/traverse": "^7.23.2",
```
addaleax, line 976: Also needs to update the mongosh dependencies, right?

alenakhineika: Yeah, did not add it yet to reduce changes to catch issues with CI.

src/test/suite/playground.test.ts
```diff
@@ -97,64 +97,64 @@ suite('Playground', function () {
     await vscode.commands.executeCommand('mdb.createPlayground');
 
     const editor = vscode.window.activeTextEditor;
-    if (!editor) {
-      throw new Error('Window active text editor is undefined');
+    expect(editor).to.exist;
+
+    if (editor) {
```
addaleax, line 102: nit: Why are we wrapping the rest of the code in an `if (editor)` conditional if we just asserted that `editor` is not null-ish? I’m guessing it’s related to typescript not understanding chai assertions as type assertions, but the previous code seemed like exactly the right thing to do in that case.

addaleax (approved):

addaleax (approved):

PR #614, opened 2023-11-24
feat(webview-controller): added a feature flag system to work with optional view rendering - VSCODE-486

himanshusinghs:
## Description

This PR adds a feature flag using environment variables. The variable `MDB_USE_NEW_CONNECTION_FORM` when set to `true` will enable the new overview page and the new connection form.

The existing webview app, its store and its tests have been moved to a subfolder `legacy`. Since we plan to remove the legacy code as part of this project, this move will make it easier to do the same.

The test cases for webview will now be written using react testing library as also documented in tech design.

To test the new overview page:
- install the local build of the extension
- run a new instance of code with the environment variable set - `MDB_USE_NEW_CONNECTION_FORM=TRUE code`
- open the extension overview page

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/test/suite/views/webview-app/components/connect-form/general-tab/general-tab.test.tsx
himanshusinghs, line 1: This file is not deleted but moved to a difference folder, not sure why this is being shown as deleted

src/test/suite/views/webviewController.test.ts
```diff
@@ -849,4 +849,55 @@ suite('Webview Test Suite', () => {
...
+      storageController: testStorageController,
+      telemetryService: testTelemetryService,
+    });
+    let messageReceived;
+
+    const fakeWebview = {
+      html: '',
+      postMessage: (message): void => {
+        assert(message.command === 'FEATURE_FLAGS_RESULTS');
+        assert(typeof message.featureFlags === 'object');
```
addaleax, line 871:
```suggestion
        assert.strictEqual(message.command, 'FEATURE_FLAGS_RESULTS');
        assert.strictEqual(typeof message.featureFlags, 'object');
```

src/views/webview-app/legacy/app-with-store.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { createStore } from 'redux';
+
+import App from './components/app';
+
+import { initialState, rootReducer } from './store/store';
+
+const store = createStore(rootReducer, initialState);
+
+export default () => (
+  <Provider store={store}>
```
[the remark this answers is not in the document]
himanshusinghs: Ohh thank you for catching this, yes the one in index.tsx should be removed. Updating it.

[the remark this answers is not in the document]
himanshusinghs: addressed in [d93f5cf](https://github.com/mongodb-js/vscode/pull/614/commits/d93f5cf8eebd05467ed4108eda2984df7740a458)

src/featureFlags.ts
```diff
@@ -1,5 +1,20 @@
...
+const FEATURE_FLAGS = {
   useNewConnectionForm: `${process.env.MDB_USE_NEW_CONNECTION_FORM}` === 'true',
-});
+};
+
+export type FeatureFlag = keyof typeof FEATURE_FLAGS;
+
+export const getFeatureFlag = (flag: FeatureFlag) => {
+  if (typeof window === 'object') {
+    return (window as any).MDB_FEATURE_FLAGS?.[flag];
```
addaleax, line 9: Might be easier to use `globalThis` instead of `window` if some tests run in e.g. a plain Node.js environemnt?

himanshusinghs: the idea here was to fetch the flags from an this `MDB_FEATURE_FLAGS` object attached on window when in html/react world. For the rest, I believe the return below would work just fine, right?

himanshusinghs: I guess I meant to say that in plan node environment I was relying on the return below to do the trick.

addaleax: Sure, whatever makes more sense to you :)

addaleax (approved):

src/test/suite/views/webviewController.test.ts
Anemy: I'm thinking we should add a lint rule for this 😅 . Or start using chai everywhere in the extension. I'm kind of leaning towards chai.

src/test/suite/views/webview-app/app.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+import * as featureFlags from '../../../../featureFlags';
+import { render, screen } from '@testing-library/react';
+import App from '../../../../views/webview-app/app';
+
+describe('App Component Test Suite', () => {
+  afterEach(() => sinon.restore());
+  test('it renders the old overview page when useNewConnectionForm is falsy', () => {
+    sinon.stub(featureFlags, 'getFeatureFlag').returns(false);
+    render(<App />);
+    assert.doesNotThrow(() => screen.getAllByTestId('legacy-app'));
```
Anemy, line 13: How do you feel about using `chai` `expect` in these new tests? That way we have a bit better error reporting and avoid needing `strictEquals` in many places.

Anemy (approved):

src/test/suite/views/webview-app/app.test.tsx
himanshusinghs: Ohh I missed that, yes. My intention was to keep consistent with the packages we use in other places, surely I missed this one. I will update that.

himanshusinghs:
Sorry I was focused on getting the CI errors out. Hopefully the last two commits fixes the CI issues. As soon as I see green, I will merge.

himanshusinghs:
Merging this for now as it is also happening on main already, will handle the snyk failing issue separately.

PR #621, opened 2023-12-05
chore(playground): codelens for active connection will inform about default connected database VSCODE-316

himanshusinghs:
## Description
Original issue highlighted in VSCODE-316 (playground should run in database specified in connection string), is not reproducible anymore. It works as expected - if no `use('database')` call is specified in playground and there is a database specified in connection string then the playground will run in the database specified in connection string.

With this PR, we will now additionally highlight, in the playground, the default database that the playground will run in, if there is one specified in connection string.

|Without default database in connection string|With default database in connection string|
|---------------------------------------------|------------------------------------------|
|`mongodb+srv://<credentials>@host`|`mongodb+srv://<credentials>@host/countries`|
|image|image|

We additionally also provide autocomplete for the collection names on the `db` object when connected using a connection string with default database and there is no explicit `use('DB')` call in the playground.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/language/mongoDBService.ts
+++ b/src/language/mongoDBService.ts
@@ -14,7 +14,6 @@ import type {
   MarkupContent,
   Diagnostic,
 } from 'vscode-languageserver/node';
-import ConnectionString from 'mongodb-connection-string-url';
 import { CliServiceProvider } from '@mongosh/service-provider-server';
 import type { Document } from '@mongosh/service-provider-core';
 import { getFilteredCompletions } from '@mongodb-js/mongodb-constants';
@@ -40,6 +39,7 @@ import type { CompletionState, NamespaceState } from './visitor';
 import LINKS from '../utils/links';
 
 import DIAGNOSTIC_CODES from './diagnosticCodes';
+import { getDBFromConnectionString } from '../utils/connection-string-db';
 
 const PROJECT = '$project';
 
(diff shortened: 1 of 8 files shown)


src/language/mongoDBService.ts
```diff
@@ -96,6 +97,16 @@ export default class MongoDBService {
...
   }
 
+  get defaultConnectedDB(): string | null {
+    if (!this.connectionString) {
+      return null;
+    }
+    const connectionString = new ConnectionString(this.connectionString);
+    return connectionString.pathname !== '/'
+      ? connectionString.pathname.substring(1)
+      : null;
```
Anemy, line 107: We use this same code in `getActiveConnectionDefaultDB`, how do you feel about adding a shared helper somewhere?
I'm thinking then we can remove the `getActiveConnectionDefaultDB` from connection controller and use only the `getActiveConnectionString` to keep the connection controller interface a bit smaller.

Anemy (approved):
lgtm, one suggestion on reducing duplicated code

src/language/mongoDBService.ts
himanshusinghs: yea that sounds about right, i will take care of that.

himanshusinghs: addressed in [cda24a](https://github.com/mongodb-js/vscode/pull/621/commits/cda24a93c4de52512432be2733da2a168bf20979)

alenakhineika (approved):

PR #625, opened 2023-12-13
chore(connections): remove keytar, keytar migration, and connection model migration VSCODE-499

Anemy:
VSCODE-499

Removes the keytar migration and the connection model migration we previously had. For anyone who didn't open the extension in the last ~5 months with the migration they will lose their saved connections. Double checked with Gaurab, he gave the ok.


src/storage/storageController.ts
```diff
@@ -32,9 +32,12 @@ export type ConnectionsFromStorage = {
   [connectionId: string]: StoreConnectionInfo;
 };
 
+// Keytar is deprecated and no longer used. All new
+// connections use 'SecretStorage'.
```
mcasimir, line 36: can't we just delete this?

Anemy: The other values here could still be on someone's stored connection if they open vscode for the first time in 5 months, I reckon we can remove these types but we should still account for when the `secretStorageLocation` isn't `SecretStorage`. Leaving these on the type I thought might give more insight into how it can be different. Do you prefer it deleted? I'm cool with either.

addaleax (approved):

PR #627, opened 2023-12-15
chore: add connection storage, simplify connection controller and storage controller interfaces

Anemy:
This branches off of the `remove-keytar` branch in https://github.com/mongodb-js/vscode/pull/625

This adds a `ConnectionStorage` class which abstracts connection storage logic from the `ConnectionController`. This pr also removes connection related logic from the `StorageController`. We can further streamline the `StorageController` interface in the future, this pr gets us some of the way there.


src/storage/connectionStorage.ts
```diff
@@ -0,0 +1,314 @@
...
+    return (
+      await Promise.all(
+        globalAndWorkspaceConnections.map(async (connectionInfo) => {
+          const connectionInfoWithSecrets =
+            await this._getConnectionInfoWithSecrets(connectionInfo);
+          if (!connectionInfoWithSecrets) {
+            return;
+          }
+
+          return connectionInfoWithSecrets;
```
addaleax, line 241: No big deal, but just curious, how is this different from
```suggestion
          return await this._getConnectionInfoWithSecrets(connectionInfo);
```
?

addaleax (approved):

PR #630, opened 2023-12-18
feat: add OIDC auth support, enable in new form VSCODE-354

Anemy:
VSCODE-354

Tests to be added in https://jira.mongodb.org/browse/VSCODE-414 . New connection form is still feature flagged.

https://github.com/mongodb-js/devtools-docker-test-envs/tree/main/docker/oidc/mock-oidc-provider is useful for testing manually for now.


scripts/check-vsix-size.ts
```diff
@@ -12,7 +12,7 @@ const vsixFileName = path.resolve(
 );
 const size = fs.statSync(vsixFileName).size;
 
-const maxSize = 8 * 1000000; // 8 MB
+const maxSize = 8.5 * 1000000; // 8.5 MB
```
addaleax, line 15:
```suggestion
const maxSize = 8_500_000; // 8.5 MB
```

(might not even need the comment anymore then :))

addaleax (approved):

PR #633, opened 2023-12-29
VSCODE-504: streams support in playground

shaketbaby:
## Description

Add streams support to playground:
- autocomplete stream processors and operations
- console logs are printed to the output channel continuously as they are logged

### Checklist

## Motivation and Context

- [ ] New feature

## Types of changes


src/editors/playgroundController.ts
```diff
@@ -159,6 +158,8 @@ export default class PlaygroundController {
       }
     );
 
+    languageServerController._consoleOutputChannel = this._outputChannel;
```
Anemy, line 161: Now that the language server controller is outputting to the console, the `_outputChannel` in `playgroundController` is only used when clearing the output before evaluating. How do you feel about removing the `_outputChannel` here so it's only in the `languageServerController`? We can `outputChannel.clear()` in the `evaluate` function in the `languageServerController`.

src/language/mongoDBService.ts
```diff
@@ -245,9 +267,16 @@ export default class MongoDBService {
...
 
-        worker?.on(
-          'message',
-          ({ error, data }: { data?: ShellEvaluateResult; error?: any }) => {
+        worker?.on('message', ({ name, payload }) => {
+          if (name === ServerCommands.SHOW_CONSOLE_OUTPUT) {
+            void this._connection.sendNotification(name, payload);
+          }
+
+          if (name === ServerCommands.EXECUTE_CODE_FROM_PLAYGROUND) {
```
Anemy, line 275: Not quite related to these changes, currently we use this `EXECUTE_CODE_FROM_PLAYGROUND` command enum both for the request to the language server worker, and for the response from the worker. 
How do you feel about having two separate commands here? That way it's a bit easier to follow and search in the code. Maybe `EXECUTE_CODE_FROM_PLAYGROUND_REQUEST` and `EXECUTE_CODE_FROM_PLAYGROUND_RESPONSE`? 
Maybe we can have a separate enum called `WorkerCommands` to help distinguish the two?

src/test/suite/language/mongoDBService.test.ts
```diff
@@ -2441,22 +2960,31 @@ suite('MongoDBService Test Suite', () => {
...
+
       const result = await testMongoDBService.evaluate(
         {
           connectionId: 'pineapple',
           codeToEvaluate: 'print("Hello"); console.log(1,2,3); 42',
         },
         source.token
       );
+
+      Sinon.restore();
```
Anemy, line 2982: With `Sinon.restore()` we usually call it in `afterEach` functions to ensure that even if the test fails we restore any stubs. That way if it does fail we don't impact other tests.

src/language/mongoDBService.ts
```diff
@@ -965,6 +1060,7 @@ export default class MongoDBService {
       this._provideIdentifierObjectValueCompletionItems.bind(this, state),
       this._provideTextObjectValueCompletionItems.bind(this, state),
       this._provideCollectionSymbolCompletionItems.bind(this, state),
+      this._provideStreamProcessorSymbolCompletionItems.bind(this, state),
```
Anemy, line 1063: Kind of a broader question, would it make sense to conditionally set the stream processor completion items and only run the visitor checks when we're connected to an atlas stream? Is there any reason we want these completion items for normal connections?

src/editors/playgroundController.ts
shaketbaby: Good point. Makes sense to move the `_outputChannel` to the `LanguageServerController`. Do you think the current channel name `Playground output` is still valid?

src/language/mongoDBService.ts
shaketbaby: Good call out, I'll add a new `WorkerCommands` enum.

src/language/mongoDBService.ts
shaketbaby: I think generally we don't disable things in shell. For example, `db` can still be used even if we are connected to an Atlas Stream Processing and `sp` is also there when connected to a normal Mongo Cluster. This will be an empty array so shouldn't have any effect.

src/language/mongoDBService.ts
shaketbaby: Have opted to introduce a new server command `CODE_EXECUTE_RESULT` for sending the result from worker. As we will still need 2 different commands with a new `WorkerCommands` enum. I didn't change `EXECUTE_CODE_FROM_PLAYGROUND` to minimize the changes.

src/editors/playgroundController.ts
shaketbaby: Have moved output channel to `LanguageServerController`.

Anemy: Looking at how folks see this:
Screenshot 2024-01-03 at 1 15 08 PM

Maybe we should make it named something MongoDB specific, maybe `MongoDB Language Server Output`?
Or just `MongoDB Output` to keep it simple?
also fine as is, feel free to ignore.

Anemy (approved):
nice!

PR #634, opened 2023-12-29
chore(ci): move draft releases to github actions

mcasimir:
A proposal to move also this one to CI, and be fully managed via Github actions, the set of privileges is the same of pushing tags, this is more practical and accessible for people not actively working on vscode (ie. first responder, a PM @GaurabAryal :))

The idea is to kickoff draft releases from https://github.com/mongodb-js/vscode/actions selecting the type of bump (patch, minor, major) from a dropdown or the exact new version.

The workflow has safe-guard checks for pre-existing tags and version format.

NOTE: To do this i've extracted the test and build steps in a [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action), which is called by test and set and create draft release.


Anemy (commented):
Sweet, nice changes. One thing on documentation we probably want to look at:
In our `CONTRIBUTING.md` document in this repo we write about the release process. With these changes we should update that releasing section so that its not out of date.

https://github.com/mongodb-js/vscode/blob/main/CONTRIBUTING.md#releasing

Anemy (approved):

PR #645, opened 2024-01-05
OIDC e2e tests

himanshusinghs:
Draft version
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


.depcheckrc
```diff
@@ -23,3 +23,10 @@ ignores:
  - style-loader
  - ts-loader
  - mongodb-runner
+ - "@wdio/local-runner"
+ - "@wdio/mocha-framework"
+ - "@wdio/spec-reporter"
+ - wdio-vscode-service
+ - wdio-wait-for
+ - "@wdio/globals"
+ - "@wdio/types"
```
addaleax, line 32: In Compass and mongosh our depcheckrcs often have comments about _why_ a specific dependency is in the ignore list, this might be an opportunity to also introduce this pattern here 🙂

src/test/e2e/oidc.e2e.ts
```diff
@@ -0,0 +1,150 @@
...
+  let oidcMockProviderEndpointAccesses: Record<string, number>;
+
+  let i = 0;
+  let tmpdir: string;
+  let cluster: MongoCluster;
+  let connectionString: string;
+
+  before(async function () {
+    if (process.platform !== 'linux') {
+      // return;
```
addaleax, line 68: ^

just so we don't forget

src/test/e2e/oidc.e2e.ts
```diff
@@ -0,0 +1,150 @@
...
+    await connectionStringInput.setText(connectionString);
+    await connectionStringInput.confirm();
+    await browser.waitUntil(invisibilityOf(connectionStringInput.elem));
+
+    // Check if we have connection successful notification
+    await browser.waitUntil(
+      async () => {
+        const notifs = await workbench.getNotifications();
+        const messages = await Promise.all(notifs.map((n) => n.getMessage()));
+        return messages.includes('MongoDB connection successful.');
```
addaleax, line 143: Is there some easy way of verifying that we connected as an authenticated user (and in particular that the connection attempt did not accidentally succeed because it is fully unauthenticated and makes use of the MongoDB localhost exception)?

In Compass and mongosh we use `connectionStatus` and/or check that the OIDC IDP has actually issued a token by looking at the contents of `oidcMockProviderEndpointAccesses`, those might be good things to try here

himanshusinghs: yea in addition to what is suggested I also wanted to include the authInfo check which unfortunately due to limitations being posed right now, I am unable to. Will be looking into this today but thank you for this ⭐

.depcheckrc
himanshusinghs: Makes sense - I will make sure to do that.

PR #652, opened 2024-01-09
chore: adds tests for oidc in vscode - VSCODE-414

himanshusinghs:
## Description
In this PR we have added tests for OIDC feature that covers all the applicable cases covered by compass as well as a few scenarios specific to vscode.

These are not E2E tests unfortunately, that is still an exploratory work in this [PR](https://github.com/mongodb-js/vscode/pull/645)

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


himanshusinghs:
Converting to draft since we might need to move the CI to evergreen. I will be looking into that.

src/test/suite/oidc.test.ts
```diff
@@ -0,0 +1,417 @@
...
+    afterReauth = true;
+
+    // Wait for auth to expire
+    await new Promise((resolve) => setTimeout(resolve, 1100));
+
+    // Trigger a command on data service for reauthentication
+    await testConnectionController.getActiveDataService()?.count('x.y', {});
+
+    // Minor pause
+    await new Promise((resolve) => setTimeout(resolve, 500));
```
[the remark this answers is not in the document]
himanshusinghs: Just addressed this in my recent commit.

src/test/suite/oidc.test.ts
```diff
@@ -0,0 +1,431 @@
+import os from 'os';
+import path from 'path';
+import assert from 'assert';
```
Anemy, line 3: For new tests can we use `expect` from chai? We'd like to move to expect eventually (assert is used pretty much everywhere right now though).

src/test/suite/oidc.test.ts
```diff
@@ -0,0 +1,431 @@
...
+        oidcMockProviderEndpointAccesses[pathname] ??= 0;
+        oidcMockProviderEndpointAccesses[pathname]++;
+        return overrideRequestHandler?.(url, req, res);
+      },
+    };
+    oidcMockProvider = await OIDCMockProvider.create(oidcMockProviderConfig);
+
+    tmpdir = path.join(
+      os.tmpdir(),
+      `vscode-oidc-${Date.now().toString(32)}-${++i}`
```
Anemy, line 110: Since this is in the `before` function isn't `i` always going to be 1 then?

src/connectionController.ts
```diff
@@ -425,6 +430,7 @@ export default class ConnectionController {
       );
 
     if (removeConfirmationResponse !== 'Confirm') {
+      await this.disconnect();
```
Anemy, line 433: If we don't disconnect here, does it have implications? In Compass we don't disconnect when a user declines re-authentication. I think it's fine either way, mostly curious on the reason for the change.

src/test/suite/oidc.test.ts
```diff
@@ -0,0 +1,431 @@
...
+
+    const connected =
+      await testConnectionController.addNewConnectionStringAndConnect(
+        connectionString
+      );
+    assert.strictEqual(connected, true);
+    afterReauth = true;
+
+    // Wait for auth to expire, our auth expire in 1 second
+    await new Promise((resolve) => setTimeout(resolve, 1100));
```
Anemy, line 350: Can we make the 1 second less? Thinking how to make these tests faster. A bit to what Maurizio already mentioned, something we've been trying to move away from in our tests (VSCode unfortunately has a lot of them) is using setTimeout's to delay before a test condition. We prefer finding ways to use promise based listeners so things can resolve the moment they happen. Fine as is too.

Anemy (commented):
Couple smaller questions / thoughts. Nice tests!

src/test/suite/oidc.test.ts
himanshusinghs: Yea I actually wanted to remove this as well but couldn't figure out a straight way to do that because here we are waiting for the OIDC token to expire which is marked to expire in a second. Maybe I will try to do some loops until an event is emitted or something like that. I will update here.

src/connectionController.ts
himanshusinghs: So this is a result of my test failing [here](https://github.com/mongodb-js/vscode/pull/652/files#diff-507deef688826dd8859c67e31edf5e8e1ef8dcfa58ebd3718b3e8ffca85e63e8R397). I was expecting connectionController to be disconnected after a declined re-authentication. Later I checked this in the UI as well, without this the extension works as if we are still connected but then all the actions that you do like listing collections, etc just fails so it seemed like a reasonable thing to do. But I am open to doing it like compass as well, no strong preference here.

src/test/suite/oidc.test.ts
himanshusinghs: Yea, thank you I will update it.

src/test/suite/oidc.test.ts
himanshusinghs: The only reason it used assert here was to keep it consistent with the rest of the tests. Happy to move this test to expect since we will do the move to chai and mocha, soon-ish.

src/test/suite/oidc.test.ts
himanshusinghs: Removed these in my recent commits

src/test/suite/oidc.test.ts
himanshusinghs: Addressed this as well in my recent commit

Anemy (approved):

PR #655, opened 2024-01-11
feat(connections): add edit connection context menu action VSCODE-406

Anemy:
VSCODE-406

Screenshot 2024-01-11 at 3 00 25 AM


src/views/webviewController.ts
```diff
@@ -242,7 +254,30 @@ export default class WebviewController {
...
+      id: string;
+      name?: string;
+      connectionOptions: ConnectionOptions;
+    };
+    context: vscode.ExtensionContext;
+  }) => {
+    const webviewPanel = this.openWebview(context);
+
+    // Wait for the panel to open.
+    await new Promise((resolve) => setTimeout(resolve, 200));
```
himanshusinghs, line 271: I wonder if we could avoid this. One quick thought is to listen to one of our own created event on webview to mark the webview as ready or maybe vscode api provides something?

src/views/webview-app/vscode-api.ts
```diff
@@ -11,6 +11,17 @@ interface VSCodeApi {
 declare const acquireVsCodeApi: () => VSCodeApi;
 const vscode = acquireVsCodeApi();
 
+export const sendEditConnectionToExtension = (
```
himanshusinghs, line 14: I don't think we used this anywhere in this PR, a leftover probably?

src/test/suite/connectionController.test.ts
```diff
@@ -843,43 +905,26 @@ suite('Connection Controller Test Suite', function () {
       StorageVariables.GLOBAL_SAVED_CONNECTIONS
     );
 
-    assert.strictEqual(
-      !!workspaceStoreConnections,
-      true,
-      `Expected workspace store to have connections, found ${workspaceStoreConnections}`
-    );
-
+    expect(workspaceStoreConnections).to.not.be.empty;
```
himanshusinghs, line 908: How do you feel about keeping assert for consistency sake (as long as we don't move to chai+mocha)? No strong feelings about it but thought I should call it out.

himanshusinghs: Or maybe we can leave it as is - I just saw you moved the align workflows ticket into "Ready for work" so that might cover it?

himanshusinghs (approved):
Looking amazing! I just posted a small thing and three questions to clarify.

himanshusinghs:
Ahh forgot to ask - I saw that we are removing the option for users to ask where to save connections, that is intentional right?

Anemy:
> Ahh forgot to ask - I saw that we are removing the option for users to ask where to save connections, that is intentional right?

Yes, that's intentional. Synced with Gaurab on it. The setting currently has a bug in it and we're happy to see it gone.

src/test/suite/connectionController.test.ts
Anemy: I'd like to update all of the assert of this file over to expect in the soon, so I'm thinking the lack of consistency is alright for now.

src/mdbExtensionController.ts
```diff
@@ -156,12 +156,14 @@ export default class MDBExtensionController implements vscode.Disposable {
...
     // ------ CONNECTION ------ //
-    this.registerCommand(EXTENSION_COMMANDS.MDB_OPEN_OVERVIEW_PAGE, () =>
-      this._webviewController.openWebview(this._context)
-    );
-    this.registerCommand(EXTENSION_COMMANDS.MDB_CONNECT, () =>
-      this._webviewController.openWebview(this._context)
-    );
+    this.registerCommand(EXTENSION_COMMANDS.MDB_OPEN_OVERVIEW_PAGE, () => {
+      this._webviewController.openWebview(this._context);
+      return Promise.resolve(true);
```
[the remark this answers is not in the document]
Anemy: Right now we have the `registerCommand` interface expect a boolean always. I don't think we need to be so prescriptive there and it would be good to get rid of that strict extra typing. I'm thinking it would be better in another refactor pr though.

src/views/webviewController.ts
Anemy: We definitely can, I like that idea. I'm going to merge in these changes and then open a follow up pr. As the diff in this pr is fairly large I want to be sure it's not going to cause large merge conflicts and block any other refactor work.

src/views/webview-app/use-connection-form.ts
```diff
@@ -1,65 +1,214 @@
...
-      setConnectionErrorMessage('');
+      dispatch({
+        type: 'close-connection-form',
+      });
     },
     handleCancelConnectClicked: () => {
       sendCancelConnectToExtension();
     },
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    handleSaveConnectionClicked: (connectionAttempt: {
```
Anemy, line 191: I'll make a connection-form pr and make this optional so we don't need this here anymore.

PR #658, opened 2024-01-14
feat(connect-form): add OIDC device auth flow with preference VSCODE-503

Anemy:
VSCODE-503

Screenshot 2024-01-14 at 6 33 05 PM


himanshusinghs (approved):
🚀 Should we also think about having a button to copy url and auth code in the notification message?

Anemy:
>  Should we also think about having a button to copy url and auth code in the notification message?

We could, I'll defer that to another pr. This is a not-recommended backup auth flow so I don't think we need to spend too much time making it super user friendly.

PR #667, opened 2024-01-17
chore: cancel connection attempt when removing connection, update assert to expect in webview and connection controller tests

Anemy:
Also removed some `setTimeout` usage in our tests so they're a bit quicker.


Anemy:
Going to merge in, feel free to revert if anything comes up.

PR #668, opened 2024-01-17
chore: replace jest with mocha VSCODE-516

mcasimir:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/test/suite/views/webview-app/overview-page.test.tsx
```diff
@@ -23,59 +25,57 @@ describe('OverviewPage test suite', function () {
...
+    it('is able to open and close the new connection form', async function () {
       render(<OverviewPage />);
 
       expect(screen.queryByTestId(connectionFormTestId)).to.not.exist;
-      const postMessageSpy = Sinon.spy(vscode, 'postMessage');
+      const postMessageSpy = sinon.spy(vscode, 'postMessage');
       expect(postMessageSpy).to.not.be.called;
 
-      screen.getByText('Open form').click();
+      await userEvent.click(screen.getByTestId('open-connection-form-button'));
```
Anemy, line 47: Opening the connection form takes ~4 seconds for all of the animations to complete. Going to see if this is something we can fix on the connection form side. There's also an issue with focus-trap not finding the fallback node. It's likely fixed in newer versions, which we'll get with newer ConfirmationModals so I'm going to try that as well.

If the delay can't go away then we can skip the animation waiting  (await -> void) and just insert a quick fake delay here with a comment.

Anemy: Used the `tabbable` register jsdom fix/workaround we also use in Compass. Still getting that slow connection form rendering. Haven't dug into it yet. It does slow down tests which is a bit painful. I don't notice it when running the extension though.

Anemy:
Updated the tests to all run and the config to pass. This did increase the bundle size by ~2MB though which is substantial (do not merge).
I don't think it's react causing it. It could be we're now packing some testing libraries by accident, or it's the ejson-shell-parser add.

Anemy:
| main | this branch |
| - | - |
| Screenshot 2024-08-09 at 1 56 36 PM | Screenshot 2024-08-09 at 1 18 05 PM | 

Slightly unrelated, noticed we're bundling a ton of frontend deps in the main extension bundle we don't need. Created https://jira.mongodb.org/browse/COMPASS-8154 for that.

Anemy:
Looks like merging from main fixed it. Should be good to go now.

addaleax (approved):

package.json
```diff
@@ -1076,6 +1076,7 @@
     }
   },
   "dependencies": {
+    "@babel/core": "^7.24.6",
```
alenakhineika, line 1079: Do we use this dependency somewhere?

src/test/suite/views/webview-app/overview-page.test.tsx
```diff
@@ -12,9 +13,10 @@ const connectionFormTestId = 'connection-form-modal';
 describe('OverviewPage test suite', function () {
   afterEach(() => {
     cleanup();
-    Sinon.restore();
+    sinon.restore();
```
alenakhineika, line 16: nit: I think everywhere in codebase, we stick to sandbox. e.g. `sandbox.restore()`, `sandbox.spy()`.

PR #672, opened 2024-01-23
chore: set browser auth without github actions for mongosh int tests

Anemy:
Hopefully this fixes the failing evergreen CI mongosh tests that run the VSCode tests. 

The browser command wouldn't be set in those tests as we set them in our actions. This updates the test to not use the environment variable so we can run the OIDC tests without the github env test setup. If this doesn't work because we're not using a deployment that supports OIDC in mongosh evergreen I might make a separate pr that disables these tests when not run in github actions.

### DIFF
--- a/src/test/suite/oidc.test.ts
+++ b/src/test/suite/oidc.test.ts
@@ -32,7 +32,7 @@ function hash(input: string): string {
 
 // Need to be provided via CI env because we can't get a hold for node.js exec
 // path in our tests - they run inside a vscode process in the built dir.
-const browserShellCommand = `echo "$(which node) ${__dirname}/../../../src/test/fixture/curl.js"`;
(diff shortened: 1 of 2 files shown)


src/test/suite/oidc.test.ts
```diff
@@ -31,8 +31,8 @@ function hash(input: string): string {
 }
 
 // Need to be provided via CI env because we can't get a hold for node.js exec
-// path in our tests - they run inside a vscode process
-const browserShellCommand = process.env.BROWSER_AUTH_COMMAND;
+// path in our tests - they run inside a vscode process in the built dir.
+const browserShellCommand = `$(echo "$(which node) ${__dirname}/../../../src/test/fixture/curl.js")`;
```
himanshusinghs, line 35:
```suggestion
const browserShellCommand = `$(echo "$(which node) ${__dirname}/../fixture/curl.js")`;
```

himanshusinghs (approved):

PR #677, opened 2024-01-25
fix: include export mode when retrieving import statements VSCODE-440

paula-stacho:
## Description

The `getImports` transpiler method expects two arguments
- `mode: 'Pipeline' | 'Query' | 'Delete Query' | 'Update Query'`
- `driverSyntax: boolean`

source: https://github.com/mongodb-js/compass/blob/92f267e369d188c5f11c2450f3d65f73725558e4/packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js#L94

In VSCode we have only two modes:
- when an array is selected -> aggregate ~ Pipeline
- when an object is selected -> find ~ Query

The main fix is sending two arguments (which allows for the second to be read correctly).

As far as I can tell, the exact mode only matters for Java. Given that the method applied is a wild guess, user might need to adjust the import together with the code anyway, but at least the imports now match the code.

### Steps to reproduce:
- select an object/array in the playground
- export to Java
- include Driver Syntax and Import Statements

### Current behaviour:
The import statements are only
```
import org.bson.Document;
``` 

### After the fix:
```
import org.bson.Document;
import com.mongodb.MongoClient;
import com.mongodb.MongoClientURI;
import com.mongodb.client.MongoCollection;
import com.mongodb.client.MongoDatabase;
import org.bson.conversions.Bson;
import java.util.concurrent.TimeUnit;
import org.bson.Document;
```

If you've selected an array, you'll also see -
```
import com.mongodb.client.AggregateIterable;
```
And for an object -
```
import com.mongodb.client.FindIterable;
```

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions
I feel weird about the test placement, but this was the best I found.
I made an attempt to at least separate the test into three smaller pieces. I thought it would work since mocha runs the tests sequentially, but it didn't work and I gave up 😅 . I haven't worked with mocha for a long time, so maybe I'm missing something

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/test/suite/editors/playgroundSelectedCodeActionProvider.test.ts
+++ b/src/test/suite/editors/playgroundSelectedCodeActionProvider.test.ts
@@ -195,51 +195,80 @@ suite('Playground Selected CodeAction Provider Test Suite', function () {
       }
     });
 
-    test('exports to java and includes builders and import statements', async () => {
-      const textFromEditor = "{ name: '22' }";
-      const selection = {
-        start: { line: 0, character: 0 },
-        end: { line: 0, character: 14 },
-      } as vscode.Selection;
-      const mode = ExportToLanguageMode.QUERY;
-      const activeTextEditor = {
-        document: { getText: () => textFromEditor },
-      } as vscode.TextEditor;
+    suite('exports to java', () => {
+      const expectedResult = {
+        namespace: 'DATABASE_NAME.COLLECTION_NAME',
+        type: null,
+        content: 'new Document("name", "22")',
+        language: 'java',
+      };
 
-      mdbTestExtension.testExtensionController._playgroundController._selectedText =
-        textFromEditor;
-      mdbTestExtension.testExtensionController._playgroundController._playgroundSelectedCodeActionProvider.selection =
-        selection;
-      mdbTestExtension.testExtensionController._playgroundController._playgroundSelectedCodeActionProvider.mode =
-        mode;
-      mdbTestExtension.testExtensionController._playgroundController._activeTextEditor =
-        activeTextEditor;
+      beforeEach(async () => {
+        const textFromEditor = "{ name: '22' }";
+        const selection = {
+          start: { line: 0, character: 0 },
+          end: { line: 0, character: 14 },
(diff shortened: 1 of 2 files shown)


src/test/suite/editors/playgroundSelectedCodeActionProvider.test.ts
```diff
@@ -195,7 +195,7 @@ suite('Playground Selected CodeAction Provider Test Suite', function () {
       }
     });
 
-    test('exports to java and includes builders', async () => {
+    test('exports to java and includes builders and import statements', async () => {
```
alenakhineika, line 198: Can we have actually two tests, one when we only include builders and another one for both builders and imports?

alenakhineika:
Can we add one more test? :) The `include driver syntax and then add import statements` test, which would do all the steps that the `include driver syntax (only)` does. And then call once again `await vscode.commands.executeCommand(actionCommand.command);` to emulate the case when a user runs the same playground to export to language and sees the state of the code lenses from the previous run, like `Include Import Statements | Exclude Driver Syntax | Exclude Driver Syntax`. And then we call:

```js
await vscode.commands.executeCommand(
  'mdb.changeExportToLanguageAddons',
  {
    ...mdbTestExtension.testExtensionController._playgroundController._exportToLanguageCodeLensProvider._exportToLanguageAddons,
     importStatements: true // We set a single addon here
  }
);
```

And check that the playground pane has both:

```js
expect(content).to.include(mongoClientImport);
expect(content).to.include(queryImport);
```

paula-stacho:
> Can we add one more test? :) The `include driver syntax and then add import statements` test, which would do all the steps that the `include driver syntax (only)` does. And then call once again `await vscode.commands.executeCommand(actionCommand.command);` to emulate the case when a user runs the same playground to export to language and sees the state of the code lenses from the previous run,

Ah, thanks! I knew we said five tests but I couldn't remember what the last one was

Anemy (approved):

PR #701, opened 2024-03-06
feat: show simpler uuid format VSCODE-470

paula-stacho:
https://jira.mongodb.org/browse/VSCODE-470

## Description
Adding postprocessing in the Document Service, which replaces the extended `$binary` notation with `$uuid` notation when presenting the document. Originally I thought we should reverse this on `replaceDocument`, but it looks like this is not necessary - I assume this is handled by the driver. I left the test case for regression, but perhaps it's not necessary.

Aside of the new tests, I have tested this with the `all_types` collection of `compass-data-sets/test`. I added a couple of new fields to the first document - `nested_uuid` and `nested_array_uuid`
I tried
- opening a document from the left side panel
- playground results for `findOne()`
- playground results for `find()`

### Checklist

## Motivation and Context
- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/editors/mongoDBDocumentService.ts
+++ b/src/editors/mongoDBDocumentService.ts
@@ -1,5 +1,4 @@
 import type * as vscode from 'vscode';
-import { EJSON } from 'bson';
 import type { Document } from 'bson';
 
 import type ConnectionController from '../connectionController';
@@ -9,6 +8,7 @@ import type { EditDocumentInfo } from '../types/editDocumentInfoType';
 import formatError from '../utils/formatError';
 import type { StatusView } from '../views';
 import type TelemetryService from '../telemetry/telemetryService';
+import { getEJSON } from '../utils/ejson';
 
 const log = createLogger('document controller');
 
@@ -17,9 +17,6 @@ export const DOCUMENT_ID_URI_IDENTIFIER = 'documentId';
 export const DOCUMENT_SOURCE_URI_IDENTIFIER = 'source';
 
 export const VIEW_DOCUMENT_SCHEME = 'VIEW_DOCUMENT_SCHEME';
-
-const isObject = (value: unknown) =>
-  value !== null && typeof value === 'object' && !Array.isArray(value);
 
 export default class MongoDBDocumentService {
   _context: vscode.ExtensionContext;
@@ -111,33 +108,6 @@ export default class MongoDBDocumentService {
     }
   }
 
-  _simplifyEJSON(document: Document): Document {
-    for (const [key, item] of Object.entries(document)) {
-      // UUIDs might be represented as {"$uuid": <canonical textual representation of a UUID>} in EJSON
-      // Binary subtypes 3 or 4 are used to represent UUIDs in BSON
-      // But, parsers MUST interpret the $uuid key as BSON Binary subtype 4
-      // For this reason, we are applying this representation for subtype 4 only
-      // see https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#special-rules-for-parsing-uuid-fields
-      if (
-        isObject(item) &&
-        Object.prototype.hasOwnProperty.call(item, '$binary') &&
-        item.$binary?.subType === '04' &&
-        typeof item.$binary.base64 === 'string'
-      ) {
-        const hexString = Buffer.from(item.$binary.base64, 'base64').toString(
-          'hex'
-        );
-        const match = /^(.{8})(.{4})(.{4})(.{4})(.{12})$/.exec(hexString);
-        if (!match) continue;
-        const asUUID = match.slice(1, 6).join('-');
-        document[key] = {
-          $uuid: asUUID,
-        };
-      }
-    }
-    return document;
-  }
-
   async fetchDocument(data: EditDocumentInfo): Promise<Document | void> {
     log.info('Fetch document from MongoDB', data);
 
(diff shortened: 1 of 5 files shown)


src/editors/mongoDBDocumentService.ts
```diff
@@ -108,6 +111,32 @@ export default class MongoDBDocumentService {
...
+    for (const [key, item] of Object.entries(document)) {
+      // UUIDs might be represented as {"$uuid": <canonical textual representation of a UUID>} in EJSON
+      // Binary subtypes 3 or 4 are used to represent UUIDs in BSON
+      // But, parsers MUST interpret the $uuid key as BSON Binary subtype 4
+      // For this reason, we are applying this representation for subtype 4 only
+      // see https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#special-rules-for-parsing-uuid-fields
+      if (
+        isObject(item) &&
+        item.hasOwnProperty('$binary') &&
+        item.$binary.subType === '04'
```
addaleax, line 124:
```suggestion
        Object.prototype.hasOwnProperty.call(item, '$binary') &&
        item.$binary?.subType === '04' &&
        typeof item.$binary.base64 === 'string'
```

You'll almost [never](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#using_hasownproperty_as_a_property_name) want to use `hasOwnProperty` directly, and you'll want to account for objects like `{ $binary: null }` or `{ $binary: { subType: '04', base64: null } }` because, unfortunately, any of these can be the result of EJSON output (e.g. if they were literal parts of the input)

src/editors/mongoDBDocumentService.ts
```diff
@@ -147,7 +176,8 @@ export default class MongoDBDocumentService {
         return;
       }
 
-      return JSON.parse(EJSON.stringify(documents[0]));
+      const ejson = JSON.parse(EJSON.stringify(documents[0]));
+      return this._simplifyEJSON(ejson);
```
addaleax, line 180: What about a document that is like `{ nested: { property: UUID(...) } }`?

src/editors/mongoDBDocumentService.ts
paula-stacho: ooh, good to know!

src/editors/mongoDBDocumentService.ts
paula-stacho: This is one thing that I wanted to ask about actually. Can there be nested UUIDs, and can they be nested to any level?

addaleax: Yes, and yes :)

src/utils/ejson.ts
```diff
@@ -0,0 +1,46 @@
+import { EJSON } from 'bson';
+import type { Document } from 'bson';
+
+const isObject = (value: unknown) =>
+  value !== null && typeof value === 'object';
+
+function simplifyEJSON(documents: Document[] | Document): Document {
```
alenakhineika, line 7: also nit: since it is either one or many documents, we could call the argument `item` or something, but `documents` confuses when reading code, because it can be actually one document that is not an array.

src/utils/ejson.ts
```diff
@@ -0,0 +1,46 @@
...
+    );
+  }
+
+  // UUIDs might be represented as {"$uuid": <canonical textual representation of a UUID>} in EJSON
+  // Binary subtypes 3 or 4 are used to represent UUIDs in BSON
+  // But, parsers MUST interpret the $uuid key as BSON Binary subtype 4
+  // For this reason, we are applying this representation for subtype 4 only
+  // see https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#special-rules-for-parsing-uuid-fields
+  if (
+    Object.prototype.hasOwnProperty.call(documents, '$binary') &&
```
alenakhineika, line 22: Do we need to check `hasOwnProperty` here? Isn't it enough to check `documents.$binary?.something`?

src/utils/ejson.ts
```diff
@@ -0,0 +1,46 @@
...
+    documents.$binary?.subType === '04' &&
```
alenakhineika, line 23: Have you tried what happens for the legacy UUID of subtype 3? To be honest, spec is a bit confusing. It says `Parsers MUST interpret the $uuid key as BSON Binary subtype 4`, maybe this is why we try to convert $uuid to $binary 🤔

src/utils/ejson.ts
```diff
@@ -0,0 +1,46 @@
...
+  return Object.fromEntries(
+    Object.entries(documents).map(([key, value]) => [
+      key,
+      isObject(value) ? simplifyEJSON(value) : value,
+    ])
+  );
+}
+
+export function getEJSON(documents: Document[] | Document) {
+  const ejson = JSON.parse(EJSON.stringify(documents));
```
addaleax, line 44:
```suggestion
  const ejson = EJSON.serialize(documents);
```

That's exactly what `EJSON.serialize()` does :)

src/utils/ejson.ts
```diff
@@ -0,0 +1,46 @@
+import { EJSON } from 'bson';
+import type { Document } from 'bson';
+
+const isObject = (value: unknown) =>
```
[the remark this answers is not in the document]
paula-stacho: that's fair, I will rename it!

src/utils/ejson.ts
paula-stacho: yeah, I wasn't sure about the naming on this one. I can use item if that's more clear.

src/utils/ejson.ts
paula-stacho: oh, thanks!

alenakhineika (approved):

addaleax (approved):

PR #718, opened 2024-04-11
feat(playground): add support local require VSCODE-468

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/language/worker.ts
+++ b/src/language/worker.ts
@@ -36,14 +36,25 @@ const getLanguage = (evaluationResult: EvaluationResult) => {
   return 'plaintext';
 };
 
+type ExecuteCodeOptions = {
+  codeToEvaluate: string;
+  connectionString: string;
+  connectionOptions: MongoClientOptions;
+  filePath?: string;
+};
+
 /**
  * Execute code from a playground.
  */
-const execute = async (
-  codeToEvaluate: string,
-  connectionString: string,
-  connectionOptions: MongoClientOptions
-): Promise<{ data?: ShellEvaluateResult; error?: any }> => {
+const execute = async ({
+  codeToEvaluate,
+  connectionString,
+  connectionOptions,
+  filePath,
+}: ExecuteCodeOptions): Promise<{
+  data?: ShellEvaluateResult;
+  error?: any;
+}> => {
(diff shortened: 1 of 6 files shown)


src/language/worker.ts
```diff
@@ -67,6 +78,17 @@ const execute = async (
       },
     });
 
+    // In order to support local require direclty from the file where code lives, we can not wrap the
```
addaleax, line 81:
```suggestion
    // In order to support local require directly from the file where code lives, we can not wrap the
```

src/language/worker.ts
```diff
@@ -67,6 +78,17 @@ const execute = async (
...
+    // whole code in a function for two reasons:
+    // 1. We need to return the response and can not simply add return. And
+    // 2. We can not use eval to evaluate the *codeToEvaluate* as mongosh async-rewriter can’t see into the eval.
+    // We are also not direclty concatenating the require with the code either due to "use strict"
```
addaleax, line 85:
```suggestion
    // We are also not directly concatenating the require with the code either due to "use strict"
```

src/language/worker.ts
```diff
@@ -67,6 +78,17 @@ const execute = async (
...
+    if (filePath) {
+      await runtime.evaluate(`(function () {
+        require = require('module').createRequire('${filePath}');
```
addaleax, line 88:
```suggestion
        globalThis.require = require('module').createRequire(${JSON.stringify(filePath)});
```

- Using `JSON.stringify()` is the easiest way to make sure that strings are inserted literally when programmatically creating JS code, it takes care of all necessary escaping – without it, any file path containing a `\` or `'` would break, and the former is obviously pretty common on Windows 🙂 
- We're assigning to the global object here anyway, but that might not be obvious to readers, so let's make it obvious 🙂

alenakhineika (approved):
Nice! Super handy feature and really wanted by users of the extension ✨ The code also looks good 👍

addaleax (approved):
Looks good! Can we add a test for this?

PR #734, opened 2024-05-23
fix: disconnect extension when server is closed VSCODE-536

alenakhineika:
## Description

Bump some minor dependencies and fix [VSCODE-536](https://jira.mongodb.org/browse/VSCODE-536).

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [x] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


src/connectionController.ts
```diff
@@ -458,15 +459,25 @@ export default class ConnectionController {
 
   // Used to re-authenticate with OIDC.
   async _reauthenticationHandler() {
+    console.log('----------------------');
```
Anemy, line 462: Some leftover logs?

src/test/suite/oidc.test.ts
```diff
@@ -352,7 +357,8 @@ suite('OIDC Tests', function () {
     expect(testConnectionController.isCurrentlyConnected()).to.be.true;
   });
 
-  test('can decline re-authentication if wanted', async function () {
+  // eslint-disable-next-line mocha/no-exclusive-tests
+  test.only('can decline re-authentication if wanted', async function () {
```
Anemy, line 361: .only and some logs here

Anemy:
Are we going to continue this pr re https://github.com/mongodb-js/vscode/pull/737 ?

alenakhineika:
@Anemy you probably missed it during standup, yes this PR was in the debug mode therefore there are some prints around. I will clean this and ping you when ready to take a look.

src/connectionController.ts
alenakhineika: Logs in use 😄 but these I cleared up for now, don't be surprised if you see more. Added WIP label to the pr.

src/test/suite/oidc.test.ts
alenakhineika: Same as other places - WIP.

src/connectionController.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
Anemy, line 1:
```suggestion
```

src/connectionController.ts
```diff
@@ -989,9 +982,13 @@ export default class ConnectionController {
     this._connectingConnectionId = '';
   }
 
-  // Exposed for testing.
   setActiveDataService(newDataService: DataService): void {
     this._activeDataService = newDataService;
+
+    // Disconnect the extension if the MongoDB client is closed.
+    this._activeDataService?.once('close', () => {
+      void this.disconnect();
```
Anemy, line 990: When does the `close` event occur? Doesn't it also happen when we disconnect?
Should we cleanup these listeners? Or are they already cleaned up here https://github.com/mongodb-js/compass/blob/e7b5391b6b80f5524378da2e1d3988f7da61715b/packages/data-service/src/data-service.ts#L2634

src/test/suite/connectionController.test.ts
```diff
@@ -189,32 +180,45 @@ suite('Connection Controller Test Suite', function () {
...
+        connectionsDidChangeEventFiredCount++;
       }
     );
 
-    void testConnectionController.addNewConnectionStringAndConnect(
+    await testConnectionController.addNewConnectionStringAndConnect(
       TEST_DATABASE_URI
     );
-  });
+    await sleep(100);
```
Anemy, line 200: Is there a way we can avoid sleeping an arbitrary time here? Maybe we can poll/wait for the success state?

I added a bunch of `sleep`s in tests in the past and overtime they can really slow down how fast all tests are run, and they also can introduce flakes from slow networks and things, so I've been avoiding them now.

src/test/suite/oidc.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
Anemy, line 1:
```suggestion
```

src/test/suite/oidc.test.ts
```diff
@@ -398,6 +406,7 @@ suite('OIDC Tests', function () {
     }
 
     await reAuthPromise;
+    await sleep(100);
```
Anemy, line 409: Why do we need this sleep now? Can this be down without it?

src/views/webview-app/use-connection-form.ts
```diff
@@ -145,7 +145,9 @@ export default function useConnectionForm() {
       }
     };
     window.addEventListener('message', handleConnectResultResponse);
-    () => window.removeEventListener('message', handleConnectResultResponse);
+    return () => {
+      window.removeEventListener('message', handleConnectResultResponse);
```
Anemy, line 149: Was this not working before? 🙈 
Nice catch

Anemy (commented):
Nice, left a question/small comment or two

src/connectionController.ts
alenakhineika: Correct, when data service disconnects, the `_cleanup()` helper removes all listeners.

src/test/suite/oidc.test.ts
alenakhineika: I wanted to ensure here that it is enough time to capture the closing event. Let's try without sleeps, and see how tests react.

src/test/suite/connectionController.test.ts
alenakhineika: Removed as answered below.

src/views/webview-app/use-connection-form.ts
alenakhineika: It was printing warnings in the terminal that I guess we just ignored 🙈

src/test/suite/oidc.test.ts
alenakhineika: Yeah we do need sleep here otherwise the connection Is not yet closed, but I cleaned up some other places that you have mentioned.

Anemy (approved):

src/test/suite/oidc.test.ts
Anemy: Instead of `sleep` we can use something like a `waitFor` with a timeout that keeps increasing, that way we will complete in 5ms if the end condition is already met.

PR #757, opened 2024-07-10
feat: copilot new api VSCODE-550

alenakhineika:
## Description

- Create a MongoDB participant using the latest VSCode [Language Model API](https://code.visualstudio.com/api/extension-guides/language-model)
- Create the `/query` command that generated MongoDB queries and can open them in a new playground or run directly
- Call a generic handler if the participant called without a command

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/participant/participant.ts
```diff
@@ -0,0 +1,332 @@
...
+
+    stream.button({
+      command: EXTENSION_COMMANDS.RUN_PARTICIPANT_QUERY,
+      title: vscode.l10n.t('▶️ Run'),
+    });
+
+    stream.button({
+      command: EXTENSION_COMMANDS.OPEN_PARTICIPANT_QUERY_IN_PLAYGROUND,
+      title: vscode.l10n.t('Open in playground'),
+    });
```
Anemy, line 206: Maybe not for this pr, but I think there is a chance the ai could return multiple code blocks in their response. 
Like the prompt:
> What are 3 ways I could do X

I'm thinking we probably want to add a play and open in playground button for each code block as they come. Ideally we show them inline, right after the code block closes, while we stream the response.

src/editors/playgroundController.ts
```diff
@@ -164,6 +169,7 @@ export default class PlaygroundController {
     this._exportToLanguageCodeLensProvider = exportToLanguageCodeLensProvider;
     this._playgroundSelectedCodeActionProvider =
       playgroundSelectedCodeActionProvider;
+    this._participantController = participantController;
```
Anemy, line 172: Why do we have this on the playgroundController class? Maybe we can remove it?

Anemy (approved):
lgtm, left one small question and suggestion. I'm thinking we'll be doing a good amount of iterating on prompts,  response handling, and more so not a need for now.

src/editors/playgroundController.ts
alenakhineika: Planned for the feature usage, but I can remove it for now. As you wish. I just fixed tests and didn't want to revert this work 🙈

src/participant/participant.ts
alenakhineika: Good point. Let's address this in further PRs 👍

src/editors/playgroundController.ts
Anemy: All good ! No worries!

PR #772, opened 2024-08-09
chore(build): externalize connection-form frontend dependencies from main extension webpack bundle

Anemy:
| before | after |
| - | - |
| Screenshot 2024-08-09 at 1 56 36 PM | Screenshot 2024-08-09 at 1 51 20 PM |

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -142,6 +142,16 @@ module.exports = (env, argv) => {
     });
 
   const extensionConfig = merge(nodeTargetConfig(), {
+    externals: {
+      // @mongodb-js/connection-form pulls in a number of large front dependencies.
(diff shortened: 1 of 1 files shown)


Anemy:
Bundle fails when installed, will fix then remove the WIP.

Anemy:
Another option would be to move the `adjustConnectionOptionsBeforeConnect` util to a separate place from connection-form. Maybe connection-info?

Anemy:
Yeah this really won't work nicely with webpack things, the way the connection form handles things is a lot. Going to create a ticket and describe the issue.

Anemy:
https://jira.mongodb.org/browse/COMPASS-8154

PR #774, opened 2024-08-12
chore: align precommit with devtools

Anemy:
I found the precommit that existed in this repo would slow things down, so I figured we should use the same one as Compass/mongosh/devtools.
I'd oftentimes try to commit, there's a formatting issue, it would block the commit, I'd run reformat, forget to add the staged files, commit the old changes (the check would pass), then realize after pushing up that I hadn't added the files.

This pr aligns the pre-commit with other devtools projects.
Uses https://github.com/mongodb-js/devtools-shared/blob/main/packages/monorepo-tools/src/precommit.ts


addaleax (approved):

alenakhineika:
To be honest I liked this formatting pre-commit check. I find it more annoying when you forget to run prettier and then see it only when ci fails.

PR #777, opened 2024-08-13
chore(chat-participant): refactor prompt locations

Anemy:
This is so we can test and iterate on them a bit easier.

### DIFF
--- a/src/participant/constants.ts
+++ b/src/participant/constants.ts
@@ -1,0 +1,1 @@
+export const CHAT_PARTICIPANT_ID = 'mongodb.participant';
--- a/src/participant/prompts/query.ts
+++ b/src/participant/prompts/query.ts
@@ -1,0 +1,41 @@
+import * as vscode from 'vscode';
+
+import { getHistoryMessages } from './history';
+
+export class QueryPrompt {
+  static getSystemPrompt(): vscode.LanguageModelChatMessage {
+    const prompt = `You are a MongoDB expert.
+Your task is to help the user craft MongoDB queries and aggregation pipelines that perform their task.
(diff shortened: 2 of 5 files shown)


src/participant/constants.ts
```diff
@@ -0,0 +1 @@
+export const CHAT_PARTICIPANT_ID = 'mongodb.participant';
```
alenakhineika, line 1: Curious, why would you want to store it separately? How is it different from e.g. CHAT_PARTICIPANT_MODEL or other constants that we have around?

Anemy: Only moved it separate to avoid a circular dependency, it's used in two places. I imagine more things will move here over time.

src/participant/prompts/query.ts
```diff
@@ -0,0 +1,41 @@
...
+  static buildMessages({
+    context,
+    request,
+  }: {
+    request: vscode.ChatRequest;
+    context: vscode.ChatContext;
+  }): vscode.LanguageModelChatMessage[] {
+    const messages = [
+      QueryPrompt.getSystemPrompt(),
+      ...getHistoryMessages({ context }),
```
alenakhineika, line 35: It looks like we don't need to attach the history explicitly to the message. It remembers the conversations within the current open chat.

alenakhineika: Hmm looks like it can be a bug on the copilot side, because according to documentation the history shouldn't be added automatically https://code.visualstudio.com/api/extension-guides/chat#use-the-chat-message-history

alenakhineika (approved):

Issue #379, opened 2021-12-23
Feature Request: Auto Hide Notitications After 5 - 15 seconds.


elibroftw:
## Feature Request

Title

## Detailed Description
![image](https://user-images.githubusercontent.com/21298211/147186243-1b30bedf-880f-4902-83e7-f98d82351b66.png)
Just another extension that makes me click the x on its constant notifications...

## Context

Improves productivity

Anemy:
Hi @elibroftw - Currently VSCode doesn't provide us a way to auto hide notifications after a certain amount of time - https://github.com/Microsoft/vscode/issues/2732

Thanks for bringing this up, it makes sense that these notifications would become cumbersome after you're familiar with the extension.  I think we could revisit if these notifications are really necessary and possibly remove them completely, or we could provide a setting in our extension's settings to disable these notifications. Do you have any preferences? I'll ask the team as well.

elibroftw:
I think you should start with an option to disable successful notifications.

SethFalco:
> or we could provide a setting in our extension's settings to disable these notifications.

I'd favor this option too! I'd be happy to implement it myself later today and see what you all think.

Really, VSC should just have a toast service so we can show toasts to user. (i.e. messages with a timeout that require no particular action)

They seem to be keen on using the status bar for things like that, though.

I guess in their vision, this extension would use the status bar with "Not Connected" or "Connected to {databaseName}".

Perhaps this extension should also contribute a status bar item like this as well?
It can have states like the following:

* Not Connected
* Connecting to {databaseName}
* Connected to {databaseName}

PR #778, opened 2024-08-14
feat: identify a namespace to place in generated query VSCODE-579

alenakhineika:
## Description

- Select a connection from the participant [VSCODE-577](https://jira.mongodb.org/browse/VSCODE-577)
- Select a namespace from the participant [VSCODE-579](https://jira.mongodb.org/browse/VSCODE-579)
- Replace connected/disconnected messages in modals with status view messages [#379](https://github.com/mongodb-js/vscode/issues/379)
- Add the allowed Shell commands to the query prompt [VSCODE-576](https://jira.mongodb.org/browse/VSCODE-576) (Do not close the ticket, more work needs to be done)

The alternative solution to connect via the button in the chat response that opens the command palette was explored [here](https://github.com/mongodb-js/vscode/pull/776).

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/participant/prompts/query.ts
```diff
@@ -3,14 +3,47 @@ import * as vscode from 'vscode';
...
-    const prompt = `You are a MongoDB expert.
-Your task is to help the user craft MongoDB queries and aggregation pipelines that perform their task.
-Keep your response concise.
-You should suggest queries that are performant and correct.
-Respond with markdown, suggest code in a Markdown code block that begins with \'\'\'javascript and ends with \`\`\`.
-You can imagine the schema, collection, and database name.
-Respond in MongoDB shell syntax using the \'\'\'javascript code block syntax.`;
+  static getSystemPrompt({
+    databaseName = 'mongodbVSCodeCopilotDB',
+    collectionName = 'results',
```
Anemy, line 8: How should we decide on these? Maybe `test` db with `mongoDBVSCodeCopilot` as the collection instead?

Anemy (commented):
Couple small things I noticed, let's sync online

src/connectionController.ts
```diff
@@ -603,7 +603,7 @@ export default class ConnectionController {
       'mdb.isAtlasStreams',
       false
     );
-    void vscode.window.showInformationMessage('MongoDB disconnected.');
+    this._statusView.showMessage('MongoDB disconnected.');
 
     this._disconnecting = false;
     this._statusView.hideMessage();
```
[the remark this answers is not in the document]
alenakhineika: Nice catch! I think we can just remove the hiding since there is not so much you can do when disconnected. And we replace it with the connected status when there is an active connection.

src/participant/prompts/query.ts
alenakhineika: In playgrounds, we use `mongodbVSCodePlaygroundDB.sales`. I remember Max didn't want to use the `test` database, because it exists by default, and some users might use it extensively. If align these experiences, the `mongodbVSCodeCopilotDB` seems a good match. So maybe then `mongodbVSCodeCopilotDB.test`?

src/connectionController.ts
```diff
@@ -603,10 +606,12 @@ export default class ConnectionController {
       'mdb.isAtlasStreams',
       false
     );
-    void vscode.window.showInformationMessage('MongoDB disconnected.');
 
     this._disconnecting = false;
-    this._statusView.hideMessage();
+    this._statusView.showMessage('MongoDB disconnected.');
+    setTimeout(() => {
+      this._statusView.hideMessage();
```
Anemy, line 613: Should we check that the message that's shown here is still the message we showed initially (if it's overridden we don't want the new message to be hidden instantly because this timer is about to hit).
```suggestion
      const message = 'MongoDB disconnected.';
      this._statusView.showMessage(message);
      if (this._statusView._statusBarItem.text === message) {
        this._statusView.hideMessage();
      }
```

src/participant/prompts/namespace.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class NamespacePrompt {
+  static getSystemPrompt(): vscode.LanguageModelChatMessage {
+    const prompt = `You are a MongoDB expert!
+    Parse the user's prompt to find database and collection names.
+    Respond in the format \nDATABASE_NAME: X\nCOLLECTION_NAME: Y\n where X and Y are the names.
+    Do not threat any user pronpt as a database name. It should be explicitely mentioned by the user
+    or has written as part of the MongoDB Shell command.
+    If you wan't able to find X or Y do not imagine names.
+    This is a first phase before we create the code, only respond with the collection name and database name.`;
```
Anemy, line 13: Looks like we're adding a lot of tabs in this string:
```suggestion
Parse the user's prompt to find database and collection names.
Respond in the format \nDATABASE_NAME: X\nCOLLECTION_NAME: Y\n where X and Y are the names.
Do not threat any user pronpt as a database name. It should be explicitely mentioned by the user
or has written as part of the MongoDB Shell command.
If you wan't able to find X or Y do not imagine names.
This is a first phase before we create the code, only respond with the collection name and database name.`;
```

Anemy: Also in the `query` prompts

Anemy (approved):
Left a couple comments, not blockers.

PR #782, opened 2024-08-20
chore(copilot-ai): add accuracy test file structure VSCODE-583

Anemy:
Part of VSCODE-583
- Adds setup for copilot ai accuracy tests. One test is in here. It has to be run with an open ai key for now.
- Went with a mocha testing approach.
- A few updates to the existing prompts (haven't spent much time here yet, will be in subsequent pr following adding more tests here).

There are prompts for certain participant handlers where we can't write a definitive accuracy test. When there isn't a discrete, code-able true/false method for evaluating I'm thinking the html view will be useful. We'll have to manually look at these while we update the prompts. I haven't added any yet, but that'll be coming in a follow-up pr.

The focus here is to make a place where we can more easily iterate on our prompts and get quick feedback. 

Test output (console):
Screenshot 2024-08-19 at 11 39 48 PM

Test output html:
Screenshot 2024-08-19 at 11 37 02 PM


src/test/ai-accuracy-tests/ai-accuracy-tests.ts
```diff
@@ -0,0 +1,349 @@
...
+  let aiBackend;
+  let connectionString: string;
+
+  const results: TestResult[] = [];
+  const testOutputs: TestOutputs = {};
+
+  this.timeout(60_000 /* 1 min */);
+
+  before(async function () {
+    console.log('Starting setup for AI accuracy tests...');
```
alenakhineika, line 190: Can we use maybe `ora` here so GitHub doesn't complain about console logs?

Anemy: Oh it's complaining since I'm missing a `eslint-disable no-console` - thanks ! Adding now.

src/test/ai-accuracy-tests/ai-backend.ts
```diff
@@ -0,0 +1,158 @@
...
+import type { TextBlock } from '@anthropic-ai/sdk/resources';
+
+import { CHAT_PARTICIPANT_MODEL } from '../../participant/constants';
+
+let anthropic: Anthropic;
+if (process.env.ANTHROPIC_API_KEY) {
+  anthropic = new Anthropic({
+    apiKey: process.env.ANTHROPIC_API_KEY,
+  });
+}
```
alenakhineika, line 13: Do we need lines 8-13 when it looks like `getAnthropicClient()` does the same?

src/test/ai-accuracy-tests/ai-backend.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  return anthropic;
+}
+
+let openai: OpenAI;
+if (process.env.OPENAI_API_KEY) {
+  openai = new OpenAI({
+    apiKey: process.env.OPENAI_API_KEY,
+  });
+}
```
alenakhineika, line 30: Same here, or am i missing how this works?

src/test/ai-accuracy-tests/ai-backend.ts
Anemy: Ah yeah we don't need those, removed. Thanks

package.json
```diff
@@ -1186,6 +1187,7 @@
     "vscode-languageserver-textdocument": "^1.0.12"
   },
   "devDependencies": {
+    "@anthropic-ai/sdk": "^0.25.2",
```
alenakhineika, line 1190: Do we want to support both `@anthropic-ai/sdk` and `openai`? It adds extra dependencies to the project and we are not even planning to run these tests nightly or on schedule. How do you feel about keeping just `openai`, since this is something that we also use in compass?

src/test/ai-accuracy-tests/ai-accuracy-tests.ts
```diff
@@ -0,0 +1,350 @@
...
+
+    await mongoClient?.close();
+    await cluster?.close();
+
+    const htmlPageLocation = await createTestResultsHTMLPage({
+      testResults: results,
+      testOutputs,
+    });
+    console.log('View prompts and responses here:');
+    console.log(htmlPageLocation);
```
alenakhineika, line 247: 🔥

package.json
Anemy: Sounds good to me, I'll remove anthropic

alenakhineika (approved):

PR #783, opened 2024-08-20
test: cover participant with unit tests VSCODE-592

alenakhineika:
## Description

- Cover the rest of the participant controller with tests
- Along the way fix the handling of empty prompts

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/test/suite/participant/participant.test.ts
+++ b/src/test/suite/participant/participant.test.ts
@@ -48,20 +48,6 @@ suite('Participant Controller Test Suite', function () {
       connectionController: testConnectionController,
       storageController: testStorageController,
     });
-    sinon.replace(
-      testParticipantController._connectionController,
-      'getActiveDataService',
-      () =>
-        ({
-          listDatabases: () => Promise.resolve([{ name: 'dbOne' }]),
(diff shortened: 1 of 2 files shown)


src/test/suite/participant/participant.test.ts
```diff
@@ -129,8 +115,144 @@ suite('Participant Controller Test Suite', function () {
...
+        'getActiveDataService',
+        () => null
+      );
+      sinon.replace(
+        testParticipantController._storageController,
+        'get',
+        sinon.fake.returns(true)
+      );
+      sinon
+        .stub(testParticipantController._connectionController, '_connections')
```
Anemy, line 142: Not a blocker, I think if possible we should avoid stubbing the interior private aspects of another module, it can increase the friction for changes and lead to some hard to track down testing issue. Instead we can stub what the public/exposed functions return. - Or, even better, avoid stubbing all together (maybe not feasible here).

Anemy (approved):

PR #784, opened 2024-08-20
chore(testing): use ts source map in tests, improve test cleanup error handling VSCODE-593

Anemy:
VSCODE-593

Updates our VSCode mocha integration tests to show the typescript file location when there's an error (use the source map). 

Also a bit of a drive-by I updated how we run the tests to remove a generic error we'd show and also to have better cleanup. Previously if someone was running the tests and ctrl + c'd then we wouldn't run the cleanup and sometimes we'd have a zombie mongod running.

| before | after |
| - | - |
| Screenshot 2024-08-20 at 2 18 04 PM | Screenshot 2024-08-20 at 2 06 54 PM |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -105,6 +105,7 @@
         "process": "^0.11.10",
         "sinon": "^9.2.4",
         "sinon-chai": "^3.7.0",
+        "source-map-support": "^0.5.21",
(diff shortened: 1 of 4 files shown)


alenakhineika (approved):
🎊

PR #786, opened 2024-08-21
feat(playgrounds): remove export to language, bson-transpilers VSCODE-594

Anemy:
VSCODE-594

See reasoning for change in ticket.

I was working on a fix for the code lenses showing on non-playground js files VSCODE-538, it ended up needing some refactoring of the export to language usage as well. As a result I figured if we remove this first it'll make that change a bit easier.


scripts/check-vsix-size.ts
```diff
@@ -12,7 +12,7 @@ const vsixFileName = path.resolve(
 );
 const size = fs.statSync(vsixFileName).size;
 
-const maxSize = 8_000_000;
+const maxSize = 7_000_000;
```
alenakhineika, line 15: 🔥

package.json
```diff
@@ -1091,7 +1019,6 @@
     "@mongosh/shell-api": "^2.2.15",
     "@segment/analytics-node": "^1.3.0",
     "bson": "^6.8.0",
-    "bson-transpilers": "^2.2.0",
```
alenakhineika, line 1094: 💀

alenakhineika:
I am unsure that we want to delete all right now. The current agreement until the design team comes up with a better alternative, we will continue to access the new export to language via code lenses. So part of the code that this pr removes still needs to be re-used. And if there is a bug with code lenses, it needs to be fixed.

Anemy:
@alenakhineika _Some_ of this code will be useful when we're implementing the new export to language functionality. However we're likely going to be providing different entry points (not the same sorts of text selections for instance), and possibly different languages. I find it's good to avoid leaving around unused code, which is why I was thinking we should remove it all now and re-add the needed parts later. We can extract the parts we need from this branch/earlier branches when we do that implementation.

Anemy:
Chatted offline, let's hold off on this.

PR #789, opened 2024-08-22
fix(editors): don't show connect code lens on regular js files VSCODE-538

Anemy:
VSCODE-538

Also included some better event listener cleanup.

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -788,6 +788,10 @@ export default class ConnectionController {
     this.eventEmitter.removeListener(eventType, listener);
   }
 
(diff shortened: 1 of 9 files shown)


alenakhineika:
PR looks good, but something introduces flaky tests. Haven't seen those failures before.

Anemy:
@alenakhineika I don't think it's from these changes. I think it's likely a result of the jest -> mocha removal. Going to keep an eye out to see if it happens on anything other than mac os. Even locally on main the connection form takes a while to render.

alenakhineika (approved):

PR #792, opened 2024-08-26
feat: send schema to the model for better results VSCODE-581

alenakhineika:
## Description

Include schema when sending requests to the copilot's model.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/participant/participant.ts
+++ b/src/participant/participant.ts
@@ -1,4 +1,5 @@
 import * as vscode from 'vscode';
+import { getSimplifiedSchema } from 'mongodb-schema';
 
 import { createLogger } from '../logging';
 import type ConnectionController from '../connectionController';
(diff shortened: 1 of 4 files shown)


src/participant/participant.ts
```diff
@@ -590,6 +615,42 @@ export default class ParticipantController {
     return true;
   }
 
+  _shouldFetchCollectionSchema(): boolean {
```
kmruiz, line 618: Hmmm this method modifies the state but the signature looks harmless. I would use a name that makes explicit there are side effects or changing the state in another method.

src/participant/schema.ts
```diff
@@ -0,0 +1,102 @@
...
+    });
+  }
+
+  getPropAsString(pProp: string): string {
+    if (pProp.match(PROPERTY_REGEX)) {
+      return pProp;
+    }
+
+    try {
+      return JSON.stringify(pProp);
```
kmruiz, line 90: Maybe using EJSON here is better, to avoid losing type info like decimal128, dates, etc.

kmruiz (approved):

src/participant/schema.ts
alenakhineika: Here we stringify only the object keys, which is safe with `JSON.stringify`.

PR #804, opened 2024-09-03
feat: improve the format of the participant answer VSCODE-582

alenakhineika:
## Description

- Use a different confirmation modal text for the Run from copilot action.
- Remove a placeholder for a database from the copilot playground template.
- Fix history so it contains values from markdown.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


Anemy (approved):
lgtm! I like that we're moving a bit more functional and keeping less things like `_codeToEvaluate` around on state. I find it makes things less coupled and easier to test/abstract.

src/editors/playgroundController.ts
```diff
@@ -579,7 +572,37 @@ export default class PlaygroundController {
...
+    const evaluateResponse: ShellEvaluateResult =
+      await this._evaluateWithCancelModal(text);
+
+    if (!evaluateResponse || !evaluateResponse.result) {
+      return false;
+    }
+
+    await this._openInResultPane(evaluateResponse.result);
+
+    return Promise.resolve(true);
```
nirinchev, line 602: [nit] I don't believe we need the `Promise.resolve` here since we're already in an async context.

```suggestion
    return true;
```

src/editors/playgroundController.ts
```diff
@@ -666,19 +686,20 @@ export default class PlaygroundController {
 
     const selections = this._activeTextEditor.selections;
 
+    let codeToEvaluate;
     if (
       !selections ||
       !Array.isArray(selections) ||
       (selections.length === 1 && this._getSelectedText(selections[0]) === '')
     ) {
```
nirinchev, line 694: I know this is code that didn't change in this PR, but it's counterintuitive to me that we need to do these evaluations here, considering we already have `this._selectedText`. I would expect that this condition is equivalent to `!this._selectedText`.

src/editors/playgroundController.ts
```diff
@@ -666,19 +686,20 @@ export default class PlaygroundController {
...
-      this._codeToEvaluate = this._getAllText();
+      codeToEvaluate = this._getAllText();
     } else if (this._selectedText) {
       this._isPartialRun = true;
-      this._codeToEvaluate = this._selectedText;
+      codeToEvaluate = this._selectedText;
     }
 
-    return this._evaluatePlayground();
+    return this._evaluatePlayground(codeToEvaluate);
```
nirinchev, line 702: [nit] With the current code structure, we may be passing `undefined` here, while `_evaluatePlayground` expects string. My suggestion would be to declare the type of `codeToEvaluate` and either return early in an `else` clause for the `if`-statement on line 690, or fallback to an empty string.

nirinchev (commented):
A few super minor comments from me - mostly on the code, as I'm not familiar enough with the project to do a deeper review yet.

nirinchev:
Ugh, didn't notice the PR was merged prior to posting my comments - they're not blocking and can be addressed as a drive-by in the future.

PR #806, opened 2024-09-05
feat: send sample documents to the model for better results VSCODE-580

alenakhineika:
## Description

Provide the passing of sample documents to the AI generations:
- Only when a setting is enabled, by default this is false.
- Only 3 sample documents are passed
   - If this passes the maximum context size, then 1 document is passed
- Only pass 3 items in arrays
- Only pass the first 20 characters in strings

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


package.json
```diff
@@ -1101,6 +1101,11 @@
           "default": true,
           "description": "Show a confirmation message before running code generated by the MongoDB participant."
         },
+        "mdb.useSampleDocsInCopilot": {
+          "type": "boolean",
+          "default": false,
+          "description": "Enable sending sample field values with the participant query command."
```
Anemy, line 1107: How do you feel about `useSampleDocsInCopilotChat` to make it mention that it's for the chat?


Also same comment as above
```suggestion
        "mdb.useSampleDocsInCopilot": {
          "type": "boolean",
          "default": false,
          "description": "Enable sending sample field values with the VSCode copilot chat @MongoDB participant /query command."
```

src/participant/sampleDocuments.ts
```diff
@@ -0,0 +1,31 @@
+import type { Document } from 'bson';
+import { toJSString } from 'mongodb-query-parser';
+
+const MAX_ARRAY_LENGTH_OF_SAMPLE_DOCUMENT_VALUE = 3;
+
+const MAX_STRING_LENGTH_OF_SAMPLE_DOCUMENT_VALUE = 20;
+
+export function formatSampleDocuments(obj: Document) {
```
nirinchev, line 8: Should the argument type be `Document[]` instead? Also, would it make sense for the return type to be `string[]` which is the truncated projection of the supplied document collection? Not sure if it's out of scope, but I was thinking we could rework our approach to take 0-4 documents instead of all or nothing. Don't know if it'll improve the accuracy of the responses, but figured it's not a lot of effort so we might as well do it.

src/participant/sampleDocuments.ts
```diff
@@ -0,0 +1,31 @@
...
+  return toJSString(format(obj));
+}
+
+function format(obj: Document) {
```
nirinchev, line 12: Not sure what the coding style of the codebase is, but should we add some type declarations for the arguments and function returns?

src/participant/participant.ts
```diff
@@ -645,8 +649,17 @@ export default class ParticipantController {
 
       const schema = await getSimplifiedSchema(sampleDocuments);
       this._schema = new SchemaFormatter().format(schema);
+
+      const useSampleDocsInCopilot = !!vscode.workspace
+        .getConfiguration('mdb')
+        .get('useSampleDocsInCopilot');
+
+      if (useSampleDocsInCopilot) {
+        this._sampleDocuments = formatSampleDocuments(sampleDocuments);
```
nirinchev, line 658: The PR description says we're passing in 3 sample docs to copilot, but here it appears we're formatting the entire sampleDocuments dataset, which is 4 docs, as far as I can tell.

nirinchev (commented):
Looks reasonable to me. I have a few nitty comments - the only meaningful suggestion is considering returning an array of formatted documents and then appending them one by one to the prompt (checking for exceeding the max prompt length before each append) rather than as a all-or-nothing kind of thing.

src/participant/participant.ts
alenakhineika: Yeah, the number should be 4. Good catch. Copy and paste from compass.

Anemy (commented):
Looks good, left a couple comments

src/participant/sampleDocuments.ts
alenakhineika: If we want to match compass behavior this would be all or nothing: https://github.com/10gen/mms/blob/master/server/src/main/com/xgen/cloud/naturallanguagequery/_public/svc/BuildQueryPromptUtils.java#L144

@Anemy maybe you can give feedback if this approach goes well so far for compass? does it worth an efforts to do 1-3?

src/editors/playgroundController.ts
```diff
@@ -684,17 +684,11 @@ export default class PlaygroundController {
...
-    const selections = this._activeTextEditor.selections;
-
-    let codeToEvaluate;
-    if (
-      !selections ||
-      !Array.isArray(selections) ||
-      (selections.length === 1 && this._getSelectedText(selections[0]) === '')
-    ) {
+    let codeToEvaluate = '';
+    if (!this._selectedText) {
```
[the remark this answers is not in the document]
alenakhineika: This change was driven by Nikola's comment in my former PR https://github.com/mongodb-js/vscode/pull/804#discussion_r1743245808. I played around and seems working well. I guess you can not select an empty string like this when it is just one line and it is an empty string. Was you able to catch this somehow?

src/participant/sampleDocuments.ts
alenakhineika: It should have a return type, but sometimes we forget :) I will add `@typescript-eslint/explicit-function-return-type` to catch those cases.

src/participant/prompts/query.ts
```diff
@@ -2,17 +2,21 @@ import * as vscode from 'vscode';
 
 import { getHistoryMessages } from './history';
 
+export const MAX_TOTAL_PROMPT_LENGTH = 10000;
```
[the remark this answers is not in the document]
alenakhineika: The number came from mms :)
https://github.com/10gen/mms/blob/master/server/src/main/com/xgen/cloud/naturallanguagequery/_public/svc/BuildQueryPromptUtils.java#L18C27-L18C50
But I will check if we can use maxInputTokens from the model, this will be more efficient if they continue to change the context window size.

src/participant/sampleDocuments.ts
Anemy: I'd like to test if the best behavior is only to supply one document for less content in the prompt to keep the ai from losing focus. For now I like reducing it to one document when 3 is too much for the context length, that seems like a good solution and shouldn't introduce much complexity.

src/editors/playgroundController.ts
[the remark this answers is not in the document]
Anemy: Nope, just calling it out, all good!

src/participant/prompts/query.ts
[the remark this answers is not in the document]
alenakhineika: I keep MAX_TOTAL_PROMPT_LENGTH as a fallback for the first prompt, and then when maxInputTokens becomes available from the model I will replace it with the real number.

src/participant/prompts/query.ts
```diff
@@ -58,11 +62,17 @@ db.getCollection('');\n\n`;
...
-      sampleDocuments &&
-      prompt.length + sampleDocuments.length < MAX_TOTAL_PROMPT_LENGTH
-    ) {
-      prompt += `Sample socuments: ${sampleDocuments}\n`;
+    if (sampleDocuments) {
+      let docs = toJSString(sampleDocuments);
+      // First check the length of all stringified sample documents.
+      // If the resulting prompt is too large, proceed with only 1 sample document.
+      if (prompt.length + docs.length > maxInputTokens) {
+        docs = toJSString([sampleDocuments[0]]);
```
Anemy, line 70: In this case should we change the prompt addition to be
`Sample document` without the `s`? Also we can remove the array wrapper `[]`.

alenakhineika: Hmm not sure it makes much difference for the model. You can experiment as part of [VSCODE-583](https://jira.mongodb.org/browse/VSCODE-583) if this gives better results.

Anemy: Simplifying the prompt is a good practice where we can, the extra [] gives more for the ai to parse and reason about.

src/participant/prompts/query.ts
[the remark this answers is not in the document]
Anemy: Ah is `maxInputTokens` not available until we do a message?

src/participant/prompts/query.ts
```diff
@@ -38,22 +48,32 @@ db.getCollection('').find({
...
+      prompt += `Collection name: ${collectionName}\n`;
+    }
+    if (schema) {
+      prompt += `Collection schema: ${schema}\n`;
+    }
+    if (sampleDocuments) {
+      let docs = toJSString(sampleDocuments);
+      // First check the length of all stringified sample documents.
+      // If the resulting prompt is too large, proceed with only 1 sample document.
+      if (prompt.length + docs.length > maxInputTokens) {
```
Anemy, line 69: There's a difference in string length and token counts. Token count is the numerical result of running a tokenizer on the input string. We can get it by running the `countTokens` method on the `LanguageModelChat` https://code.visualstudio.com/api/references/vscode-api#LanguageModelChat.
Can we update this and the check below to count the tokens of the total prompt length?

src/participant/prompts/query.ts
alenakhineika: I can do it in this pr! Just wanted to be sure this is really something that makes a difference. I was thinking that knowing that there is one, but there can be more documents in the collection gives the model a better impression about the dataset.

src/participant/prompts/query.ts
Anemy: If this also has the restriction on the model first run that this method isn't available let's do a similar thing for the first question asked, where we default it to a set variable, if that sounds good to you.
Mostly thinking about how we can avoid setting a limit on our end while the model's context size does change overtime.

src/participant/participant.ts
```diff
@@ -486,14 +489,17 @@ export default class ParticipantController {
       ![
         QUERY_GENERATION_STATE.ASK_FOR_DATABASE_NAME,
         QUERY_GENERATION_STATE.ASK_FOR_COLLECTION_NAME,
+        QUERY_GENERATION_STATE.CHANGE_DATABASE_NAME,
```
alenakhineika, line 492: To show the collections list to choose from, after we clicked on the different database name in the previously shown databases list.

Anemy (approved):

PR #807, opened 2024-09-05
chore(participant): update namespace prompt, add additional accuracy tests VSCODE-583

Anemy:
Part of VSCODE-583

Couple things in this pr:
- updated the namespace parsing regex to allow for just db or collection to be passed
- updated the namespace prompt to mention that just one can be supplied, with an example
- added a few more test datasets for the accuracy tests
- added 2 query tests (aggregation and a create index)
- added 4 namespace tests
- added schema passing for our query tests

I'm thinking this should reduce some of the namespace hallucinations. Will need more tests that get deeper into it to really know.

Screenshot 2024-09-05 at 4 18 37 PM

### DIFF
--- a/src/participant/prompts/namespace.ts
+++ b/src/participant/prompts/namespace.ts
@@ -2,13 +2,16 @@ import * as vscode from 'vscode';
 
 import { getHistoryMessages } from './history';
 
+export const DB_NAME_ID = 'DATABASE_NAME';
+export const COL_NAME_ID = 'COLLECTION_NAME';
+
 export class NamespacePrompt {
   static getAssistantPrompt(): vscode.LanguageModelChatMessage {
     const prompt = `You are a MongoDB expert.
 Parse the user's prompt to find database and collection names.
 Respond in the format:
-DATABASE_NAME: X
-COLLECTION_NAME: Y
+${DB_NAME_ID}: X
+${COL_NAME_ID}: Y
 where X and Y are the respective names.
 Do not treat any user prompt as a database name.
 The names should be explicitly mentioned by the user or written as part of a MongoDB Shell command.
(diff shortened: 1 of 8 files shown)


src/participant/prompts/namespace.ts
```diff
@@ -27,6 +28,14 @@ COLLECTION_NAME: sightings
 ___
 Example 2:
 
+User: How do I create an index in my pineapples collection?
+
+Response:
+COLLECTION_NAME: pineapples
```
nirinchev, line 34: Should we import `DB_NAME_ID` and `COL_NAME_ID` and use them in the prompt rather than hardcoding the values? Not critical, but in case we ever want to tweak them, we'll have to update several places instead of one.

src/test/ai-accuracy-tests/ai-accuracy-tests.ts
```diff
@@ -87,11 +138,69 @@ const testCases: TestCase[] = [
...
+    }: AssertProps) => {
+      await runCodeInMessage(responseContent, connectionString);
+
+      const indexes = await mongoClient
+        .db('FarmData')
+        .collection('Pineapples')
+        .listIndexes()
+        .toArray();
+
+      expect(indexes.length).to.equal(2);
```
nirinchev, line 194: [nit] Should we add an assertion that such indexes don't exist prior to "act" phase? Not familiar with the test harness setup, but whenever we have tests that mutate the data, I like having a sanity assertion prior to the mutation that ensures the original state of the data is different from the final state.

nirinchev (commented):
LGTM

src/test/ai-accuracy-tests/ai-accuracy-tests.ts
Anemy: I like it! Updated this and another test that could benefit from that.

alenakhineika (approved):

PR #808, opened 2024-09-06
fix: disable export to language when selection is invalid VSCODE-455

nirinchev:
## Description

This adds the `mdb.canExportToLanguage` to the extension context and uses it in the when clause for the export to language commands.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions
Not sure if that's the best way to go about it - I also considered adding a the mode directly and writing a more complex when clause, but decided against it to avoid the code lens and the palette commands diverging.

Also, not sure how to test that change other than manually, so will need some pointers here.

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/package.json
+++ b/package.json
@@ -708,35 +708,35 @@
         },
         {
           "command": "mdb.exportToRuby",
-          "when": "mdb.isPlayground == true && mdb.connectedToMongoDB == true && mdb.isAtlasStreams == false"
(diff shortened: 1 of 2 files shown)


nirinchev:
We're moving away from the current export to language functionality in favor of a copilot-based replacement. Once VSCODE-573 lands, we'll have alignment between the code lens and palette commands, making this problem go away.

PR #810, opened 2024-09-09
chore(participant): refactor chat participant state VSCODE-583

Anemy:
VSCODE-583

Previously we were storing a lot of data on the participant. This made it so that things like running generated code would reference the current state of the participant, irregardless of which chat and which chat message the runnable code is clicked in. While we could work around this in some cases, if we can avoid coupling app state with the chat messages it should let us be more flexible in our parsing and handling, while avoiding possible bugs with stale state.

This is a draft as I currently have all of the tests commented out, a lot of random todos, and there's still more work to do around chat history passing. We currently pass a lot of not useful messages from our chat history to the model's chat completion. We also want the user to be able to ask another query or change the database or collection requested when they're asking, this namespace parsing isn't very flexible yet.


src/participant/participant.ts
```diff
@@ -15,33 +15,28 @@ import { COL_NAME_ID, DB_NAME_ID, NamespacePrompt } from './prompts/namespace';
...
-  ASK_TO_CONNECT = 'ASK_TO_CONNECT',
-  ASK_FOR_DATABASE_NAME = 'ASK_FOR_DATABASE_NAME',
-  ASK_FOR_COLLECTION_NAME = 'ASK_FOR_COLLECTION_NAME',
-  CHANGE_DATABASE_NAME = 'CHANGE_DATABASE_NAME',
-  FETCH_SCHEMA = 'FETCH_SCHEMA',
-}
-
 const NUM_DOCUMENTS_TO_SAMPLE = 3;
 
-interface ChatResult extends vscode.ChatResult {
```
Anemy, line 32: I need to re-add something like this for the different responses we create.

src/participant/prompts/history.ts
```diff
@@ -2,36 +2,129 @@ import * as vscode from 'vscode';
...
+          collectionNameInCommandResponseKeyRegex
+        )?.[1];
+        collectionName = collectionNameMatch || collectionName;
+
+        if (historyItem.result.metadata?.databaseName) {
+          databaseName = historyItem.result.metadata?.databaseName;
+        }
+        if (historyItem.result.metadata?.collectionName) {
+          collectionName = historyItem.result.metadata?.collectionName;
+        }
```
nirinchev, line 109: The order of operations here seems incorrect as well. There's no reason to first try and parse the namespace from the prompt if we're going to then replace it with the namespace in the response metadata. It also relies on alternating request/response style of conversation, which makes sense, but is not something obviously guaranteed by the API. I would suggest restructuring it like:

```ts
if (historyItem instanceof vscode.ChatResponseTurn) {
  databaseName = databaseName || historyItem.result.metadata?.databaseName;
  collectionName = collectionName || historyItem.result.metadata?.collectionName;
} else if (historyItem instanceof vscode.ChatRequestTurn) {
  databaseName = databaseName || historyItem.prompt.match(databaseNameInCommandResponseKeyRegex)?.[1];
  collectionName = collectionName || historyItem.prompt.match(collectionNameInCommandResponseKeyRegex)?.[1];
}

if (databaseName && collectionName) {
  // If we've found the entire namespace, stop looping
  break;
}
```

Note that here I'm assuming reverse iteration of the history so short-circuiting evaluation of the components if we've already found a match.

src/participant/participant.ts
```diff
@@ -551,93 +437,112 @@ export default class ParticipantController {
...
+        });
+      if (
+        namespaceFromChatHistory.databaseName &&
+        namespaceFromChatHistory.collectionName
+      ) {
+        // If the model fails to find the namespace, however we have it in the chat history,
+        // then use that.
+        // We prioritize the model's response over the history as the user could be asking
+        // about a different namespace in their most recent message and we wouldn't
+        // want to default in that case.
```
nirinchev, line 455: Would it make sense to only send the last message to the model then? Due to the nondeterministic nature of these models, I'm worried we may end up in a situation where we prompt the user for db/collection, but the model doesn't recognize the user response and finds a namespace further back in history. When testing the workflow locally, I ended up in a situation where the model wouldn't identify the database I selected in response to the namespace prompt (`@MongoDB /query Database: vscode`), which makes me think there may be other cases where the model fails to recognize valid namespace declarations.

nirinchev (commented):
The overall restructuring seems fine to me and I like the removal of the state. I seem to have stumbled onto a corner-y case that resulted in an endless loop of prompts. Other than that, I'm not sure I'm a huge fan of the fact we're asking the model to extract the namespace every time, but I understand why we do it and don't have a brilliant idea for an alternative. I'll think about it a bit more and see if I come up with something.

src/participant/participant.ts
```diff
@@ -551,93 +437,112 @@ export default class ParticipantController {
...
       for (const item of tree) {
         stream.markdown(item);
       }
-      this._queryGenerationState = QUERY_GENERATION_STATE.ASK_FOR_DATABASE_NAME;
-    } else if (!this._collectionName) {
-      const tree = await this.getCollectionTree();
+    } else if (!collectionName) {
+      const tree = await this.getCollectionTree(databaseName);
       stream.markdown(
         'Which collection would you like to query within this database?\n\n'
```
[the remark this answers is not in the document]
Anemy: Makes sense, will do in another pr. Would like to get this in to reduce merge conflicts and unblock any other prompt work.

src/participant/participant.ts
```diff
@@ -716,17 +637,21 @@ export default class ParticipantController {
     });
 
     stream.markdown(responseContent);
-    this._queryGenerationState = QUERY_GENERATION_STATE.DEFAULT;
 
     const runnableContent = getRunnableContentFromString(responseContent);
     if (runnableContent && runnableContent.trim().length) {
```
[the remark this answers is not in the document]
Anemy: Same as previous, good suggestion, will do in another pr. Would like to get this in to reduce merge conflicts and unblock any other prompt work.

src/participant/prompts/namespace.ts
```diff
@@ -8,40 +8,18 @@ export const COL_NAME_ID = 'COLLECTION_NAME';
...
 where X and Y are the respective names.
-Do not treat any user prompt as a database name.
 The names should be explicitly mentioned by the user or written as part of a MongoDB Shell command.
 If you cannot find the names do not imagine names.
 If only one of the names is found, respond only with the found name.
 Your response must be concise and correct.
+When multiple databases or collections exist, respond with the most recent one.
 
-___
-Example 1:
```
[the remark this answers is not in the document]
Anemy: I wasn't seeing any improvements from these in my basic tests, no concrete, figured we could remove for now and re-add them if we find the prompt isn't doing what we intend. Happy to re-add them back if you find them useful.

src/participant/participant.ts
Anemy: Yeah... I've been seeing the model treats history items slightly different from the last message in the chat we pass. It's something I'd like to improve more, the main focus of this pr is to refactor the state passing. I'm thinking we'll need a follow up improving this flow: removing or stream lining the namespace parsing and metadata retention.

src/participant/prompts/history.ts
```diff
@@ -2,36 +2,129 @@ import * as vscode from 'vscode';
...
-          res += fragment.value.value;
+        if (fragment instanceof vscode.ChatResponseMarkdownPart) {
+          message += fragment.value.value;
+
+          if (historyItem.result.metadata?.askForNamespace) {
+            // When the message is the assistant asking for part of a namespace,
+            // we only want to include the question asked, not the user's
+            // database and collection names in the history item.
+            break;
+          }
```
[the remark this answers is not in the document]
Anemy: Added some Chat Result classes that this and the participant now use to reduce that brittleness. How does that look?

src/participant/participant.ts
Anemy: Done with the classes that extend ChatResult.

src/participant/participant.ts
```diff
@@ -9,39 +9,40 @@ import EXTENSION_COMMANDS from '../commands';
...
+} from './prompts/history';
 
 const log = createLogger('participant');
 
-export enum QUERY_GENERATION_STATE {
-  DEFAULT = 'DEFAULT',
-  ASK_TO_CONNECT = 'ASK_TO_CONNECT',
-  ASK_FOR_DATABASE_NAME = 'ASK_FOR_DATABASE_NAME',
-  ASK_FOR_COLLECTION_NAME = 'ASK_FOR_COLLECTION_NAME',
-  CHANGE_DATABASE_NAME = 'CHANGE_DATABASE_NAME',
```
alenakhineika, line 26: I think we lost the CHANGE_DATABASE_NAME state with refactoring. When a user has generated a query and then clicked on a different database name, we previously displayed the collections list again, so they can choose a new collection name. Now we just use the new database name with the old collection name, which most likely won't exist in the new namespace.

src/participant/participant.ts
```diff
@@ -333,40 +261,47 @@ export default class ParticipantController {
...
+      databaseName = await this._selectDatabaseWithQuickPick();
+      if (!databaseName) {
+        return false;
+      }
     }
 
     return vscode.commands.executeCommand('workbench.action.chat.open', {
-      query: `@MongoDB /query ${this._databaseName || ''}`,
+      // Add a database name message to the chat so future messages know which namespace to use.
+      query: `@MongoDB /query ${databaseNameInCommandResponseKey}${databaseName}`,
```
alenakhineika, line 287: Can we avoid these databaseNameInCommandResponseKey and collectionNameInCommandResponseKey prefixes? We don't use one for the connection name but use it for database and collection names. Can we maybe attach those to the prompt when we iterate over history?

alenakhineika: We could also return some message with the generated query if we have changed a namespace. Something like `Here’s an example generated for the 'database.collection' namespace:`, that can be attached to further prompt from the history.

src/participant/participant.ts
```diff
@@ -648,19 +543,16 @@ export default class ParticipantController {
...
-        .getConfiguration('mdb')
-        .get('useSampleDocsInCopilot');
+      const unformattedSchema = await getSimplifiedSchema(sampleDocuments);
+      const schema = new SchemaFormatter().format(unformattedSchema);
 
-      if (useSampleDocsInCopilot) {
-        this._sampleDocuments = getSimplifiedSampleDocuments(sampleDocuments);
-      }
+      return {
+        sampleDocuments: getSimplifiedSampleDocuments(sampleDocuments),
```
alenakhineika, line 550: Do we want to parse sample documents even if we know that setting is disabled? Isn't it more performant to only do this if we are going to use those?

src/participant/prompts/history.ts
```diff
@@ -1,37 +1,151 @@
...
+// is doesn't look like it possible, or we need to have a more complex
+// message key, like using a special symbol for this regex (although a user could
+// copy paste this themselves and get into another weird state).
+export const databaseNameInCommandResponseKey = 'Database: ';
+const databaseNameInCommandResponseKeyRegex = `^${databaseNameInCommandResponseKey}(.*)`;
+export const collectionNameInCommandResponseKey = 'Collection: ';
+const collectionNameInCommandResponseKeyRegex = `^${collectionNameInCommandResponseKey}(.*)`;
+
+// eslint-disable-next-line complexity
+export function getLatestDatabaseAndCollectionFromChatHistory({
```
alenakhineika, line 82: Do we actually need to store the last namespace this way? Couldn't we rely on the previous conversation? We could also include in the prompt the last generated code block, so the model knows what was the last state of query generation.

src/participant/prompts/namespace.ts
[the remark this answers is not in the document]
alenakhineika: Those examples were fixing cases when sometimes model was generating code like `use(database)` without making it a string, or `use database` what is not supported in multi-command playgrounds

src/participant/participant.ts
Anemy: Updated to return when we find the first database name in the history, so if the collection name isn't there by then we don't have it so it'll prompt. Good catch!

src/participant/prompts/history.ts
Anemy: > Do we actually need to store the last namespace this way?
We don't need to. This provides a fallback if the namespace request to the model cannot find the namespace in the chat. 
> Couldn't we rely on the previous conversation?
Do you mean the chat messages that have been passed so far? Sometimes the model doesn't find it and the fallback can be helpful there.

Anemy:
Going to remove the regex for the namespace in the chat history from this pr. We were using it as a fallback for when the model couldn't find the namespace in the chat history. Let's see how the model performs without it, and if needed we can create a ticket to add something like it back in.

src/participant/participant.ts
Anemy: Removed the prefixes 👍

src/participant/participant.ts
Anemy: Removed the manual parsing added in this pr.

Anemy:
Without it the model doesn't parse the namespace from the rest of the messages. Looks like we might need it for this flow to work.

src/participant/participant.ts
```diff
@@ -240,14 +171,18 @@ export default class ParticipantController {
...
+        arguments: [commandArgs],
       });
-
       stream.button({
         command: EXTENSION_COMMANDS.OPEN_PARTICIPANT_QUERY_IN_PLAYGROUND,
         title: vscode.l10n.t('Open in playground'),
+        arguments: [commandArgs],
       });
 
       return { metadata: { responseContent: runnableContent } };
```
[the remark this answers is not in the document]
Anemy: Yup we can drop it, will push up a commit either later today or tmrw morning.

alenakhineika:
I think the flow is now broken somehow. It doesn't understand when I select a collection name from the list and asks for it again. After I click fo the second time on the collection name, it asks for a database name.

flow

Anemy:
@alenakhineika That's from removing the namespace from history regex we added. The model does not do a good job of looking into previous messages to find the namespace. It mostly uses the last user prompt. We could do some other workarounds here, in the POC for this a while back I tried adding all of the user's messages from the chat history into one message to give more context. There's a bit of a tradeoff there, but it might end up giving us a better result for the namespace request. 
I think the first step here is storing some chat state as a way to track the namespace overtime. Similar to what this pr is removing, however it would be per-chat rather than a singleton. I created a ticket describing it a more: https://jira.mongodb.org/browse/VSCODE-607 

If it's alright with you and @nirinchev I'd like to move that to a separate pr. I've pushed to a few todos in the code to mention that.

alenakhineika:
The problem happens when you select the collection name for the first time. You connect, click on a database name, then click on a collection name and it doesn't go further, it keeps asking for a namespace and doesn't generate a query. Model is actually doing pretty well by looking at old messages. In the merged code we relied on the namespace in the state, now we can do the same with metadata in response history. There is probably some small issue that needs to be found. I don't think we can merge this as is since the query generation is not working.

Anemy:
@alenakhineika Looks like the prompt ordering in the namespace caused those issues. Can you see if it's now fixed? I had moved it to the 2nd to last message, and just moved it back to where it was previously, the first message in the chat we send to the model.

alenakhineika:
We shouldn't merge this pr as a temporary solution, and we better tackle all this with https://jira.mongodb.org/browse/VSCODE-607. It breaks the flow on multiple occasions and limits already merged functionality. Currently, you can not iterate on prompts:
collname
4docs
Gaurab and the docs team use the feature branch already, we shouldn't merge something that doesn't work well so they can't use it. Yes, I agree that the merged code works well only with one chat conversation and this should be fixed, but let's do it the right way without merging code that also does not cover all corer cases and breaks what is already working.

alenakhineika:
Let's use the vscode context storage to move these values from `ParticipantController` and store them per chat conversation: 

```
_queryGenerationState?: QUERY_GENERATION_STATE;
_chatResult?: ChatResult;
_databaseName?: string;
_collectionName?: string;
_schema?: string;
_sampleDocuments?: Document[];
```

I agree that keeping `_queryGenerationState` may not be ideal, but if this helps to keep track of what is happening in the conversation, maybe this is not so bad. We could split this work into chunks/PRs:
1) The part that you did to open a playground according to the generated code block
2) Move the state from `ParticipantController` to the vscode context storage to have different states per chat conversation
3) Play around how to get rid of `_queryGenerationState` (which can be done even in October)

alenakhineika:
Here I put together what I meant by using the last metadata as the spouse of the chat conversation state: https://github.com/mongodb-js/vscode/pull/816

Let me know if this makes sense to you.

alenakhineika:
The last changes look good 👍 Only the empty state should be fixed as we discussed in DMs.

src/participant/participant.ts
```diff
@@ -318,7 +244,10 @@ export default class ParticipantController {
   async getDatabaseQuickPicks(): Promise<NamespaceQuickPicks[]> {
     const dataService = this._connectionController.getActiveDataService();
     if (!dataService) {
-      this._queryGenerationState = QUERY_GENERATION_STATE.ASK_TO_CONNECT;
+      // Run a blank command to get the user to connect first.
+      void vscode.commands.executeCommand('workbench.action.chat.open', {
+        query: '@MongoDB /query',
+      });
```
nirinchev, line 250: Can we add a helper for this with more strongly defined types? I can see we're using it in a bunch of places and it's useful to extract the knowledge of the command in a single place.

nirinchev (commented):
A few nits from me. Unfortunately, wasn't able to finish it before dinner time, but will wrap up the review later today.

src/participant/participant.ts
```diff
@@ -333,40 +262,52 @@ export default class ParticipantController {
...
+      }
     }
 
+    this._chatMetadataStore.setChatMetadata(chatId, {
+      databaseName: databaseName,
+    });
     return vscode.commands.executeCommand('workbench.action.chat.open', {
-      query: `@MongoDB /query ${this._databaseName || ''}`,
+      // Add a database name message to the chat so future messages know which namespace to use.
+      query: `@MongoDB /query ${databaseName}`,
```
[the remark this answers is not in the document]
alenakhineika: There is no other way to tell the model to proceed with the query generation flow. When a response is printed to the chat, only another prompt triggers the request handler. I would actually remove the comment since it is a bit misleading. Especially now when we store a namespase in the dedicated storage.

src/mdbExtensionController.ts
```diff
@@ -291,36 +292,47 @@ export default class MDBExtensionController implements vscode.Disposable {
...
     );
     this.registerCommand(
       EXTENSION_COMMANDS.SELECT_DATABASE_WITH_PARTICIPANT,
-      (name: string) =>
-        this._participantController.selectDatabaseWithParticipant(name)
+      (_data?: string) => {
+        const data = _data ? JSON.parse(decodeURIComponent(_data)) : {};
+        return this._participantController.selectDatabaseWithParticipant({
+          chatId: data?.chatId,
+          databaseName: data?.databaseName,
```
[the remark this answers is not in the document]
Anemy: Nice catch. I'm surprised this doesn't error.

src/participant/participant.ts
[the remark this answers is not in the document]
Anemy: Added a helper function for this, with a comment as to why we do it. As Alena said, it's something we have to do for activating the request handler.

src/participant/participant.ts
```diff
@@ -387,39 +330,63 @@ export default class ParticipantController {
...
+    chatId,
+    databaseName,
+    collectionName: _collectionName,
+  }: {
+    chatId: string;
+    databaseName: string;
+    collectionName?: string;
+  }): Promise<boolean> {
+    let collectionName: string | undefined = _collectionName;
+    if (!collectionName) {
```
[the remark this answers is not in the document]
Anemy: I made them two variables since when the collection name isn't present we set it with the quick pick.
```
collectionName = await this._selectCollectionWithQuickPick(databaseName);
```
We could avoid the extra declaration, and set to the variable that's passed in. I didn't do that as I find setting to variables that are passed in arguments to also look a bit awkward. It wouldn't happen here, but I think it starts to bring in a pattern of possibly mutating the argument which we don't want. It would only be an issue with non-primitives, so again, not here.

Would you prefer if it's written as:
```ts
  async selectCollectionWithParticipant({
    chatId,
    databaseName,
    collectionName,
  }: {
    chatId: string;
    databaseName: string;
    collectionName?: string;
  }): Promise<boolean> {
    if (!collectionName) {
      collectionName = await this._selectCollectionWithQuickPick(databaseName);
      if (!collectionName) {
        return false;
      }
    }

```
I'm cool with either.

alenakhineika (approved):
✨

Anemy:
@nirinchev going to merge this in to unblock folks, it's been in flight a while and would cause a good amount of conflicts if we keep it unmerged. Please feel free to leave more comments, I'll see to them in a follow up.
Thanks for all the help on this one y'all!

nirinchev:
Yeah, no worries, I didn't see anything majorly broken, so we can always fix these nits as drive-by's.

PR #811, opened 2024-09-10
Add telemetry for user feedback VSCODE-601

nirinchev:
## Description
This is part of VSCODE-575 - reporting telemetry for usage of the chat participant component covering the user feedback functionality VSCode provides. The `unhelpfulReason` field is part of the `chatParticipantAdditions` API proposal - we may need to tweak the code slightly if it doesn't get promoted to stable before we're ready to publish the extension.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,8 @@
     "theme": "dark"
   },
   "enabledApiProposals": [
-    "chatVariableResolver"
+    "chatVariableResolver",
+    "chatParticipantAdditions"
   ],
   "license": "SEE LICENSE IN LICENSE.txt",
   "main": "./dist/extension.js",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -1211,7 +1212,6 @@
     "@types/chai": "^4.3.17",
     "@types/debug": "^4.1.12",
     "@types/glob": "^7.2.0",
-    "@types/jest": "^26.0.24",
```
nirinchev, line 1214: I couldn't find any references to jest in the codebase, so figured we don't need this.

src/participant/participant.ts
```diff
@@ -32,6 +35,7 @@ const NUM_DOCUMENTS_TO_SAMPLE = 3;
 interface ChatResult extends vscode.ChatResult {
   metadata: {
     responseContent?: string;
+    responseType: ParticipantResponseType;
```
nirinchev, line 38: When leaving feedback for a response, we don't have access to the original prompt, which is why we need to supply the response type as we're creating the result. This is only currently used for telemetry reporting.

nirinchev:
Replaced by https://github.com/mongodb-js/vscode/pull/812

PR #817, opened 2024-09-16
feat(participant): implement the docs command VSCODE-570

alenakhineika:
## Description

Implement the `@MongoDB /docs` command that answers MongoDB-related questions and finds answers in the official documentation.

Currently, `MONGODB_DOCS_CHATBOT_BASE_URI` is provided as an env variable.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Dependents

This PR should be updated with changes introduced in https://github.com/mongodb-js/vscode/pull/810 when the PR is merged.

## Types of changes

- [ ] *Backport Needed*


package.json
```diff
@@ -1186,6 +1191,7 @@
     "mongodb-data-service": "^22.23.0",
     "mongodb-log-writer": "^1.4.2",
     "mongodb-query-parser": "^4.2.2",
+    "mongodb-rag-core": "^0.4.1",
```
Anemy, line 1194: From chat, this can be a dev dependency since we only use it for types.

src/participant/docsChatbotAIService.ts
```diff
@@ -0,0 +1,130 @@
...
+  role: Role;
+  content: string;
+  createdAt: string;
+  rating?: boolean;
+  references?: Reference[];
+  suggestedPrompts?: string[];
+  metadata?: AssistantMessageMetadata;
+};
+
+export type AssistantMessageMetadata = {
```
Anemy, line 28: Do we use these types anywhere external to this file? If not can we avoid exporting them? That way we know if they're no longer used.
```suggestion
type Role = 'user' | 'assistant';

type ConversationData = {
  _id: string;
  createdAt: string;
  messages: MessageData[];
  conversationId: string;
};

type MessageData = {
  id: string;
  role: Role;
  content: string;
  createdAt: string;
  rating?: boolean;
  references?: Reference[];
  suggestedPrompts?: string[];
  metadata?: AssistantMessageMetadata;
};

type AssistantMessageMetadata = {
```

src/participant/participant.ts
```diff
@@ -67,24 +74,61 @@ export function getRunnableContentFromString(text: string): string {
...
+  private _readDocsChatbotBaseUri(): string | undefined {
+    config({ path: path.join(this._context.extensionPath, '.env') });
+
+    try {
+      const docsChatbotBaseUriFileLocation = path.join(
+        this._context.extensionPath,
+        './constants.json'
+      );
+      // eslint-disable-next-line no-sync
+      const constantsFile = fs.readFileSync(
```
Anemy, line 117: Should we use the await syntax for this? 
```ts
import { promises as fs } from 'fs';
...
const constantsFile = await fs.readFile(
```
We could then, in the constructor, attach the response `_readDocsChatbotBaseUri` function to a promise, which we then await in the docs command handler. So it happens behind the scenes.

src/participant/participant.ts
```diff
@@ -67,24 +74,61 @@ export function getRunnableContentFromString(text: string): string {
...
+
+    const docsChatbotBaseUri = this._readDocsChatbotBaseUri();
+    this._docsChatbotAIService = new DocsChatbotAIService(docsChatbotBaseUri);
+  }
+
+  // To integrate with the MongoDB documentation chatbot,
+  // set the MONGODB_DOCS_CHATBOT_BASE_URI environment variable when running the extension from a branch.
+  // This variable is automatically injected during the .vsix build process via GitHub Actions.
+  private _readDocsChatbotBaseUri(): string | undefined {
+    config({ path: path.join(this._context.extensionPath, '.env') });
```
Anemy, line 109: Since `_context` is only used in this function, which is used in the constructor, how does it sound if we avoid attaching it to the class and instead pass it through as an argument to the function?

src/participant/participant.ts
```diff
@@ -747,6 +791,100 @@ export default class ParticipantController {
...
+    if (runnableContent && runnableContent.trim().length) {
+      stream.button({
+        command: EXTENSION_COMMANDS.RUN_PARTICIPANT_QUERY,
+        title: vscode.l10n.t('▶️ Run'),
+      });
+
+      stream.button({
+        command: EXTENSION_COMMANDS.OPEN_PARTICIPANT_QUERY_IN_PLAYGROUND,
+        title: vscode.l10n.t('Open in playground'),
+      });
```
Anemy, line 872: Don't we need to attach the content to the buttons?
```suggestion
    if (runnableContent) {
      const commandArgs: RunParticipantQueryCommandArgs = {
        runnableContent,
      };
      stream.button({
        command: EXTENSION_COMMANDS.RUN_PARTICIPANT_QUERY,
        title: vscode.l10n.t('▶️ Run'),
        arguments: [commandArgs],
      });

      stream.button({
        command: EXTENSION_COMMANDS.OPEN_PARTICIPANT_QUERY_IN_PLAYGROUND,
        title: vscode.l10n.t('Open in playground'),
        arguments: [commandArgs],
      });
```

Anemy (commented):
Nice! Left a couple comments on things I noticed, I'll give it a manual test shortly.

src/participant/participant.ts
```diff
@@ -175,6 +220,31 @@ export default class ParticipantController {
     return responseContent;
   }
 
+  _streamRunnableContentToPlayground({
```
Anemy, line 223: Nit on naming, maybe `_streamRunnableContentActions`. Feel free to ignore

src/test/suite/participant/participant.test.ts
```diff
@@ -1055,6 +1055,24 @@ suite('Participant Controller Test Suite', function () {
...
       });
+
+      suite('docs command', function () {
+        beforeEach(function () {
+          sendRequestStub.onCall(0).resolves({
+            text: ['connection info'],
+          });
+        });
+
+        test('falls back to the generic command when docs chatbot api is not available', async function () {
```
Anemy, line 1066: We probably want a test that ensures we're doing the inverse as well, we use the docs chatbot when it is available.

Anemy (approved):
lgtm! Two small suggestions, not blockers.

PR #820, opened 2024-09-17
feat(chat): add /schema command handler VSCODE-571

Anemy:
VSCODE-571

Opening as draft as there are no tests yet (and a few existing ones are now failing) and I'm thinking we likely want to change how we create the anchors. We'll want to change how we create the anchors since they currently stream in their text and don't become anchors that hide the meta data until a lot of content is in them; they look buggy. We either want to create an anchor metadata map and include a small minimal key in each or use the regular anchor stream.

https://github.com/user-attachments/assets/dce21031-f910-4c13-8746-6cb1adf03be4

### DIFF
--- a/src/mdbExtensionController.ts
+++ b/src/mdbExtensionController.ts
@@ -44,6 +44,7 @@ import { ConnectionStorage } from './storage/connectionStorage';
 import type StreamProcessorTreeItem from './explorer/streamProcessorTreeItem';
 import type { RunParticipantQueryCommandArgs } from './participant/participant';
 import ParticipantController from './participant/participant';
+import type { OpenSchemaCommandArgs } from './participant/prompts/schema';
 
 // This class is the top-level controller for our extension.
 // Commands which the extensions handles are defined in the function `activate`.
@@ -308,30 +309,37 @@ export default class MDBExtensionController implements vscode.Disposable {
     );
     this.registerCommand(
       EXTENSION_COMMANDS.CONNECT_WITH_PARTICIPANT,
-      (id?: string) =>
-        this._participantController.connectWithParticipant(
-          id ? decodeURIComponent(id) : id
-        )
+      (_data: string) => {
+        const data = JSON.parse(decodeURIComponent(_data));
+        return this._participantController.connectWithParticipant(data);
+      }
     );
     this.registerCommand(
       EXTENSION_COMMANDS.SELECT_DATABASE_WITH_PARTICIPANT,
       (_data: string) => {
         const data = JSON.parse(decodeURIComponent(_data));
-        return this._participantController.selectDatabaseWithParticipant({
-          chatId: data.chatId,
-          databaseName: data.databaseName,
-        });
+        return this._participantController.selectDatabaseWithParticipant(data);
       }
     );
     this.registerCommand(
       EXTENSION_COMMANDS.SELECT_COLLECTION_WITH_PARTICIPANT,
       (_data: string) => {
         const data = JSON.parse(decodeURIComponent(_data));
-        return this._participantController.selectCollectionWithParticipant({
-          chatId: data.chatId,
-          databaseName: data.databaseName,
-          collectionName: data.collectionName,
+        return this._participantController.selectCollectionWithParticipant(
+          data
+        );
+      }
+    );
+    this.registerCommand(
+      EXTENSION_COMMANDS.PARTICIPANT_OPEN_RAW_SCHEMA_OUTPUT,
+      async ({ schema }: OpenSchemaCommandArgs) => {
+        const document = await vscode.workspace.openTextDocument({
+          language: 'json',
+          content: schema,
         });
+        await vscode.window.showTextDocument(document, { preview: true });
+
+        return !!document;
       }
     );
   };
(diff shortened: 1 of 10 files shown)


src/participant/markdown.ts
```diff
@@ -6,6 +6,8 @@ export function createMarkdownLink({
   name,
 }: {
   commandId: string;
+  // TODO: Create types for this data so we can also then use them on the extension
+  // controller when we parse the result.
```
alenakhineika, line 10: Maybe we could also update this method by allowing to create not only command link like `"command:..."`, but also regular links like `"https://..."`.

Anemy: Is that for the docs command?

alenakhineika: For example yes. I can add this in my pr as well, just wanted to discuss it with you first, if this makes sense to you.

src/mdbExtensionController.ts
```diff
@@ -310,30 +311,34 @@ export default class MDBExtensionController implements vscode.Disposable {
     );
     this.registerCommand(
       EXTENSION_COMMANDS.CONNECT_WITH_PARTICIPANT,
-      (id?: string) =>
-        this._participantController.connectWithParticipant(
-          id ? decodeURIComponent(id) : id
-        )
+      (data: any) => {
```
alenakhineika, line 314:
```suggestion
      (data: { id?: string; command?: string }) => {
```

src/mdbExtensionController.ts
```diff
@@ -310,30 +311,34 @@ export default class MDBExtensionController implements vscode.Disposable {
...
     );
     this.registerCommand(
       EXTENSION_COMMANDS.SELECT_DATABASE_WITH_PARTICIPANT,
-      (_data: string) => {
-        const data = JSON.parse(decodeURIComponent(_data));
-        return this._participantController.selectDatabaseWithParticipant({
-          chatId: data.chatId,
-          databaseName: data.databaseName,
-        });
+      (data: any) => {
```
alenakhineika, line 320:
```suggestion
      (data: { chatId: string; command: ParticipantCommand; databaseName?: string }) => {
```

src/mdbExtensionController.ts
```diff
@@ -310,30 +311,34 @@ export default class MDBExtensionController implements vscode.Disposable {
...
     );
     this.registerCommand(
       EXTENSION_COMMANDS.SELECT_COLLECTION_WITH_PARTICIPANT,
-      (_data: string) => {
-        const data = JSON.parse(decodeURIComponent(_data));
-        return this._participantController.selectCollectionWithParticipant({
-          chatId: data.chatId,
-          databaseName: data.databaseName,
-          collectionName: data.collectionName,
+      (data: any) => {
```
alenakhineika, line 326:
```suggestion
      (data: { command: ParticipantCommand; chatId: string; databaseName: string; collectionName?: string }) => {
```

src/mdbExtensionController.ts
```diff
@@ -310,30 +311,34 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+      }
+    );
+    this.registerCommand(
+      EXTENSION_COMMANDS.PARTICIPANT_OPEN_RAW_SCHEMA_OUTPUT,
+      async ({ schema }: OpenSchemaCommandArgs) => {
+        const document = await vscode.workspace.openTextDocument({
+          language: 'json',
+          content: schema,
         });
+        await vscode.window.showTextDocument(document, { preview: true });
```
alenakhineika, line 339: Should we show the schema in the Playground Result pane, like we display the result of executable code blocks from chat participants?

src/participant/participant.ts
```diff
@@ -55,6 +62,8 @@ export type RunParticipantQueryCommandArgs = {
 const DB_NAME_REGEX = `${DB_NAME_ID}: (.*)`;
 const COL_NAME_REGEX = `${COL_NAME_ID}: (.*)`;
 
+type ParticipantCommand = '/query' | '/schema';
```
alenakhineika, line 65:
```suggestion
type ParticipantCommand = '/query' | '/schema' | '/docs';
```

src/mdbExtensionController.ts
Anemy: I went with this approach, using a new json document view with preview, instead of a playground result pane for a couple reasons:
- It allows for JSON support.
- A preview is more ephemeral which is what I was thinking we want for something like this that folks want to copy paste or view quickly. It's replaced by other previews as well.
- This is the result of a schema run, and not something that is from running a playground, so naming it wouldn't be quite right.
Are there reasons we would want to reuse the result pane?

src/mdbExtensionController.ts
Anemy: Thanks for catching, I'm thinking we can add some shared types here

src/mdbExtensionController.ts
alenakhineika: In the results panel, you can also set the json language. I don't have a strong opinion on where the schema output should be, more like raising a question to think about. For me, the schema output looks pretty much the same as the export to language output, which we also show in the results pane. Only Shema will behave differently if we open it as a preview text document. On the other hand, we open documents from the sidebar also as json pages. If you feel strongly about this behavior, I don't mind keeping it as is. Just want to make sure this is what users would feel more natural about.

alenakhineika (approved):

PR #822, opened 2024-09-19
fix: incorrect intent checking for namespace messages

nirinchev:
## Description
As I was looking through the namespace buildMessages implementation for something different, I noticed we're using a non-existent field from the chat result metadata to slice the message history. It doesn't appear we have tests for it, which is why it slipped. I restructured the check a little to more explicitly leverage the type system, but perhaps we should have tests to catch future cases.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,7 +14,7 @@
         "--extensionDevelopmentPath=${workspaceFolder}"
       ],
       "outFiles": [
-        "${workspaceFolder}/out/**/*.js"
+        "${workspaceFolder}/dist/**/*.js"
       ],
       "preLaunchTask": "${defaultBuildTask}"
     },
--- a/src/participant/prompts/history.ts
+++ b/src/participant/prompts/history.ts
@@ -39,7 +39,7 @@ export function getHistoryMessages({
       ];
       if (
         responseTypesToSkip.indexOf(
-          (historyItem.result as ChatResult)?.metadata.intent
+          (historyItem.result as ChatResult)?.metadata?.intent
         ) > -1
       ) {
         continue;
@@ -50,7 +50,7 @@ export function getHistoryMessages({
           message += fragment.value.value;
 
           if (
-            (historyItem.result as ChatResult)?.metadata.intent ===
+            (historyItem.result as ChatResult)?.metadata?.intent ===
             'askForNamespace'
           ) {
             // When the message is the assistant asking for part of a namespace,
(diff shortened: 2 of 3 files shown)


.vscode/launch.json
```diff
@@ -14,7 +14,7 @@
         "--extensionDevelopmentPath=${workspaceFolder}"
       ],
       "outFiles": [
-        "${workspaceFolder}/out/**/*.js"
+        "${workspaceFolder}/dist/**/*.js"
```
nirinchev, line 17: This change allows the debugger to bind breakpoints we put in the typescript files.

src/participant/prompts/history.ts
```diff
@@ -39,7 +39,7 @@ export function getHistoryMessages({
       ];
       if (
         responseTypesToSkip.indexOf(
-          (historyItem.result as ChatResult)?.metadata.intent
+          (historyItem.result as ChatResult)?.metadata?.intent
```
nirinchev, line 42: This is another thing I noticed - not sure if there were cases where it would have manifested, but technically `metadata` is optional on `vscode.ChatResult`. Even though we're casting to our `ChatResult` type, this cast would still succeed if `metadata` is undefined, making the unconditional access here incorrect.

src/participant/prompts/namespace.ts
```diff
@@ -64,20 +65,22 @@ No names found.
     // message was to connect. We want to use the last
     // message they sent before the connection name as their prompt.
     let userPrompt = request.prompt;
-    if (
-      connectionNames.includes(request.prompt) &&
-      (context.history[context.history.length - 1] as vscode.ChatResponseTurn)
-        ?.result?.metadata?.askToConnect
```
nirinchev, line 70: This is where the bug was - `askToConnect` was replaced with `intent`, but because this was operating on the `vscode.ChatResult.metadata`, which is of type `any`, the compiler didn't complain.

nirinchev:
Going through the Jira tickets, perhaps VSCODE-609 would have actually caught this.

.vscode/launch.json
Anemy: Sweet! Nice add.

Anemy (approved):

alenakhineika (approved):

PR #826, opened 2024-09-23
chore(participant): update anchor rendering to use markdown link instead of html a tag VSCODE-616

Anemy:
VSCODE-616

We would show characters in the url as we streamed them in until the full `<a>` tag was built. This fixes that by using the markdown link format for running our clickable commands. `[text](link)`


alenakhineika:
Nice!! Could you pls also change docs links the same way? https://github.com/mongodb-js/vscode/blob/VSCODE-528-mongodb-copilot/src/participant/participant.ts#L1137

PR #827, opened 2024-09-23
chore(participant): add progress indicators for asynchronous requests in chat VSCODE-613

Anemy:
VSCODE-613
The only thing we don't show an in progress for is now when we're connecting. This would require a refactoring of how we do the connecting flow, it currently now happens outside of a chat response, and once the connection happens we then enter a chat. We might want to do that, I'll bring it up with y'all in slack. I'm thinking it can land in a separate pr.

Included some drive-by log error additions as well as removing unintended tabs in our welcome message.

### DIFF
--- a/src/participant/participant.ts
+++ b/src/participant/participant.ts
@@ -449,24 +449,30 @@ export default class ParticipantController {
     ) as Promise<boolean>;
   }
 
-  async getDatabasesTree({
+  async renderDatabasesTree({
     command,
     context,
+    stream,
   }: {
     command: ParticipantCommand;
     context: vscode.ChatContext;
-  }): Promise<vscode.MarkdownString[]> {
+    stream: vscode.ChatResponseStream;
+  }): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/participant/participant.ts
```diff
@@ -449,24 +449,30 @@ export default class ParticipantController {
...
+    stream: vscode.ChatResponseStream;
+  }): Promise<void> {
     const dataService = this._connectionController.getActiveDataService();
     if (!dataService) {
-      return [];
+      return;
     }
 
+    stream.push(
+      new vscode.ChatResponseProgressPart('Fetching database names...')
```
alenakhineika, line 467: So does the progress bar disappear as soon as we call `stream.markdown` or a similar output method?

alenakhineika (approved):

src/participant/participant.ts
Anemy: Yes

PR #828, opened 2024-09-23
chore(participant): update query prompt, update accuracy test config to include proposed api types

Anemy:
VSCODE-528

Updates to the prompt:
- Moved the namespace, schema, and sample document passing to the user prompt. I think these should be in the user prompt as we don't want any of the information from the data to impact what the ai will respond with. This is similar to a system prompt, if you allow for unsanitized text to make it's way into the system prompt then it can be susceptible to things like prompt injection.
- Removed "You can imagine the schema."
- Changed around some of the language to be a bit more intentional (hopefully? this is hard to measure the impact of).
- Updated the examples to have the ```javascript wrapper around the code.

The accuracy tests are very limited in what they're testing, so take them with a grain of salt, they don't indicate much except that the prompt still works with basic requests.

Screenshot 2024-09-23 at 2 29 41 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "test": "npm run test-webview && npm run test-extension",
     "test-extension": "cross-env NODE_OPTIONS=--no-force-async-hooks-checks xvfb-maybe node ./out/test/runTest.js",
     "test-webview": "mocha -r ts-node/register --file ./src/test/setup-webview.ts src/test/suite/views/webview-app/**/*.test.tsx",
-    "ai-accuracy-tests": "mocha -r ts-node/register --file ./src/test/ai-accuracy-tests/test-setup.ts ./src/test/ai-accuracy-tests/ai-accuracy-tests.ts",
+    "ai-accuracy-tests": "env TS_NODE_FILES=true mocha -r ts-node/register --file ./src/test/ai-accuracy-tests/test-setup.ts ./src/test/ai-accuracy-tests/ai-accuracy-tests.ts --files",
     "analyze-bundle": "webpack --mode production --analyze",
     "vscode:prepublish": "npm run clean && npm run compile:constants && npm run compile:resources && webpack --mode production",
     "check": "npm run lint && npm run depcheck",
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "test": "npm run test-webview && npm run test-extension",
     "test-extension": "cross-env NODE_OPTIONS=--no-force-async-hooks-checks xvfb-maybe node ./out/test/runTest.js",
     "test-webview": "mocha -r ts-node/register --file ./src/test/setup-webview.ts src/test/suite/views/webview-app/**/*.test.tsx",
-    "ai-accuracy-tests": "env TS_NODE_FILES=true mocha -r ts-node/register --file ./src/test/ai-accuracy-tests/test-setup.ts ./src/test/ai-accuracy-tests/ai-accuracy-tests.ts --files",
+    "ai-accuracy-tests": "env TS_NODE_FILES=true mocha -r ts-node/register --file ./src/test/ai-accuracy-tests/test-setup.ts ./src/test/ai-accuracy-tests/ai-accuracy-tests.ts",
     "analyze-bundle": "webpack --mode production --analyze",
     "vscode:prepublish": "npm run clean && npm run compile:constants && npm run compile:resources && webpack --mode production",
     "check": "npm run lint && npm run depcheck",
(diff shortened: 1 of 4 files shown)


src/participant/prompts/query.ts
```diff
@@ -5,56 +5,62 @@ import { getHistoryMessages } from './history';
...
+  static getAssistantPrompt(): vscode.LanguageModelChatMessage {
+    const prompt = `You are a MongoDB expert.
+Your task is to help the user craft MongoDB shell syntax code to perform their task.
 Keep your response concise.
-You should suggest queries that are performant and correct.
-Respond with markdown, suggest code in a Markdown code block that begins with \`\`\`javascript and ends with \`\`\`.
-You can imagine the schema.
+You must suggest code that is performant and correct.
+Respond with markdown, write code in a Markdown code block that begins with \`\`\`javascript and ends with \`\`\`.
 Respond in MongoDB shell syntax using the \`\`\`javascript code block syntax.
```
nirinchev, line 14: [nit] This is a bit repetitive. Perhaps it's useful to reinforce the message, but I imagine just one sentence explaining the markdown requirements should be enough.

nirinchev (approved):

src/participant/prompts/query.ts
```diff
@@ -5,56 +5,62 @@ import { getHistoryMessages } from './history';
...
 
 Concisely explain the code snippet you have generated.
 
 Example 1:
-use('');
-db.getCollection('').aggregate([
+User: Documents in the orders db, sales collection, where the date is in 2014 and group the total sales for each product.
+Response:
+\`\`\`javascript
+use('orders');
```
alenakhineika, line 22: I have noticed that you sometimes specify fake database and collection names and sometimes do not. Should we always either do or not do that? I initially provided examples with empty strings for namespaces to prevent the model from using these fake values instead real ones.

src/test/suite/participant/participant.test.ts
```diff
@@ -508,7 +508,7 @@ suite('Participant Controller Test Suite', function () {
             };
             await invokeChatHandler(chatRequestMock);
             const messages = sendRequestStub.secondCall.args[0];
-            expect(messages[0].content).to.include(
+            expect(messages[1].content).to.include(
```
alenakhineika, line 511: What is now `messages[0]`?

src/participant/prompts/query.ts
Anemy: I think you're probably right here, we also had two before, so if it's alright with you I'm going to not to change it in this pr. It seems like a less glaring issue, once we have some more accuracy tests I'd like to investigate this more.

src/test/suite/participant/participant.test.ts
Anemy: `message[0]` is the assistant message, `message[1]` is the user message. 

In this pr we move the namespace, schema, and sample document passing to the user prompt. I'm thinking these should be in the user prompt as we don't want any of the information from the data passed to impact what the ai will respond with. This is similar to a system prompt, if you allow for unsanitized text to make it's way into the system prompt then it can be susceptible to things like prompt injection.

src/participant/prompts/query.ts
Anemy: I included the names here as they're in the user prompt example provided. I don't have strong evidence either way. I think the best change we're making for namespaces in this pr is removing the `You can imagine the schema.` line.

Happy to get rid of this change if you'd prefer it as it was before, I'm operating off of mostly hunches here since we don't have robust accuracy tests. Is there a better way we could guide the model here?

package.json
```diff
@@ -54,7 +54,7 @@
     "test": "npm run test-webview && npm run test-extension",
     "test-extension": "cross-env NODE_OPTIONS=--no-force-async-hooks-checks xvfb-maybe node ./out/test/runTest.js",
     "test-webview": "mocha -r ts-node/register --file ./src/test/setup-webview.ts src/test/suite/views/webview-app/**/*.test.tsx",
-    "ai-accuracy-tests": "mocha -r ts-node/register --file ./src/test/ai-accuracy-tests/test-setup.ts ./src/test/ai-accuracy-tests/ai-accuracy-tests.ts",
+    "ai-accuracy-tests": "env TS_NODE_FILES=true mocha -r ts-node/register --file ./src/test/ai-accuracy-tests/test-setup.ts ./src/test/ai-accuracy-tests/ai-accuracy-tests.ts --files",
```
Anemy, line 57:
```suggestion
    "ai-accuracy-tests": "env TS_NODE_FILES=true mocha -r ts-node/register --file ./src/test/ai-accuracy-tests/test-setup.ts ./src/test/ai-accuracy-tests/ai-accuracy-tests.ts",
```

src/participant/prompts/query.ts
alenakhineika: got it! let's try this way 👍

alenakhineika (approved):

PR #830, opened 2024-09-23
feat(participant): route generic prompt by intent and update generic prompt VSCODE-572

Anemy:
VSCODE-572

This pr adds an `intent` prompt. This is used when a user adds a message without a command handler. It tries to route the user's message to a command handler if possible, and if it doesn't route it'll go to the default/generic handler.

There are a few accuracy test adds as well, as well as adding links to the test names for quicker reference in the html.
Screenshot 2024-09-25 at 2 17 40 PM


src/participant/participant.ts
```diff
@@ -229,8 +231,15 @@ export default class ParticipantController {
...
-  async handleGenericRequest(
+  _handleCancelledRequest({
+    context,
+  }: {
+    context: vscode.ChatContext;
+  }): ChatResult {
+    return createCancelledRequestChatResult(context.history);
+  }
+
+  async _handleRoutedGenericRequest(
```
Anemy, line 242: Moving this cancelled request handling and the other abort controller changes into a separate pr.

src/participant/prompts/intent.ts
```diff
@@ -0,0 +1,78 @@
...
+Your task is to help guide a conversation with a user to the correct handler.
+You will be provided a conversation and your task is to determine the intent of the user.
+The intent handlers are:
+- Query
+- Schema
+- Docs
+- Default
+Rules:
+1. Respond only with the intent handler.
+2. Use the "Query" intent handler when the user is asking for code that relates to a specific collection.
```
alenakhineika, line 19: What is the user has not specified a namespace? We don't have information about a specific collection, but it still can be a query generation request.

src/participant/prompts/intent.ts
```diff
@@ -0,0 +1,78 @@
...
+3. Use the "Docs" intent handler when the user is asking a question that involves MongoDB documentation.
```
alenakhineika, line 20: Did you try this out? I am wondering how the model interprets this instruction. Should users explicitly say find in the documentation? Or they can just ask "How to..." questions?

src/participant/prompts/intent.ts
Anemy: That's a good question, and raises a tradeoff we have to take. _Some_ users may ask a question that would result in code for a database, like how can I create a user with specific database permissions or what's the size of X database. In that case the /query handler would be forcing them to choose a collection which doesn't really make sense. I was aiming to make this prompt fairly intentionally selective so we don't restrict the user capabilities. The generic prompt has a lot of the capabilities of the /query resolver. 
Let me know if you think we could word this differently or guide the behavior to the query resolver differently!

src/participant/prompts/intent.ts
Anemy: The model did successfully route a fairly docs related question to the docs handler in my manual testing. Haven't been that extensive to give a good answer on when it does and when it doesn't. We'd need more accuracy tests.

src/participant/prompts/generic.ts
```diff
@@ -1,20 +1,24 @@
...
-You can imagine the schema, collection, and database name.
-Respond in MongoDB shell syntax using the \`\`\`javascript code block syntax.`;
+Your task is to help the user with MongoDB related questions.
+When applicable, you may suggest MongoDB code, queries, and aggregation pipelines that perform their task.
+Rules:
+1. Keep your response concise.
+2. You should suggest code that is performant and correct.
+3. Respond with markdown.
+4. When relevant, provide code in a Markdown code block that begins with \`\`\`javascript and ends with \`\`\`.
+5. Use MongoDB shell syntax for code unless the user requests a specific language.
```
alenakhineika, line 16: This is just a reminder that code blocks should be aligned with `codeBlockIdentifier` from https://github.com/mongodb-js/vscode/pull/835

src/test/ai-accuracy-tests/ai-accuracy-tests.ts
```diff
@@ -34,7 +34,7 @@ type AssertProps = {
 
 type TestCase = {
   testCase: string;
-  type: 'generic' | 'query' | 'namespace';
+  type: 'intent' | 'generic' | 'query' | 'namespace';
```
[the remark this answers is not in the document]
Anemy: This is for the accuracy tests, we don't have any docs tests yet. Once we do we'll add it.

alenakhineika (approved):

nirinchev (approved):

PR #831, opened 2024-09-23
chore(participant): add handling for token cancellation, pass to docs fetch request, remove unused abort controllers

Anemy:
We want to fail early and not ping the model when the chat token's token is cancelled.

This pr:
- Adds more places for us to fail early when the token is cancelled. 
- Adds an abort signal to the docs ai requests. 
- Removes 3 unused abort controllers we had.

### DIFF
--- a/src/participant/docsChatbotAIService.ts
+++ b/src/participant/docsChatbotAIService.ts
@@ -54,7 +54,7 @@ export class DocsChatbotAIService {
     return `${this._serverBaseUri}api/${MONGODB_DOCS_CHATBOT_API_VERSION}${path}`;
   }
 
-  async _fetch({
+  _fetch({
     uri,
     method,
     body,
@@ -67,7 +67,7 @@ export class DocsChatbotAIService {
     body?: string;
     headers?: { [key: string]: string };
   }): Promise<Response> {
-    return await fetch(uri, {
+    return fetch(uri, {
       headers: {
         origin: this._serverBaseUri,
         'User-Agent': `mongodb-vscode/${version}`,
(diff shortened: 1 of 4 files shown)


src/participant/docsChatbotAIService.ts
```diff
@@ -58,29 +58,37 @@ export class DocsChatbotAIService {
...
     headers,
   }: {
     uri: string;
     method: string;
+    signal?: AbortSignal;
     body?: string;
     headers?: { [key: string]: string };
   }): Promise<Response> {
-    return fetch(uri, {
+    return await fetch(uri, {
```
nirinchev, line 70: [nit] We can remove both the async and the await from this function.

src/participant/participant.ts
```diff
@@ -1096,11 +1118,7 @@ export default class ParticipantController {
       vscode.CancellationToken
     ]
   ): Promise<ChatResult> {
-    const [request, context, stream, token] = args;
-    const abortController = new AbortController();
-    token.onCancellationRequested(() => {
-      abortController.abort();
-    });
+    const [request, context, stream] = args;
```
nirinchev, line 1121: Why not decompose token here as well and pass it as a named variable further down rather than as `args[3]`?

src/participant/participant.ts
```diff
@@ -1115,11 +1133,20 @@ export default class ParticipantController {
...
       });
     } catch (error) {
       // If the docs chatbot API is not available, fall back to Copilot’s LLM and include
       // the MongoDB documentation link for users to go to our documentation site directly.
       log.error(error);
+
+      if (args[3].isCancellationRequested) {
+        return this._handleCancelledRequest({
+          context,
+          stream: args[2],
```
nirinchev, line 1146:
```suggestion
          stream,
```

nirinchev (approved):
A few nits, otherwise looks good.

alenakhineika (approved):

PR #834, opened 2024-09-24
chore(chat): Refactor prompt hierarchy

nirinchev:
## Description

This is a bit controversial, so putting it as a draft to get some feedback on the direction. It's preparation work for VSCODE-606 - it unifies all prompts under a common base class, which I'm planning to then extend with capability for counting user/participant message lengths as well as a flag for whether sample docs have been included in the prompt (per [TD](https://docs.google.com/document/d/1PE5flrsfWQ3KPmmJMx0pnqGLdqj0OrU6oviZJPP-O6M/edit#bookmark=id.302kj2mq0ejk). While there are less disruptive ways to achieve this, this one felt the cleanest in terms of separation of concerns and extensibility.

It should also set us up for VSCODE-614 as abstracting the base functionality in a common place would allow us to count the tokens and drop history messages that push us past the model token limit.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/participant/prompts/namespace.ts
```diff
@@ -39,44 +47,33 @@ User: Where is the best hummus in Berlin?
...
 
-  static getUserPrompt(prompt: string): vscode.LanguageModelChatMessage {
-    // eslint-disable-next-line new-cap
-    return vscode.LanguageModelChatMessage.User(prompt);
+  protected getUserPrompt({ request }: NamespacePromptArgs): Promise<string> {
+    return Promise.resolve(request.prompt);
   }
 
-  static buildMessages({
+  async buildMessages({
```
nirinchev, line 56: This one is the only one I couldn't map super cleanly and had to override the base class implementation. I have some ideas for how to restructure the abstraction to support these more complex history/prompt rewrites, but those would be more complicated than the current implementation and didn't feel it's justified to pursue until we have more than one use case.

src/participant/prompts/namespace.ts
```diff
@@ -39,44 +47,33 @@ User: Where is the best hummus in Berlin?
...
-      const intent = (previousResponse?.result as ChatResult).metadata.intent;
+      const intent = (previousResponse?.result as ChatResult)?.metadata.intent;
       if (intent === 'askToConnect') {
         // Go through the history in reverse order to find the last user message.
         for (let i = historyMessages.length - 1; i >= 0; i--) {
           if (
             historyMessages[i].role === vscode.LanguageModelChatMessageRole.User
           ) {
-            userPrompt = historyMessages[i].content;
+            request.prompt = historyMessages[i].content;
```
Anemy, line 76: This is going to mutate the argument to the method. Can we avoid that? I find mutation to be a slippery slope into having things changing unexpectedly.  Maybe we can create a new prompt variable in this method that we then pass to the user prompt function.

src/participant/prompts/promptBase.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  // When passing the history to the model we only want contextual messages
+  // to be passed. This function parses through the history and returns
+  // the messages that are valuable to keep.
+  // eslint-disable-next-line complexity
+  protected getHistoryMessages({
+    connectionNames,
+    context,
+  }: {
+    connectionNames?: string[]; // Used to scrape the connecting messages from the history.
```
Anemy, line 34:
```suggestion
    connectionNames: string[]; // Used to scrape the connecting messages from the history.
```
[nit] Are there any cases where we don't send this? Maybe we can make the type stricter.

src/participant/prompts/namespace.ts
Anemy: We'll revisit this function and also how the other prompt message building functions handle namespace and connect messages/prompts in:
https://jira.mongodb.org/browse/VSCODE-611 
I think the other prompts would benefit from a similar prompt change when the last message was selecting a namespace.

Anemy (approved):
lgtm, two some suggestions, I'd like if we can avoid the argument mutation one, not blockers.

src/participant/prompts/schema.ts
```diff
@@ -1,85 +1,42 @@
...
+  amountOfDocumentsSampled: number;
+  connectionNames: string[];
+}
+
+export class SchemaPrompt extends PromptBase<SchemaPromptArgs> {
+  getAssistantPrompt({ amountOfDocumentsSampled }: SchemaPromptArgs): string {
+    return `You are a senior engineer who describes the schema of documents in a MongoDB database.
 The schema is generated from a sample of documents in the user's collection.
-You must follows these rules.
+You must follow these rules.
```
Anemy, line 21: ❤️

src/participant/prompts/namespace.ts
nirinchev: Agreed - I was following up on this PR with telemetry stuff and noticed history scraping could be applied to all messages, not just namespace.

src/participant/sampleDocuments.ts
```diff
@@ -59,11 +58,14 @@ export async function getStringifiedSampleDocuments({
   }
 
   const stringifiedDocuments = toJSString(additionToPrompt);
-  promptInputTokens = await model.countTokens(prompt + stringifiedDocuments);
+
+  // TODO: model.countTokens will sometimes return undefined - at least in tests. We should investigate why.
```
Anemy, line 62: This was something Alena ran into, `The model returned by vscode.lm.selectChatModels is always undefined in tests`. So there's no model to count the tokens.

Anemy (approved):
lgtm! Nice update to share the connect message prompt from the rest of the prompts.

PR #835, opened 2024-09-25
fix(chat): update response handling to stream and inline code block parsing VSCODE-620

Anemy:
VSCODE-620

These changes make all of the model responses feel a faster. We're now streaming them as they come in instead of building one string out of the response and then streaming that.
Also adds support for multiple code blocks. We now inline the code actions as the code comes in. Hopefully it's written in a way we can change that out as we update to a different code block identifier in an upcoming ticket.

Screenshot 2024-09-25 at 5 49 38 PM


src/participant/participant.ts
```diff
@@ -170,48 +162,113 @@ export default class ParticipantController {
...
+        this._streamCodeBlockActions({ runnableContent: content, stream });
+      },
+      inputIterable: chatResponse.text,
+      identifier: codeBlockIdentifier,
+    });
+  }
+
+  // This will stream all of the response content and create a string from it.
+  // It should only be used when the entire response is needed at one time.
+  async getChatResponseContent({
```
alenakhineika, line 255: Do we use this function anywhere? Can't find.

src/participant/streamParsing.ts
```diff
@@ -0,0 +1,95 @@
...
+    let lastIndex = 0;
+    let match;
+    while ((match = regex.exec(contentSinceLastIdentifier)) !== null) {
+      const endIndex = regex.lastIndex;
+
+      // Stream content up to the end of the identifier.
+      const contentToStream = contentSinceLastIdentifier.slice(
+        lastIndex,
+        endIndex
+      );
```
alenakhineika, line 45: Do you know if the code block can be mixed with the normal text withing one fragment? Or maybe the code block always takes the whole fragment.

src/participant/participant.ts
Anemy: It's used for the namespace prompt currently, and will also be used with the intent prompt.

src/participant/participant.ts
```diff
@@ -170,48 +162,113 @@ export default class ParticipantController {
...
+      token,
+    });
+
+    await processStreamWithInsertionsOnIdentifier({
+      processStreamFragment: (fragment: string) => {
+        stream.markdown(fragment);
+      },
+      onIdentifierStreamed: (content: string) => {
+        this._streamCodeBlockActions({ runnableContent: content, stream });
+      },
```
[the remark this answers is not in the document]
Anemy: I wanted to build it with some flexibility so if we want to add additional handlers for other languages it's easy.

src/participant/streamParsing.ts
Anemy: There are tests for both cases, code block all in one fragment, code block and identifiers split between fragments. Typically the fragments are 1-3 characters.

alenakhineika (approved):

src/participant/participant.ts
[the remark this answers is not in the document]
nirinchev: I like that it's isolated and potentially reusable. Processing the content of the stream while still displaying it to the user seems like a common-enough scenario that justifies making it a bit more general-purpose.

nirinchev (approved):
After spending some time with streamParsing.ts, I am 98% convinced the algorithm is correct, though it seems a bit more complex than strictly necessary. I don't think we need to refactor it and the tests are solid, so we can revisit at some point in the future. Similarly with the participant chat functions - we now have a handful of those, so I'd like to move them to a separate wrapper class in the future.

PR #836, opened 2024-09-26
chore(participant) Wire up telemetry for user prompts VSCODE-606

nirinchev:
## Description
Adds some telemetry stats when constructing prompts. Part of VSCODE-575.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/telemetry/telemetryService.ts
```diff
@@ -108,6 +108,13 @@ type ParticipantResponseFailedProperties = {
   error_name: ParticipantErrorTypes;
 };
 
+export type ParticipantPromptProperties = {
+  command: string;
+  user_input_length: number;
+  total_message_length: number;
+  has_sample_documents: boolean;
```
Anemy, line 115: Should we also track the amount of messages in the chat? Could be useful to see how much people start new chats vs keep existing ones going.

src/participant/participant.ts
```diff
@@ -164,10 +165,10 @@ export default class ParticipantController {
   }
 
   async _getChatResponse({
-    messages,
+    prompt,
     token,
   }: {
-    messages: vscode.LanguageModelChatMessage[];
+    prompt: PromptResult;
```
Anemy, line 171: [nit] No action needed here, mostly thinking out loud. I wonder if we're over using the term `prompt` here. (or maybe it's in other places we could change it).
Is there a way we can name this to get the meaning quicker on first read? Maybe `ChatRequest`? `ChatCompletionRequest`?

src/participant/participant.ts
```diff
@@ -176,20 +177,22 @@ export default class ParticipantController {
       throw new Error('Copilot model not found');
     }
 
-    return await model.sendRequest(messages, {}, token);
+    this._telemetryService.trackCopilotParticipantPrompt(prompt.stats);
```
Anemy, line 180: This is going to track for our metadata requests - `namespace` and `intent`. Should we want track those separately, or least add an indicator for them?

Anemy (commented):
Nice test adds, a few questions on what we're tracking, I think we want to think about how we can distinguish the metadata prompt chaining requests from others.

src/participant/participant.ts
alenakhineika: Also, it is a bit hard to get while reading, that prompt is an input argument but has PromptResult type. Like input is a result.

src/participant/participant.ts
nirinchev: I've added `internal_purpose` to differentiate namespace and intent requests.

src/participant/participant.ts
nirinchev: Agree - wasn't quite happy with the name myself. I changed it to `ModelInput` - still don't think it's ideal, but I feel it's more descriptive than before.

src/telemetry/telemetryService.ts
nirinchev: Added `history_length` for this. Note that this only tracks the number of history messages and not the length of the messages themselves.

Anemy (approved):

PR #837, opened 2024-09-26
VSCODE-639: Replace regex fragment matching with streaming KMP

nirinchev:
## Description
I got [nerd-sniped](https://xkcd.com/356/) by https://github.com/mongodb-js/vscode/pull/835, and wanted to try out a regex-less implementation of the identifier matching algorithm. I've implemented a stateful streaming Knuth-Morris-Pratt algorithm for the identifier matching and then a stateful matcher for the content. Those could have been trivially combined, but I opted to keep them separate to 1) keep the concerns more cleanly separate and 2) to have the building blocks available in case we need to do a simple substring search in the stream without extracting some content between the identifiers.

While I think this approach is slightly more readable and should (theoretically) be more performant, I don't believe we should try and get it in before the code freeze as it's not fixing an actual bug.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/participant/streamParsing.ts
+++ b/src/participant/streamParsing.ts
@@ -114,22 +114,10 @@ class FragmentMatcher {
       // We haven't matched the start identifier yet, so try and do that
       const startIndex = this._startMatcher.match(fragment);
       if (startIndex !== -1) {
-        let endIndex = this._endMatcher.match(fragment.slice(startIndex));
-        if (endIndex !== -1) {
-          // This is the case where both the start and the end identifiers are contained in the same fragment.
-          // In this case, we emit the content between the two identifiers and continue processing the rest of the fragment.
-
-          // endIndex is relative to the slice, so we need to add startIndex to it.
-          endIndex = startIndex + endIndex;
-
-          this._matchedContent = fragment.slice(startIndex, endIndex);
-          this._contentMatched();
-          this.process(fragment.slice(endIndex));
-        } else {
-          // If we only matched the start, we add the partial fragment to the matched content and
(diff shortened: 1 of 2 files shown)


alenakhineika:
Love it!! Looking forward to this work being finished and merged 🔥

src/participant/streamParsing.ts
```diff
@@ -22,74 +162,13 @@ export async function processStreamWithIdentifiers({
...
-            contentSinceLastIdentifier.slice(processUpTo);
-        }
-      }
-    }
-  }
-
-  // Finish up anything not streamed yet.
-  if (contentSinceLastIdentifier.length > 0) {
-    processStreamFragment(contentSinceLastIdentifier);
+  for await (const fragment of inputIterable) {
```
Anemy, line 170: One change in behavior that will happen is when a fragment is very large, and could contain two code blocks. It would now print all of the fragment before it adds the actions. 
This won't impact our use of this since the fragments are pretty much always 2-4 characters.

Anemy (approved):
lgtm! Nice writeup. Called out one behavioral change, but I don't think we need to make any changes to accommodate. This is already improving things by making the streaming not block until the whole code block is parsed.

src/participant/streamParsing.ts
nirinchev: Thanks for spotting that. The fix for it was simple enough, so I added it anyway just to be on the safe side.

PR #840, opened 2024-09-27
chore: update vsce dep

Anemy:


package.json
```diff
@@ -1243,7 +1243,7 @@
     "@typescript-eslint/eslint-plugin": "^5.62.0",
     "@typescript-eslint/parser": "^5.62.0",
     "@vscode/test-electron": "^2.4.1",
-    "@vscode/vsce": "^2.31.1",
+    "@vscode/vsce": "^3.1.0",
```
alenakhineika, line 1246: Interesting that it doesn't require to bump a vscode engine in package.json. Usually, they are dependent.

alenakhineika (approved):

PR #843, opened 2024-09-27
fix(chat): add message content to blank message request VSCODE-626

Anemy:
VSCODE-626

When the message sent to the ai is empty, like how folks will use /schema, we would receive an error: `unhandled status from server, please report the following request id`
This only happens in non-insiders VSCode.
It looks to be fixed in insiders, but since we do want to release the extension now we should work around it. We're adding a not-empty message in those cases so it doesn't error. This is a bit of a hack, I think we'll improve this in a follow up (or remove it). We probably want the check closer to the actual model request.

Also adds a few logs around when we're making requests and get a response.


src/participant/participant.ts
```diff
@@ -137,6 +138,13 @@ export default class ParticipantController {
       errorName = ParticipantErrorTypes.OTHER;
     }
 
+    log.error('Participant encountered an error', {
+      command,
+      err,
```
alenakhineika, line 143: Is it an original error? We should ensure that we don't log sensitive information that might come from a model with the error.

src/participant/participant.ts
```diff
@@ -176,7 +184,21 @@ export default class ParticipantController {
       throw new Error('Copilot model not found');
     }
 
-    return await model.sendRequest(messages, {}, token);
+    log.info('Sending request to model', {
+      truncatedMessages: messages.map(
+        (message: vscode.LanguageModelChatMessage) =>
+          util.inspect({
+            role: message.role,
+            content: message.content.slice(0, 50),
```
alenakhineika, line 192: Same here, we can't log requests.

src/participant/participant.ts
```diff
@@ -239,16 +261,25 @@ export default class ParticipantController {
...
       },
       onStreamIdentifier: (content: string) => {
+        codeBlocksInResponse++;
         this._streamCodeBlockActions({ runnableContent: content, stream });
       },
       inputIterable: chatResponse.text,
       identifier: codeBlockIdentifier,
     });
+
+    log.info('Streamed response to chat', {
```
alenakhineika, line 279: The same about responses.

src/participant/participant.ts
```diff
@@ -707,14 +742,39 @@ export default class ParticipantController {
...
+    // skip the request to the model. This would happen with /schema.
+    if (Prompts.doMessagesContainUserInput(messagesWithNamespace)) {
+      // VSCODE-626: When there's an empty message sent to the ai model,
+      // it currently errors (not on insiders, only main VSCode).
+      // Here we're defaulting to have some content as a workaround.
+      // TODO: Remove this when the issue is fixed.
+      messagesWithNamespace[messagesWithNamespace.length - 1].content =
+        messagesWithNamespace[
+          messagesWithNamespace.length - 1
+        ].content.trim() || 'see previous messages';
```
alenakhineika, line 767: How this can happen that we have user input but content is empty? And what are those 'see previous messages'? Which kind of messages does it refer to?

Anemy: An empty user message can happen when previous messages in the history have user input, then the most recent message is either empty, or the most recent user message is an empty one that is pulled from chat history. We get rid of some messages in the message building, any connection messages for instance, so this could be something from earlier in the chat history.

src/participant/participant.ts
```diff
@@ -376,6 +405,10 @@ export default class ParticipantController {
       token,
     });
 
+    log.info('Received intent response from model', {
+      responseContent,
```
alenakhineika, line 409: One more log with the response.

alenakhineika (approved):

PR #844, opened 2024-09-27
fix(chat): avoid passing file path when its not used in playground run VSCODE-625

Anemy:
VSCODE-625

Did a bit of drive by renaming `query` -> `code`. There's a good amount of code we could cleanup in the playground controller around running playgrounds, however I'm going to avoid doing any more changes there as we want to code freeze and a lot of them are already done in Alena's pr on exporting code to playground.


src/editors/playgroundController.ts
```diff
@@ -572,11 +586,18 @@ export default class PlaygroundController {
...
 
-  async evaluateParticipantQuery(text: string): Promise<boolean> {
+  async evaluateParticipantCode(codeToEvaluate: string): Promise<boolean> {
     const shouldConfirmRunCopilotCode = vscode.workspace
       .getConfiguration('mdb')
       .get('confirmRunCopilotCode');
 
+    if (!this._connectionController.isCurrentlyConnected()) {
+      // TODO(VSCODE-618): Prompt user to connect when clicked.
+      void vscode.window.showErrorMessage(connectBeforeRunningMessage);
```
Anemy, line 596: Another drive-by, adding an error message when trying to run when not connected, and a todo for the changes in VSCODE-618.

src/editors/playgroundController.ts
```diff
@@ -455,7 +466,7 @@ export default class PlaygroundController {
       result = await this._languageServerController.evaluate({
         codeToEvaluate,
         connectionId,
-        filePath: vscode.window.activeTextEditor?.document.uri.fsPath,
```
alenakhineika, line 458: I think I am missing why we want to pass it from _evaluatePlayground instead of using it directly here?

PR #845, opened 2024-09-27
chore(chat): Report response telemetry VSCODE-603

nirinchev:
## Description
Final bit from VSCODE-575. This doesn't follow exactly the design as I wasn't sure how we can track `round_trips_number`. If we decide to drop this requirement, we should update `ParticipantResponseProperties`.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/participant/participant.ts
+++ b/src/participant/participant.ts
@@ -1303,7 +1303,7 @@ export default class ParticipantController {
     this._telemetryService.trackCopilotParticipantResponse({
       command: 'docs/copilot',
       has_cta: true,
-      found_namespace: true,
+      found_namespace: false,
       has_runnable_content: hasCodeBlock,
       output_length: outputLength,
     });
--- a/src/participant/participant.ts
+++ b/src/participant/participant.ts
@@ -1301,7 +1301,7 @@ export default class ParticipantController {
     });
 
     this._telemetryService.trackCopilotParticipantResponse({
-      command: 'docs',
+      command: 'docs/copilot',
       has_cta: true,
       found_namespace: true,
       has_runnable_content: hasCodeBlock,
@@ -1364,7 +1364,7 @@ export default class ParticipantController {
       }
 
       this._telemetryService.trackCopilotParticipantResponse({
-        command: 'docs',
+        command: 'docs/chatbot',
         has_cta: !!docsResult.responseReferences,
         found_namespace: false,
         has_runnable_content: false,
--- a/src/participant/participant.ts
+++ b/src/participant/participant.ts
@@ -190,14 +190,21 @@ export default class ParticipantController {
     modelInput: ModelInput;
     stream: vscode.ChatResponseStream;
     token: vscode.CancellationToken;
-  }): Promise<void> {
+  }): Promise<{ outputLength: number }> {
     const chatResponse = await this._getChatResponse({
       modelInput,
       token,
     });
+
+    let length = 0;
     for await (const fragment of chatResponse.text) {
       stream.markdown(fragment);
+      length += fragment.length;
     }
+
+    return {
+      outputLength: length,
+    };
   }
 
   _streamCodeBlockActions({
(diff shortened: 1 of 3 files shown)


src/participant/participant.ts
```diff
@@ -236,22 +243,34 @@ export default class ParticipantController {
     modelInput: ModelInput;
     stream: vscode.ChatResponseStream;
     token: vscode.CancellationToken;
-  }): Promise<void> {
+  }): Promise<{
+    outputLength: number;
+    hasCodeBlock: boolean;
```
Anemy, line 248: Should we do codeBlockCount ? Or boolean fine? Not a blocker

src/participant/participant.ts
```diff
@@ -1252,6 +1299,14 @@ export default class ParticipantController {
       },
       stream,
     });
+
+    this._telemetryService.trackCopilotParticipantResponse({
+      command: 'docs/copilot',
+      has_cta: true,
+      found_namespace: true,
```
Anemy, line 1306:
```suggestion
      found_namespace: false,
```
Should we have `undefined` for these instead? Since we don't ask the user for it and we aren't looking for it for these requests (and generic).

src/telemetry/telemetryService.ts
```diff
@@ -119,6 +119,15 @@ export type ParticipantPromptProperties = {
   internal_purpose: InternalPromptPurpose;
 };
 
+export type ParticipantResponseProperties = {
+  command: string;
+  has_cta: boolean;
+  has_runnable_content: boolean;
+  found_namespace: boolean;
+  round_trips_number?: number; // TODO: how do we track this?
```
Anemy, line 127: Typically when adding a `TODO` in the code we have a ticket number with it, like `TODO(VSCODE-123): ...` that way we know either it's going to be resolved at some point, or if it's something like update a dependency, we can look at that ticket and see it's resolved and remove the TODO. 
Let's remove this until we decide if we want it or not?
```suggestion
```

src/participant/participant.ts
```diff
@@ -1160,10 +1197,19 @@ export default class ParticipantController {
...
+    const { hasCodeBlock, outputLength } =
+      await this.streamChatResponseContentWithCodeActions({
+        modelInput,
+        stream,
+        token,
+      });
+
+    this._telemetryService.trackCopilotParticipantResponse({
+      command: 'query',
+      has_cta: false,
```
Anemy, line 1209: Should this be true when we have a code block? There's the open in playground and run actions.

src/participant/participant.ts
```diff
@@ -287,10 +306,19 @@ export default class ParticipantController {
...
+    const { hasCodeBlock, outputLength } =
+      await this.streamChatResponseContentWithCodeActions({
+        modelInput,
+        token,
+        stream,
+      });
+
+    this._telemetryService.trackCopilotParticipantResponse({
+      command: 'generic',
+      has_cta: false,
```
Anemy, line 318: Same q as below, should this be true when we have a code block? There's the open in playground and run actions.

Anemy (commented):
Looks good, a couple questions on event types and when we report them

src/participant/participant.ts
nirinchev: Fine either way - I went with a bool because that was the type in the design.

src/telemetry/telemetryService.ts
nirinchev: Yeah, I explicitly added it without a ticket to force us to make a decision on whether we want it or not during review. Seems like @GaurabAryal is in favor of dropping it for the initial release and revisiting after.

src/participant/participant.ts
nirinchev: True, but then it seems like `has_runnable_content` and `has_cta` would be duplicated quite often. When reading the design, I interpreted `has_runnable_content` to be the flag indicating if we have a code block, while `has_cta` would contain any other cta content - such as docs links/open schema button. I don't have a strong opinion and will let you, @alenakhineika and @GaurabAryal to make the call.

src/participant/participant.ts
nirinchev: You're right - not really sure why I used `true` here.

src/participant/participant.ts
Anemy: Makes sense, I'm cool with anything here, thinking about it more it won't end up being confusing or losing any data whichever way we end up going so I'll resolve.

Anemy (approved):

PR #861, opened 2024-10-29
fix: filter long and invalid prompts in future messages VSCODE-614

gagik:
## Description
When someone sends a long message (or some message which could cause it to be i.e. filtered by the responsible AI service), we keep resending this in future prompts, causing the chat to be broken for the whole session. This PR excludes them from the results.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/telemetry/telemetryService.ts
```diff
@@ -193,14 +194,6 @@ export enum TelemetryEventTypes {
   PARTICIPANT_RESPONSE_GENERATED = 'Participant Response Generated',
 }
 
-export enum ParticipantErrorTypes {
```
gagik, line 196: I figured it'd be better to get the error types out of the scope of telemetry since it might be useful to handle specific errors.

src/participant/prompts/promptBase.ts
```diff
@@ -188,6 +189,14 @@ export abstract class PromptBase<TArgs extends PromptArgsBase> {
       }
 
       if (historyItem instanceof vscode.ChatResponseTurn) {
+        if (
+          historyItem.result.errorDetails?.message ===
+          ParticipantErrorTypes.FILTERED
+        ) {
+          messages.pop();
```
gagik, line 196: I am not a huge fan of this way of adding and then removing the message retroactively but can't think of a nicer way. 

We could be less strict with what kind of errors lead to it, right now I decided to just tackle the filtered error but maybe we don't want other error-causing messages either.

himanshusinghs: Is this message then completely gone from the history and user won't be able to see this in their UI? If so, then maybe we can also consider trimming the message short instead of completely removing it? That won't leave the user hanging wondering how their last messages are not showing up.

himanshusinghs: Completely ignorable if that's not what is happening (I did not test it locally, sorry)

alenakhineika:
I am thinking, what if instead of removing the long message from the history we should introduce a max prompt size, similar like we do in compass, and inform users after submission, that their prompt is too long. We have even mentioned doing something like this in tech design in the "Building Chat Participant" section. @Anemy @GaurabAryal do you maybe have some thoughts?

src/participant/prompts/promptBase.ts
gagik: Nope the message will still be around in the UI, this is just a matter of what we use as our message context

gagik:
> I am thinking, what if instead of removing the long message from the history we should introduce a max prompt size, similar like we do in compass, and inform users after submission, that their prompt is too long. We have even mentioned doing something like this in tech design in the "Building Chat Participant" section. @Anemy @GaurabAryal do you maybe have some thoughts?

That'd be good for that specific case, which should be common enough to be worth handling.

Beyond that I assume we may still end up in scenario where i.e. maybe the user has some bad words which are filtered by the Responsible AI Service (which I think does do filtering beyond just prompt size), in which case we'd probably want to dismiss its contents in the following messages; otherwise again we'll leave the user stuck for the whole chat just because they made a mistake.

nirinchev:
I agree - we probably want to do both. Note that the max prompt size concerns both the user prompt, but also the history, so if a user has a very long conversation, they may end up exceeding the limit, even if their last message was a short one. That means we'll need to build out a system for dropping the oldest history messages.

src/types/participantErrorTypes.ts
```diff
@@ -0,0 +1,7 @@
+export enum ParticipantErrorTypes {
+  CHAT_MODEL_OFF_TOPIC = 'Chat Model Off Topic',
+  INVALID_PROMPT = 'Invalid Prompt',
+  FILTERED = 'Filtered by Responsible AI Service',
+  OTHER = 'Other',
+  DOCS_CHATBOT_API = 'Docs Chatbot API Issue',
+}
```
Anemy, line 7: nit: I think I'd prefer for participant related types to be in the participant folder. Curious what other folks think here. Fine as is too.

Anemy (approved):
lgtm, nice test add

src/types/participantErrorTypes.ts
gagik: good point. we have some of that pattern already with some types in the `participant/constants.ts` file so I'll move it there.

PR #863, opened 2024-11-01
feat(chat): automatically pick the database & collection if there exists only one VSCODE-610

gagik:
[VSCODE-610](https://jira.mongodb.org/browse/VSCODE-610) mentions also behavior for cases where is 0, I will make a follow up PR as this one is getting too large. This handles the case where there is only 1 database or collection, so we'd want to pick that instead of prompting the user.

I was originally planning to split them, but it ended up being nicer organization wise to combine collection and database picking logic as a lot of it is mirrored and uses some helper methods.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/participant/participant.ts
```diff
@@ -1084,6 +1144,34 @@ export default class ParticipantController {
...
+          collectionName: undefined,
+          history: context.history,
+        });
+      }
+
+      // Save the collection name in the metadata.
+      const chatId = ChatMetadataStore.getChatIdFromHistoryOrNewChatId(
+        context.history
+      );
+      this._chatMetadataStore.setChatMetadata(chatId, {
```
gagik, line 1169: Does this make sense to do? If I understand right this means this will cache it but I haven't tested it yet, will soon.

src/participant/participant.ts
```diff
@@ -811,50 +784,205 @@ export default class ParticipantController {
...
       });
-    } else if (!collectionName) {
-      stream.markdown(
-        `Which collection would you like to use within ${databaseName}?\n\n`
+
+      // If the database name could not get automatically selected,
+      // then the user has been prompted for it instead.
+      if (!databaseName) {
+        return { databaseName, collectionName };
+      }
```
nirinchev, line 946: I'm not sure I understand how this works. As far as I can follow this, if I don't have any databases, I'd get the "No databases found" error and no prompt to create/select a new one.

src/participant/participant.ts
```diff
@@ -811,50 +784,205 @@ export default class ParticipantController {
...
+    
+    if (collections.length === 1) {
+      return collections[0].name;
+    }
+
+    stream.markdown(
+      vscode.l10n.t(
+        `Which collection would you like to use within ${databaseName}?\n\n`
+      )
+    );
```
nirinchev, line 866: If there are no collections in the database, would the user have to type in the collection name in the chat themselves?

src/participant/participant.ts
```diff
@@ -811,50 +784,205 @@ export default class ParticipantController {
...
+      // Save the database name in the metadata.
+      const chatId = ChatMetadataStore.getChatIdFromHistoryOrNewChatId(
+        context.history
       );
-      await this.renderCollectionsTree({
-        command,
+      this._chatMetadataStore.setChatMetadata(chatId, {
+        ...this._chatMetadataStore.getChatMetadata(chatId),
         databaseName,
+      });
```
nirinchev, line 955: [nit] This is a bit repetitive - not sure if there are other instances beyond the db and collection names, but I'd consider extracting a method in ChatMetadataStore like:

```ts
patchChatMetadata(context: vscode.ChatContext, patchedMetadata: Partial<ChatMetadata>) {
  const chatId = getChatIdFromHistoryOrNewChatId(
    context.history
  );

  this.setChatMetadata(chatId, {
    ...this.getChatMetadata(chatId),
    ...patchedMetadata
  });
}
```

src/test/suite/participant/participant.test.ts
```diff
@@ -1507,6 +1560,152 @@ suite('Participant Controller Test Suite', function () {
               'see previous messages'
             );
           });
+
+          suite('with an empty database name', function () {
```
nirinchev, line 1564: These tests seem very similar to the ones we have for `/query` - can we parameterize them somehow rather than copy-paste?

nirinchev (commented):
Looks reasonable - I have mostly code style comments/questions.

src/participant/participant.ts
gagik: yeah. when they type nothing afterwards we actually send a different text `Please select a collection by either clicking on an item in the list or typing the name manually in the chat` which I feel like is a better text for this in general, maybe we could just use that?

src/participant/participant.ts
gagik: The comment was not describing everything so I updated that. databaseName ends up as undefined either because we don't have it set yet (so the user has been prompted for it) or if there's an error. It's handled by the `_getOrAskFor...` function.

src/test/suite/participant/participant.test.ts
```diff
@@ -1026,7 +1033,7 @@ suite('Participant Controller Test Suite', function () {
             const chatResult = await invokeChatHandler(chatRequestMock);
             const askForDBMessage = chatStreamStub.markdown.getCall(0).args[0];
             expect(askForDBMessage).to.include(
-              'What is the name of the database you would like this query to run against?'
+              'Which database would you like this query to run against? Select one by either clicking on an item in the list or typing the name manually in the chat.\n\n'
```
gagik, line 1036: @nirinchev @Anemy I combined the message used by `handleEmptyNamespaceMessage` and regular namespace prompting. It's nice to mention the option to type it manually right away and just repeat the same exact prompt if they send an empty message, that also lets us re-use the helper.

PR #864, opened 2024-11-04
feat(chat): show errors if there are issues when getting databases & collections to pick VSCODE-610

gagik:
Adds better handling of 0 collections/databases and errors to https://github.com/mongodb-js/vscode/pull/863. Instead of silently quitting, it shows the error.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/participant/participant.ts
```diff
@@ -853,26 +853,37 @@ export default class ParticipantController {
...
-    if (collections !== undefined) {
-      if (collections.length === 1) {
-        return collections[0].name;
-      }
-
+    if (collections === undefined) {
       stream.markdown(
-        `Which collection would you like to use within ${databaseName}?\n\n`
+        vscode.l10n.t(
+          `An error occurred when getting the collections from the database ${databaseName}.`
```
gagik, line 859: do we want to send telemetrics for when this occurs?

src/test/suite/participant/participant.test.ts
```diff
@@ -1321,6 +1321,49 @@ suite('Participant Controller Test Suite', function () {
               sinon.restore();
             });
 
+            test('shows an error if something goes wrong with getting databases', async function () {
```
gagik, line 1324: there is a lot of copy-pasting here with these tests and little change other than the command string, I could turn this into a test matrix but it'd break the structure pattern of tests.

src/participant/participant.ts
```diff
@@ -853,26 +853,37 @@ export default class ParticipantController {
...
+      );
+      return;
+    }
+    if (collections.length === 1) {
+      return collections[0].name;
     }
 
-    return;
+    stream.markdown(
+      `Which collection would you like to use within ${databaseName}?\n\n`
```
nirinchev, line 877: [nit] Should this be wrapped in `vscode.l10n.t`?

src/participant/participant.ts
```diff
@@ -890,8 +901,14 @@ export default class ParticipantController {
   }): Promise<string | undefined> {
     const databases = await this._getDatabases({ stream });
 
-    if (databases === undefined || databases.length === 0) {
-      log.error('No databases found');
+    if (databases === undefined) {
+      stream.markdown(
+        vscode.l10n.t('An error occurred when getting the databases.')
```
Anemy, line 906: Can we surface the error to users here? Something like permission issues is more transparent then.

gagik: good point, not sure how to word it right. could something like `"An error occurred when getting the databases. You may lack the necessary permissions.` work or we can be even more explicit?

Anemy (approved):
lgtm! Nice error handling ux improvements

PR #866, opened 2024-11-05
feat(chat): filter namespace messages from history if it exists in metadata VSCODE-611

gagik:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/participant/prompts/promptBase.ts
+++ b/src/participant/prompts/promptBase.ts
@@ -13,6 +13,8 @@ export interface PromptArgsBase {
   };
   context?: vscode.ChatContext;
   connectionNames?: string[];
+  databaseName?: string;
+  collectionName?: string;
 }
 
 export interface UserPromptResponse {
@@ -163,9 +165,13 @@ export abstract class PromptBase<TArgs extends PromptArgsBase> {
   protected getHistoryMessages({
     connectionNames,
     context,
+    databaseName,
+    collectionName,
   }: {
     connectionNames?: string[]; // Used to scrape the connecting messages from the history.
     context?: vscode.ChatContext;
+    databaseName?: string;
+    collectionName?: string;
   }): vscode.LanguageModelChatMessage[] {
     const messages: vscode.LanguageModelChatMessage[] = [];
 
@@ -173,6 +179,10 @@ export abstract class PromptBase<TArgs extends PromptArgsBase> {
       return [];
     }
 
+    let previousItem:
+      | vscode.ChatRequestTurn
+      | vscode.ChatResponseTurn
+      | undefined = undefined;
     for (const historyItem of context.history) {
       if (historyItem instanceof vscode.ChatRequestTurn) {
         if (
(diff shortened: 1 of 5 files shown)


src/participant/participantErrorTypes.ts
gagik, line 1: My bad, I ended up switching between branches a lot and moved the error types into the error suite instead of the actual participant folder 🙈 Fixing it here.

alenakhineika (approved):

src/participant/prompts/promptBase.ts
```diff
@@ -163,16 +165,27 @@ export abstract class PromptBase<TArgs extends PromptArgsBase> {
   protected getHistoryMessages({
```
Anemy, line 165: This function is starting to get a bit long and hard to follow all of the things happening in it. 
It already has the `// eslint-disable-next-line complexity` which is usually an indicator that we should break it into a few functions, even if it comes with a slightly hit to performance (like running through the messages multiple times).
Should we do that now? Break this function into multiple parts where each is doing a certain thing? That'll also make it more easily unit testable.

src/test/suite/participant/participant.test.ts
```diff
@@ -2176,6 +2176,142 @@ Schema:
...
+
+      const chatRequestMock = {
+        prompt: 'localhost',
+        command: 'query',
+      };
+
+      beforeEach(function () {
+        chatContextStub = {
+          history: [
+            Object.assign(Object.create(vscode.ChatRequestTurn.prototype), {
```
Anemy, line 2193: nit: I think I probably added this to some other tests a while back. Maybe we can make a helper function like `createVSCodeChatRequestTurn` that these can use. Same with `ChatResponseTurn`. To make us avoid re-writing some of the boilerplate each time.

Anemy (commented):
Nice, a couple code quality suggestions.

src/participant/prompts/promptBase.ts
alenakhineika: More like a question about this area in code and maybe a bit of a request :) if you will refactor this code, could we split the history into something like `getUserHistoryMessages` and `getAssistantHistoryMessages`. When testing this functionality I found it difficult to print the history content because it is already wrapped into the `vscode.LanguageModelChatMessage` format when it is returned here: https://github.com/mongodb-js/vscode/blob/main/src/participant/prompts/promptBase.ts#L130

Maybe, it could be something like:

```
const messages = [
  vscode.LanguageModelChatMessage.Assistant(this.getAssistantPrompt(args)),
  vscode.LanguageModelChatMessage.Assistant(this.getAssistantHistoryMessages()),
  vscode.LanguageModelChatMessage.User(this.getUserHistoryMessages()),
  vscode.LanguageModelChatMessage.User(prompt),
];
```

Or something like that. The idea here is that we have some unformatted string value to print to see what message we send to the model.

gagik: I am also dealing changing getHistoryMessages in [VSCODE-632](https://jira.mongodb.org/browse/VSCODE-632) so I think it'll be best to deal with some refactoring for it there.

gagik:
For the sake of keeping this simple to review, going to merge and do any greater potential refactoring work in the PR for [VSCODE-632](https://jira.mongodb.org/browse/VSCODE-632).

PR #870, opened 2024-11-08
feat(copilot): export to Language with Copilot VSCODE-573

alenakhineika:
## Description

- Export to language from a playground (Shell) to a target language using the VSCode Copilot model.
- Use only the Include/exclude driver syntax code lenses.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/participant/prompts/exportToLanguage.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export class ExportToLanguagePrompt extends PromptBase<ExportToLanguagePromptArgs> {
+  protected getAssistantPrompt({
+    language,
+  }: ExportToLanguagePromptArgs): string {
+    return `You are a MongoDB expert.
+Your task is to convert a MongoDB playground script to the ${language} language.
+Take a user prompt as an input string and translate it to the target language.
+If the user specified to include import statements, add required imports at the beginning of the transpiled code.
+If the user specified to not include import statements, do not add any imports at the beginning of the transpiled code.
```
Anemy, line 17: How do you feel about conditionally showing only one of these to keep the prompt shorter with fewer instructions? 

If the user specified to include import statements then we have a line saying include import statements, and if they set it not to then we say don't include import statements.

Same with the driver syntax.

alenakhineika: I actually removed the import statements switch and forgot to update the prompt. Great that you pointed to this, thx! The model includes import statements by default when you include the driver syntax. I will update the text!

alenakhineika: Actually, it doesn't perform well without specifying booth options in the assistant message. The model then adds all helpers anyway, even if you say not to do it. It looks like when you explicitly list both available options and describe the expected behavior, only then the model properly gets the instruction. I refactored it a bit to combine the driver syntax and import statements cases.

src/participant/participant.ts
```diff
@@ -1692,6 +1692,54 @@ export default class ParticipantController {
...
+          new Promise<null>((resolve) =>
+            token.onCancellationRequested(() => {
+              resolve(null);
+            })
+          ),
+        ]);
+
+        if (result?.includes("Sorry, I can't assist with that.")) {
+          void vscode.window.showErrorMessage(
+            "Sorry, I can't assist with that."
```
Anemy, line 1731: Should we make this more descriptive? It sounds a bit generic, I'm guessing folks would expect this though if they're exporting something that generates this error. May no change needed.

Anemy (commented):
Looking good! Left a couple small suggestions/questions. I'll give this a run tomorrow

src/participant/prompts/exportToLanguage.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  getUserPrompt({
+    request,
+    includeDriverSyntax,
+  }: ExportToLanguagePromptArgs): Promise<UserPromptResponse> {
+    const prompt = request.prompt;
+    return Promise.resolve({
+      prompt: `${
+        prompt ? `The user provided additional information: "${prompt}"\n` : ''
+      }Include driver syntax: ${includeDriverSyntax}`,
```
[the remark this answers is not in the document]
alenakhineika: You can call the export to langauge from the command palette with an empty playground. I guess in this case we will miss a prompt.

src/telemetry/telemetryService.ts
```diff
@@ -49,18 +49,8 @@ type DocumentEditedTelemetryEventProperties = {
...
-  with_import_statements: boolean;
-  with_builders: boolean;
-  with_driver_syntax: boolean;
-};
-
-type QueryExportedTelemetryEventProperties = {
-  language: string;
-  with_import_statements: boolean;
-  with_builders: boolean;
   with_driver_syntax: boolean;
```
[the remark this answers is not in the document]
alenakhineika: Do you mean `num_stages`? We can export now not only arrays and objects but any code from a playground, so a number of stages value is not relevant anymore. I think I can add a new parameter to track simply the length of the exported string. It will give us a rough understanding of how much code users export.

src/editors/playgroundSelectionCodeActionProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+      this._onDidChangeCodeCodeAction.fire();
+    });
+  }
+
+  readonly onDidChangeCodeLenses: vscode.Event<void> =
+    this._onDidChangeCodeCodeAction.event;
+
+  refresh(): void {
+    this._onDidChangeCodeCodeAction.fire();
+  }
```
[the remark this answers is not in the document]
alenakhineika: Nice catch! Yeah, the `refresh` function is a leftover after the refactoring. Removed it from the class and tests. The `_onDidChangeCodeCodeAction` event tho, we need to keep it for the language server event, when the configuration changes we need to re-render code actions accordingly.

src/telemetry/telemetryService.ts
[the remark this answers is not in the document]
Anemy: I was just thinking the length of the string yup.

src/participant/participant.ts
```diff
@@ -1802,4 +1850,79 @@ Please see our [FAQ](https://www.mongodb.com/docs/generative-ai-faq/) for more i
...
+
+      log.info(`The playground was exported to ${language}`, {
+        code: codeToTranspile,
+        language,
+        includeDriverSyntax: driverSyntax,
+      });
+
+      await vscode.commands.executeCommand(
+        EXTENSION_COMMANDS.SHOW_EXPORT_TO_LANGUAGE_RESULTS,
+        { content, language }
```
Anemy, line 1904: Right now we're getting the full response from the model before we show it. We could stream it to a preview panel. Not something we need to do now, but it could be a nice improvement down the line. Mostly calling it out incase it wasn't something we were looking at before.

src/mdbExtensionController.ts
```diff
@@ -314,6 +314,12 @@ export default class MDBExtensionController implements vscode.Disposable {
         );
       }
     );
+    this.registerParticipantCommand(
+      EXTENSION_COMMANDS.SHOW_EXPORT_TO_LANGUAGE_RESULTS,
+      (data: { content: string | null; language: string }) => {
+        return this._playgroundController.showExportToLanguageResults(data);
```
Anemy, line 320: Showing code from other languages in the playground files makes the syntax lack the correct highlighting. Have we thought about showing these another way? Like a preview pane in the language that is being exported to.

Anemy (approved):
lgtm. Left a few questions/suggestions and user experience suggestions.

src/participant/participant.ts
alenakhineika: I was considering this but matched the behaviour to other cases when we open the results pane with all as one ready results without streaming them.

src/participant/participant.ts
```diff
@@ -1605,82 +1619,61 @@ export default class ParticipantController {
...
+            this.streamChatResponseWithExportToLanguage({
               modelInput,
               token,
             }),
-            new Promise<undefined>((resolve) =>
+            new Promise<null>((resolve) =>
               token.onCancellationRequested(() => {
-                resolve(undefined);
+                void vscode.window.showInformationMessage(
+                  'The export to a playground operation was canceled.'
```
Anemy, line 1645: This adds another click for users, since they are the one who clicked the cancel action I'm thinking we can just silently resolve here. How does removing this sound?

src/editors/playgroundSelectionCodeActionProvider.ts
```diff
@@ -0,0 +1,100 @@
...
+      tooltip: codeActionName,
+    };
+    return codeAction;
+  }
+
+  provideCodeActions(): vscode.CodeAction[] | undefined {
+    const editor = vscode.window.activeTextEditor;
+    const codeActions: vscode.CodeAction[] = [];
+
+    if (!isPlayground(editor?.document.uri) || !getSelectedText()) {
```
Anemy, line 85: Do we not show the export to language actions if the user doesn't have copilot? Not sure on if we're doing that in other actions either.

Anemy (commented):
Looks good, noticed 2 smaller things we probably want to change, and left one question.

src/participant/participant.ts
alenakhineika: We match the playground cancelation case. Should we remove it there as well?

src/editors/exportToLanguageCodeLensProvider.ts
```diff
@@ -2,10 +2,7 @@ import * as vscode from 'vscode';
 
 import EXTENSION_COMMANDS from '../commands';
 import type { ExportToLanguageAddons } from '../types/playgroundType';
-import {
-  ExportToLanguageMode,
-  ExportToLanguages,
-} from '../types/playgroundType';
+import { ExportToLanguages } from '../types/playgroundType';
 
 export default class ExportToLanguageCodeLensProvider
```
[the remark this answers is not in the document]
alenakhineika: Good catch! Will fix this.

src/participant/participant.ts
Anemy: I'd say yes, let's ask Gaurab. I'll create a slack thread. What are your thoughts on it? I'm also thinking back to when we showed notifications for connecting which people found annoying and created prs to turn off lol.

src/editors/playgroundSelectionCodeActionProvider.ts
alenakhineika: Added the copilot is active check 👍

src/mdbExtensionController.ts
Anemy: Ah they did work, I think this was from the previous issue with cancelling.

Anemy (approved):
lgtm! nice

PR #871, opened 2024-11-12
feat(chat): Include history from messages to docs chatbot VSCODE-632

gagik:
Include history from messages to docs chatbot

## Open Questions
Some things may be worth discussing here. cc: @GaurabAryal 

### Limitations

- There's no way to just add previous messages as context (at least not yet) through the app. We can make requests and wait for a full response for each previous message which is quite long.
- So, instead, this implementation includes previous messages by making it all prepended to the prompt with the actual user request in the end.

### Risks

1.  Prompts end up unnecessarily too long and we easily reach the maximum prompt count.
2.  The earlier messages / prompts end up overtaking the question or are drastically different (i.e. user decided to ask something about docs which has nothing to do with what he was doing before and now the docs is referring to all that context and ends up less useful.

### Observations

- I think 1 is definitely a real risk especially if the user has been using the chat for a while and suddenly decides to use documentation.
- In my testing, I don't actually see much of 2 happening even when I am trying to be nefarious; seems like the end of the prompt does take a lot of precedence, so that's good.

### Mitigation Strategies

- We can deal with 1 and 2 by introducing a small cap on how many previous messages we include, limited to as low as 2-3?
- We could also make the prompt being sent more explained to the LLM i.e start with header "For context, previous prompts were: .... \n\n The prompt is: ....` or adding a `------` after the context. So far, I have found lack of clear separation not to be a problem.
- We can also request for the API to add some context-adding endpoint but this will probably involve a lot of waiting

## Types of changes

- [ ] *Backport Needed*


src/participant/participant.ts
```diff
@@ -1446,8 +1448,29 @@ export default class ParticipantController {
...
+
+    log.info(
+      'Sending history to chatbot',
+      history.map((message: vscode.LanguageModelChatMessage) =>
+        getContent(message)
+      )
+    );
+
+    const previousMessages =
+      history.length > 0
```
gagik, line 1464: So... I left it like this as a simple implementation but this needs so more discussion.

### Limitations

- There's no way to just add previous messages as context (at least not yet) through the app. We can make requests and wait for a full response for each previous message which is quite long.
- So, instead, we could include previous messages by making it all part of the same prompt and appending the last part of the request in the end.

### Risks

1.  Prompts end up unnecessarily too long and we easily reach the maximum prompt count.
2.  The earlier messages / prompts end up overtaking the question or are drastically different (i.e. user decided to ask something about docs which has nothing to do with what he was doing before and now the docs is referring to all that context and ends up less useful.

### Observations

- I think 1 is definitely a real risk especially if the user has been using the chat for a while and suddenly decides to use documentation.
- In my testing, I don't actually see much of 2 happening even when I am trying to be nefarious; seems like the end of the prompt does take a lot of precedence, so that's good.

### Mitigation Strategies

- We can deal with 1 and 2 by introducing a small cap on how many previous messages we include, limited to as low as 2-3?
- We could also make the prompt being sent more explained to the LLM i.e start with header "For context, previous prompts were: .... \n\n The prompt is: ....` or adding a `------` after the context. So far, I have found lack of clear separation not to be a problem.
- We can also request for the API to add some context-adding endpoint but this will probably involve a lot of waiting

alenakhineika: FYI the docs team is considering changing the way the endpoint works by not keeping history on the server but instead similarly as the copilot does, requiring to send the whole context with each request. But as you said, this won't, probably, happen soonish, so I like your idea to limit the history messages we include. About 3 sounds reasonable. I am also wondering if we want to include ChatResponseTurn, maybe it would be enough to just pass ChatRequestTurn?

src/participant/prompts/docs.ts
```diff
@@ -0,0 +1,44 @@
+import type * as vscode from 'vscode';
+import type { PromptArgsBase } from './promptBase';
+import { PromptBase } from './promptBase';
+
+export class DocsPrompt extends PromptBase<PromptArgsBase> {
```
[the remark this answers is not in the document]
alenakhineika: Looks like we don't need most of the logic from buildMessages, but only getFilteredHistoryMessages. And we don't need to wrap it in vscode.LanguageModelChatMessage like buildMessages does. Maybe we don't need a dependency on promptBase here at all? Pls point me to it if I am missing something.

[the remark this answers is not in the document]
gagik: Sorry yes, the function I was referring to is `getFilteredHistoryMessages` (edited the comment to make clearer). That's what `docs.getHistoryMessages` (or equivalent) depends on. I was wondering if it'd make sense to take out `getFilteredHistoryMessages` out of `PromptBase` in this case then.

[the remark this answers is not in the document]
alenakhineika: Aaaah yes, that would be fine to make a separate module out of getFilteredHistoryMessages. Just make sure that it logically goes well with modules that depend on it. The idea is to make it more obvious from reading the code which model actually does what and make it easier to reuse the code.

src/participant/participant.ts
alenakhineika: Let's try then with 2 questions & responses 👍

src/participant/prompts/promptBase.ts
```diff
@@ -158,24 +181,116 @@ export abstract class PromptBase<TArgs extends PromptArgsBase> {
...
+          // When the message is the assistant asking for part of a namespace,
+          // we only want to include the question asked, not the user's
+          // database and collection names in the history item.
+          break;
+        }
+      }
+    }
+
+    // eslint-disable-next-line new-cap
+    messages.push(vscode.LanguageModelChatMessage.Assistant(message));
```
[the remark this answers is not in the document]
gagik: Yeah I could try the lookahead, move the push outside and also make the arguments `readonly`

src/participant/prompts/promptHistory.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    // eslint-disable-next-line new-cap
+    return vscode.LanguageModelChatMessage.User(currentTurn.prompt);
+  }
+
+  /** When passing the history to the model we only want contextual messages
+  to be passed. This function parses through the history and returns
+  the messages that are valuable to keep. */
+  static getFilteredHistory({
```
gagik, line 108: this could also just be exported functions; the class just acts as namespacing for the time being so not sure if it makes sense

src/participant/prompts/promptHistory.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    const namespaceIsKnown =
+      databaseName !== undefined && collectionName !== undefined;
+    for (let i = 0; i < history.length; i++) {
+      const currentTurn = history[i];
+
+      let addedMessage: vscode.LanguageModelChatMessage | undefined;
+      if (currentTurn instanceof vscode.ChatRequestTurn) {
+        const previousTurn = i - 1 >= 0 ? history[i - 1] : undefined;
+        const nextTurn = i + 1 < history.length ? history[i + 1] : undefined;
```
alenakhineika, line 133: Just a question, do we use nextTurn to chek for filtered errors?

gagik: could you clarify what you mean by that? we are checking the `nextTurn` here to filter out the requests which have led to errors (specifically just the ones filtered by responsible AI service at the moment)

src/participant/prompts/promptHistory.ts
```diff
@@ -0,0 +1,198 @@
...
+    if (!context) {
+      return [];
+    }
+    const historySinceLastDocs: (
+      | vscode.ChatRequestTurn
+      | vscode.ChatResponseTurn
+    )[] = [];
+
+    /** Limit included messages' history to prevent prompt overflow. */
+    const MAX_DOCS_HISTORY_LENGTH = 4;
```
alenakhineika, line 178: You were thinking about 2-4 docs. Wondering, did 4 work better? Like a bigger content bring more relevant results?

src/participant/prompts/promptHistory.ts
alenakhineika: This is exactly what I was asking for :) Like to understand why we look ahead of the current turn. Wanted to make sure I understood that correctly.

src/participant/prompts/promptHistory.ts
gagik: To be honest, I did not see a big difference, it was more of a question of at what point would it be too much content for the prompt, on average. The only real usefulness I can think of for history with docs is when someone runs i.e. queries and then says something like `/docs what does this do?` i.e. providing context to "this" is useful.

In that sense, 4 seemed fine as it's possible you could do something like `/query find all documents with age > 10` and then run `/query make it also filter for name` and then maybe do something like`/docs tell me more about this` then it _could_ be useful to have context of first message. Of course similarly there might be 3+ 

I think either way, it'd be quite fine.

alenakhineika (approved):

Anemy (approved):

PR #874, opened 2024-11-18
feat(chat): update schema assistant prompt to handle empty and short prompts better VSCODE-648

gagik:
## Description

Sometimes, when using the /query command to build a query, and then at some point using the /schema to get information about the schema of a collection without opening a new chat window, Copilot keeps thinking that a user wants to build a query. This PR tries to fix this by introducing a blank message case handling and modifying the prompt to make it clearer the schema command should always output schemas.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/participant/participant.ts
+++ b/src/participant/participant.ts
@@ -40,11 +40,12 @@ import {
 import { DocsChatbotAIService } from './docsChatbotAIService';
 import type TelemetryService from '../telemetry/telemetryService';
 import formatError from '../utils/formatError';
-import type { ModelInput } from './prompts/promptBase';
+import { getContent, type ModelInput } from './prompts/promptBase';
 import { processStreamWithIdentifiers } from './streamParsing';
 import type { PromptIntent } from './prompts/intent';
 import type { DataService } from 'mongodb-data-service';
 import { ParticipantErrorTypes } from './participantErrorTypes';
+import { PromptHistory } from './prompts/promptHistory';
 
 const log = createLogger('participant');
 
@@ -1415,10 +1416,12 @@ export default class ParticipantController {
     chatId,
     token,
     stream,
+    context,
   }: {
     prompt: string;
     chatId: string;
     token: vscode.CancellationToken;
+    context: vscode.ChatContext;
     stream: vscode.ChatResponseStream;
   }): Promise<{
     responseContent: string;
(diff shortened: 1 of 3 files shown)


src/participant/prompts/schema.ts
```diff
@@ -21,9 +22,10 @@ export class SchemaPrompt extends PromptBase<SchemaPromptArgs> {
     return `You are a senior engineer who describes the schema of documents in a MongoDB database.
 The schema is generated from a sample of documents in the user's collection.
 You must follow these rules.
-Rule 1: Try to be as concise as possible.
-Rule 2: Pay attention to the JSON schema.
-Rule 3: Mention the amount of documents sampled in your response.
+Rule 1: Your answer should always describe the schema of documents in the collection.
```
gagik, line 25: This change seems to work; I tried to trick the Copilot to get confused by previous `/query` queries and it seems to be more consistent even with short prompts i.e. `/schema users`

Open to suggestions if you see some other way.

src/test/suite/participant/participant.test.ts
```diff
@@ -1335,47 +1333,6 @@ suite('Participant Controller Test Suite', function () {
               'Which database would you like to use? Select one by either clicking on an item in the list or typing the name manually in the chat.\n\n'
             );
           });
-
-          test('with history, and a blank prompt, it sets a message so it does not cause model error (VSCODE-626)', async function () {
```
gagik, line 1339: this behavior becomes outdated as it'd always show the special empty message response instead.

src/test/suite/participant/participant.test.ts
```diff
@@ -1300,21 +1313,6 @@ suite('Participant Controller Test Suite', function () {
             });
           });
 
-          test('without a prompt it asks for the database name without pinging ai', async function () {
```
gagik, line 1303: this behavior becomes outdated; unless I am overlooking something, we do not gain too much by asking for the namespace if the user just runs `/schema`. It'd be nicer to show an explainer first to help them discover the syntax.

Anemy: If I remember correctly this was intentional. We want the user to be able to get the schema of a collection by just writing `/schema` and then clicking their namespace.

gagik: Hm I see. I can see how it's convenient but this does also make it hard for a user to actually discover what the `/schema` command expects as an argument (something which we could definitely instead handle by introducing something like `/help` or equivalent for internal extension info or questions). Seems like the only way to discover that you are meant to ask a question with the `/schema` is through reading documentation.

If we do want to support this way, I could remove the empty message response.

gagik:
The new rule actually works pretty great even with empty prompts in my manual testing, so I'll revert the changes for empty prompt and just go ahead with the rule change.

PR #875, opened 2024-11-18
feat(participant): in-app notification to introduce users to copilot extension VSCODE-633

alenakhineika:
## Description

Show an information toast to notify users to check out the MongoDB Copilot participant. Show the notification regardless of whether users have copilot installed or not.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


Anemy (approved):
Changes look good, I am a bit worried on the spam potential though, left a comment there.

src/mdbExtensionController.ts
```diff
@@ -72,6 +72,7 @@ export default class MDBExtensionController implements vscode.Disposable {
   _editDocumentCodeLensProvider: EditDocumentCodeLensProvider;
   _exportToLanguageCodeLensProvider: ExportToLanguageCodeLensProvider;
   _participantController: ParticipantController;
+  _surveyShown = false;
```
Anemy, line 75: nit: Could we rename this to `_startupNotificationShown`, and then update the logic to be that both of the notification methods check for this, not just the copilot one? That way it's not bound to where we call it in the code. Things may shift around and we might add more notifications in the future so it'll make it a bit more resilient.

Anemy (approved):

gagik (approved):

PR #881, opened 2024-11-22
feat(playground): add Generate Query with Copilot code lens in playgrounds VSCODE-650

gagik:
Part of [VSCODE-650](https://jira.mongodb.org/browse/VSCODE-650).

This does not include the new icons; this is included in https://github.com/mongodb-js/vscode/pull/885.


src/telemetry/telemetryService.ts
```diff
@@ -130,6 +130,10 @@ export type CopilotIntroductionProperties = {
   is_copilot_active: boolean;
 };
 
+export type ParticipantOpenedFromInputBoxProperties = {
+  source?: 'query with copilot codelens';
```
gagik, line 134: We have the `DOCUMENT_SOURCE` enum which maybe could be relevant here but this seems quite different?

src/editors/queryWithCopilotCodeLensProvider.ts
```diff
@@ -0,0 +1,46 @@
...
+    // but not whether it has an active subscription.
+    const hasCopilotChatActive =
+      vscode.extensions.getExtension(COPILOT_EXTENSION_ID)?.isActive === true;
+
+    if (!hasCopilotChatActive) {
+      return [];
+    }
+
+    const options: SendMessageToParticipantFromInputOptions = {
+      prompt: 'Describe the query you would like to generate.',
```
alenakhineika, line 30:
```suggestion
      prompt: 'Describe the query you would like to generate',
```

Let's maybe remove `.` so it looks less weird with the line that vscode adds automatically.

Screenshot 2024-11-27 at 16 24 53

src/participant/participant.ts
```diff
@@ -130,9 +134,51 @@ export default class ParticipantController {
    * in the chat. To work around this, we can write a message as the user, which will
    * trigger the chat handler and give us access to the model.
    */
-  writeChatMessageAsUser(message: string): Thenable<unknown> {
+  async sendMessageToParticipant(
```
alenakhineika, line 137: Could you maybe also to refactor this part to use the sendMessageToParticipant helper?

https://github.com/mongodb-js/vscode/blob/main/src/mdbExtensionController.ts#L952-L959

src/participant/participantTypes.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export type SendMessageToParticipantOptions = {
+  message: string;
+  isNewChat?: boolean;
+  isPartialQuery?: boolean;
+};
+
+export type SendMessageToParticipantFromInputOptions = {
+  messagePrefix?: string;
+  source?: 'query with copilot codelens';
```
alenakhineika, line 11:
```suggestion
  source?: string;
```
Can it be maybe more generic and just a string?

src/telemetry/telemetryService.ts
alenakhineika: Do we use `source` only for telemetry or how else is this used?

src/participant/participantTypes.ts
gagik: I agree that this should be flexible but I was thinking of this as documentation of the sources we have to prevent inconsistencies, like an enum, so in the future if we have more we can add more. I could make a literal enum similar to `DocumentSource`

src/telemetry/telemetryService.ts
gagik: Yep, just telemetry. Specifically would be useful for Gaurab to track how much query with copilot is used. I thought it'd be better to do it like this instead of introducing a new event just for query with copilot

src/participant/participantTypes.ts
alenakhineika: Yeah, let's make it maybe a part of DocumentSource then.

alenakhineika (approved):

PR #882, opened 2024-11-22
fix: show export to language icon only when copilot is active VSCODE-655

alenakhineika:
## Description

Show export to language icon only when the copilot is active.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/mdbExtensionController.ts
+++ b/src/mdbExtensionController.ts
@@ -168,6 +168,21 @@ export default class MDBExtensionController implements vscode.Disposable {
     this.showOverviewPageIfRecentlyInstalled();
     void this.showSurveyForEstablishedUsers();
     void this.showCopilotIntroductionForEstablishedUsers();
+
+    const copilot = vscode.extensions.getExtension('GitHub.copilot');
+    void vscode.commands.executeCommand(
+      'setContext',
+      'mdb.isCopilotActive',
+      copilot?.isActive
+    );
+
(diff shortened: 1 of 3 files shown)


src/mdbExtensionController.ts
```diff
@@ -168,6 +168,21 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+      'setContext',
+      'mdb.isCopilotActive',
+      copilot?.isActive
+    );
+
+    vscode.extensions.onDidChange(() => {
+      void vscode.commands.executeCommand(
+        'setContext',
+        'mdb.isCopilotActive',
+        copilot?.isActive
```
gagik, line 183: Could it be that this value doesn't actually update in this case? Would it not use old set variable and we might need to freshly run `vscode.extensions.getExtension('GitHub.copilot')` here ?

alenakhineika: I have manually tested and this updated the value and shows/hides the icon accordingly.

alenakhineika: Played with this more and removed the `vscode.extensions.onDidChange` listener. See for more details the vscode ticket: https://github.com/microsoft/vscode/issues/234426

gagik (approved):

PR #885, opened 2024-11-25
feat(playground): add leaf connection icons inside playground VSCODE-650

gagik:
This adds a script to generate an icon font from the SVGs we have and use them inside VSCode, specifically the playground connection codelens.


scripts/generate-icon-font.ts
```diff
@@ -0,0 +1,63 @@
...
+/** Icons to include in the generated icon font */
+const INCLUDED_ICONS = ['connection-active', 'connection-inactive'];
+
+/**
+ * Generates an icon font from the included icons and outputs package.json
+ * configuration field for those icons as specified in
+ * https://code.visualstudio.com/api/references/icons-in-labels
+ */
+async function main(): Promise<void> {
+  const font = await webfont({
```
gagik, line 14: I think it's worth it to have a external dev dependency / script for this. We can potentially utilize Vscode's support for icons fonts for i.e. easily having different color versions of the same icons and, as in this case, easily including icons in different places.
The link from the comment:
https://code.visualstudio.com/api/references/icons-in-labels

package.json
```diff
@@ -1154,6 +1155,22 @@
           "description": "Specify a shell command that is run to start the browser for authenticating with the OIDC identity provider for the server connection. Leave this empty for default browser."
         }
       }
+    },
+    "icons": {
+      "connection-active": {
```
Anemy, line 1160: Should we prefix this with a MongoDB identifier? That way if there's ever a `connection-active` added to the long icon list https://code.visualstudio.com/api/references/icons-in-labels#icon-listing or by another extension we don't conflict.
Maybe `mongodb-connection-active`?

gagik: agreed, it'd actually help with the JSON schema for this as VSCode always expects at least one `-`. Could `mdb-` be alright too to keep the strings with icons short?

Anemy: `mdb-` sgtm!

alenakhineika:
Maybe the style of icons would look more aligned if the `ai` icon also didn't have a color like the `connected` icon?

Screenshot 2024-11-27 at 14 26 46

gagik:
> Maybe the style of icons would look more aligned if the `ai` icon also didn't have a color like the `connected` icon?
> Screenshot 2024-11-27 at 14 26 46

We discussed that with @GaurabAryal and figured color could be cool in terms of encouraging people to click on them for this specific case.

alenakhineika (approved):

Anemy (approved):

PR #887, opened 2024-11-26
feat(copilot): optimize namespace for export to playground VSCODE-654

alenakhineika:
## Description

- Update prompt with the `use()` example to generate a playground with a proper namespace usage
- Rebrand to MongoDB Copilot Extension instead of participant in the context menu

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


gagik (approved):

src/participant/prompts/exportToPlayground.ts
```diff
@@ -4,6 +4,30 @@ export class ExportToPlaygroundPrompt extends PromptBase<PromptArgsBase> {
   protected getAssistantPrompt(): string {
     return `You are a MongoDB expert.
 Your task is to convert user's code written in any programming language to the MongoDB mongosh shell script.
+If the user's code contains a database and collection name, preserve them in the transpiled code,
+otherwise use '<YOUR_DATABASE_NAME>' and 'YOUR_COLLECTION_NAME' placeholders.
```
nirinchev, line 8: nit: not sure if the model actually consolidates those when generating the response, but the formatting here is inconsistent - we should choose whether we want to include `<>` or not and use a consistent formatting for both db and col.

alenakhineika: Good catch! It should definitely be aligned. I will open a tiny pr.

PR #890, opened 2024-11-28
feat(tree-explorer): add buttons to ask Copilot and create playgrounds from tree view VSCODE-651

gagik:
Includes everything in [VSCODE-651](https://jira.mongodb.org/browse/VSCODE-651) other than telemetry as I ended up doing some minor refactoring in it for in https://github.com/mongodb-js/vscode/pull/891


src/participant/participant.ts
```diff
@@ -120,9 +123,44 @@ export default class ParticipantController {
...
+      void vscode.commands.executeCommand(
+        'setContext',
+        'mdb.isCopilotActive',
+        copilot?.isActive === true
+      );
+      if (copilot !== undefined && !copilot.isActive) {
+        // TODO: This is a workaround for https://github.com/microsoft/vscode/issues/234426
+        // The onDidChange event only gets fired if extensions get added or removed but
+        // not when they are activated; so retry after a delay and check again.
+        setTimeout(() => {
```
gagik, line 148: I could not think of a better way to handle this... we could get away with just a check in other places as it'd usually happen later than Copilot activation but because this happens at extension activation point, we don't have much guarantee over the order extensions get activated

src/participant/participant.ts
```diff
@@ -182,6 +226,34 @@ export default class ParticipantController {
     });
   }
 
+  async askCopilotFromTreeItem(
+    treeItem: DatabaseTreeItem | CollectionTreeItem
+  ): Promise<unknown> {
```
paula-stacho, line 231: why not `Promise<void>`?

src/templates/playgroundCreateCollectionTemplate.ts
```diff
@@ -2,14 +2,11 @@ const template = `/* global use, db */
 // MongoDB Playground
 // Use Ctrl+Space inside a snippet or a string literal to trigger completions.
 
-const database = 'NEW_DATABASE_NAME';
-const collection = 'NEW_COLLECTION_NAME';
-
 // Create a new database.
-use(database);
+use('CURRENT_DATABASE');
```
alenakhineika, line 6: Did Gaurab request it?

src/editors/playgroundController.ts
```diff
@@ -316,13 +317,38 @@ export default class PlaygroundController {
...
+    if (treeItem instanceof DatabaseTreeItem) {
+      content = playgroundInsertDocumentTemplate.replace(
+        'CURRENT_DATABASE',
+        treeItem.databaseName
+      );
+      this._telemetryService.trackPlaygroundCreated('fromDatabaseTreeItem');
+    } else if (treeItem instanceof CollectionTreeItem) {
+      content = playgroundInsertDocumentTemplate
+        .replace('CURRENT_DATABASE', treeItem.databaseName)
+        .replace('CURRENT_COLLECTION', treeItem.collectionName);
```
alenakhineika, line 333: Looks like in the `src/templates/playgroundCreateCollectionTemplate.ts` we keep using the `NEW_COLLECTION_NAME` name, not `CURRENT_COLLECTION`.

src/templates/playgroundCreateCollectionTemplate.ts
gagik: > Additionally, we should also add an icon for Playgrounds here for so users can directly jump to Playgrounds from the connection tree. If users are navigating at the DB level, we can pre-populate playgrounds with `use(<DB_Name>);` and if they are clicking from the collection level, we can additionally include a simple db.<collection_name>.findOne().

This was the request from Guarab; I thought I'd just reuse this template but just noticed the `findOne()` detail so will likely create a new one. Other than that, just thought it seemed nice to make the template shorter by moving the variables in but I could just leave it as is.

src/participant/participant.ts
alenakhineika: We already do it during the extension activation: https://github.com/mongodb-js/vscode/blob/main/src/mdbExtensionController.ts#L180-L185. Do we need to do it twice?

src/templates/playgroundCreateCollectionTemplate.ts
alenakhineika: Setting names as variables first was intentional to draw users' attention that those are placeholders and need to be changes according to their namespase.

src/participant/participant.ts
gagik: On my Mac at least, it doesn't work when we set it on activation; at this point `copilot?.isActive` is `false` for me so the buttons don't show up 🫨

alenakhineika: Interesting, i have tried your branch without `subscribeToCopilotStatus` and it shows me icons in the tree, so the extension activation catches `copilot?.isActive` as `true`. Maybe you can try to move `setTimeout` to the extension activation method, so we retry there?

src/editors/playgroundController.ts
addaleax, line 333: Keep in mind that both database and collection names can contain special characters, such as newlines and quotes 🙃

gagik: Ah good point (and wow a database name can contain a newline...), though I guess we haven't been mindful about this generally. @alenakhineika thoughts on whether we'd want to handle this now or maybe have some ticket to keep track of special cases for corner case database/collection names?

addaleax: I'd handle it, if at all possible. We can use `JSON.stringify()` here, since these are just strings, that takes care of escaping as necessary.

src/participant/participant.ts
gagik: yeah that did work for me but only with the set timeout. My guess is that for me Copilot always gets activated _after_ the MongoDB extension, maybe I just have more extensions than your setup or something else like that? 

It is always false otherwise at the moment of MongoDB extension activation.

alenakhineika: Let's place then this workaround with timeout inside the extension `activate` method https://github.com/mongodb-js/vscode/blob/main/src/mdbExtensionController.ts#L164 until we hear something from https://github.com/microsoft/vscode/issues/234426.

src/templates/templateHelpers.ts
```diff
@@ -0,0 +1,10 @@
+/** Wraps a template function and escapes given string arguments. */
+export function createTemplate<T extends (...args: string[]) => string>(
```
gagik, line 2: While we're going for string escaping, I thought it might be nice to switch to something more type-safe and use functions + string templates, would help keep track of variables used better. 

I could convert other templates at a later time also, but this could be a nice starting point, unless there's benefits to replacing that I might be overseeing.

In case we end up with content that needs to not be escaped somehow, we could expand this function to support more complex cases but for now I just wanted to create a wrapper.

src/mdbExtensionController.ts
```diff
@@ -330,6 +344,13 @@ export default class MDBExtensionController implements vscode.Disposable {
         return true;
       }
     );
+    this.registerParticipantCommand(
+      EXTENSION_COMMANDS.ASK_COPILOT_FROM_TREE_ITEM,
+      async (treeItem: DatabaseTreeItem) => {
```
nirinchev, line 349: Should `treeItem` here be `DatabaseTreeItem | CollectionTreeItem`?

src/mdbExtensionController.ts
```diff
@@ -742,6 +763,11 @@ export default class MDBExtensionController implements vscode.Disposable {
       EXTENSION_COMMANDS.MDB_CREATE_PLAYGROUND_FROM_TREE_VIEW,
       () => this._playgroundController.createPlayground()
     );
+    this.registerCommand(
+      EXTENSION_COMMANDS.MDB_CREATE_PLAYGROUND_FROM_CONNECTION_TREE_VIEW,
+      (treeItem: DatabaseTreeItem | CollectionTreeItem) =>
+        this._playgroundController.createPlaygroundFromTreeView(treeItem)
```
nirinchev, line 769: [super nit] The naming here is a bit inconsistent/confusing. On line 348, we're asking "from tree item", whereas here it's "from connection tree view" - why do we have a connection here and why is it "view" and not "item"? Then on line 769, we're dropping "connection", but keep calling it "view", even though the argument is "item".

src/participant/participant.ts
```diff
@@ -83,6 +84,7 @@ export default class ParticipantController {
   _docsChatbotAIService: DocsChatbotAIService;
   _telemetryService: TelemetryService;
   _playgroundResultProvider: PlaygroundResultProvider;
+  _copilotStatusSubscription?: vscode.Disposable;
```
nirinchev, line 87: Is this ever assigned to something? I can only see it being disposed, but never assigned.

src/participant/participant.ts
```diff
@@ -182,6 +195,33 @@ export default class ParticipantController {
...
+        isNewChat: true,
+      });
+    } else {
+      throw new Error('Unsupported tree item type');
+    }
+
+    await this.sendMessageToParticipant({
+      message: '',
+      isPartialQuery: true,
+    });
```
nirinchev, line 222: What's the purpose of this? As far as I can tell, once we send the first message, it keeps the `@MongodB` scope.

src/templates/templateHelpers.ts
nirinchev, line 2: Not sure I understand the benefit/need for this to be a generic function.

nirinchev: Would it make sense to  use handlebars or a similar templating engine rather than roll something ourselves?

gagik: The purpose is to preserve type safety / IDE suggestions so if you do something like `const template = createTemplate((one, two) => ....)`
the IDE would enforce `one` and `two` as the string arguments for `template`.

src/participant/participant.ts
gagik: ah this was left over from a previous approach to the workaround that was here; needs to be removed.

src/templates/templateHelpers.ts
gagik: that might be nice though not sure how complex our templates can get; in any case I'd maybe personally prefer an engine written with TypeScript in mind like [eta](https://www.npmjs.com/package/eta)

With stuff like replacing characters we'd probably end up handling it ourselves or use a plugin with these template engines. Honestly to me having a function build a string seems like a nicer experience mainly because of arguments being exposed in the IDE though perhaps a template engine would be more scalable for the future.

cc: @alenakhineika

nirinchev: Ah, I see, makes sense.

gagik: With Gaurab we have discussed a revamp of how we do templates with things like adding shortcut information to them, it might be a better moment to discuss introducing i.e. a templating engine, I created [VSCODE-661](https://jira.mongodb.org/browse/VSCODE-661) to keep track of that.

I could go ahead with this setup for now just to have it in for release and we can change it to whatever we decide more globally later.

src/participant/participant.ts
```diff
@@ -120,6 +121,7 @@ export default class ParticipantController {
       participantId: this._participant?.id,
     });
     this._participant.onDidReceiveFeedback(this.handleUserFeedback.bind(this));
+
```
alenakhineika, line 124:
```suggestion
```

alenakhineika (approved):

PR #891, opened 2024-11-29
feat(tree-explorer): add treeview telemetry and refactor to use a more standardized format VSCODE-651

gagik:
Built on top of https://github.com/mongodb-js/vscode/pull/890

Makes some minor quality of life improvements to our current telemetry setup, especially for `sendMessageToParticipant`.

1. Introduce a `telemetry` field in `sendMessageToParticipant` (ideally a structure we could use in other places in the future?), which is meant to be a scalable structure for us to pass different arguments to functions we want to track without messing with our objects and too much manual work.
2. Make `command` field in various telemetry functions stricter; introduce and cleanup some types to better fit this. Right now we are not keeping a good track of what we call a `command` when sending telemetry, so this is hoping to document this better with types.
3. Remove `Copilot` prefix in places where it may be redundant or not match its property type. 

Let me know what you think; this was mainly me a result of me trying to understand & document our current setup better, would be happy to adjust accordingly 🙂


src/participant/participant.ts
```diff
@@ -188,10 +191,28 @@ export default class ParticipantController {
         'workbench.action.chat.clearHistory'
       );
     }
+    const commandPrefix = command ? `/${command} ` : '';
+    const query = `@MongoDB ${commandPrefix}${message}`;
+
+    if (telemetry) {
```
gagik, line 197: This is an example of it might be nice to structure tracking in other functions as some hot-pluggable thing.

src/telemetry/telemetryService.ts
```diff
@@ -477,14 +517,14 @@ export default class TelemetryService {
       command,
       error_code: errorCode,
       error_name: errorName,
-    });
+    } as ParticipantResponseFailedProperties);
```
addaleax, line 520: [non-blocking]

```suggestion
    } satisfies ParticipantResponseFailedProperties);
```

I think we can do this now, our TS version and tooling seems to be fine with it (and it's generally nicer since it avoids the behavior of `as` that allows type coercions even in case of mismatches)

src/test/suite/participant/participant.test.ts
```diff
@@ -1838,6 +1839,10 @@ Schema:
             {
               message: `I want to ask questions about the \`${mockDatabaseItem.databaseName}\` database.`,
               isNewChat: true,
+              telemetry: {
+                source: DocumentSource.DOCUMENT_SOURCE_TREEVIEW,
+                sourceDetails: 'copilot button on database tree item',
```
alenakhineika, line 1844: Can it be a case with one source but different sourceDetails? Maybe we could pass only one arg something like telemetrySource and then match them somewhere in one place to sourceDetails message?

gagik: I was thinking of having `telemetry` be an object in case in other cases (i.e. in other functions) we'd like to include more properties to include in tracking. Then we could have somewhat of a consistent structure when we pass telemetry information to functions / commands we use.

Could you clarify what you mean by matching in one place to sourceDetails? Not sure I follow

alenakhineika: What is the telemetry event that is sent in this case?

gagik:
```
  PARTICIPANT_INPUT_BOX_SUBMITTED = 'Participant Inbox Box Submitted'
```

alenakhineika: Yeah, so my question is, why do we need textual `sourceDetails`, when we already have `source`? Like based on `source=DocumentSource.DOCUMENT_SOURCE_TREEVIEW` we know this was triggered from the tree view. And based on 'Participant Inbox Box Submitted', we know this is from copilot. Do we really need to send `copilot button on database tree item`?

gagik: that's fair. I was just thinking whether we'd care about whether users care from a collection item or a database item instead of tree view in general. `DocumentSource` so far has been kind of high-level so far i.e. 
```ts
export enum DocumentSource {
  DOCUMENT_SOURCE_TREEVIEW = 'treeview',
  DOCUMENT_SOURCE_PLAYGROUND = 'playground',
  DOCUMENT_SOURCE_COLLECTIONVIEW = 'collectionview',
  DOCUMENT_SOURCE_CODELENS = 'codelens',
}
```
I could either just not care about details or add new categories if we do care about tracking which type of it this is.

gagik: applied this; maybe still `source_details` or `source_info`? `source_type` seems too similar to `source` and its hard to differentiate

alenakhineika: Maybe then something very explicit like `source_tree_item`? `source_details` and `source_info` make me expect some kind of object or a long description.

alenakhineika: But it can be no a tree next time... Ok I agree on as it is now with source_details :)

PR #893, opened 2024-12-02
chore: add automatic font generation and fix VSCode settings

gagik:
This just improves the icon generator script by making it automatically modify the `package.json` with the new icon configuration.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,5 @@
 // Place your settings in this file to overwrite default and user settings.
 {
-  "editor.formatOnSave": false,
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": "explicit"
   },
--- a/scripts/generate-icon-font.ts
+++ b/scripts/generate-icon-font.ts
@@ -1,6 +1,7 @@
 import webfont from 'webfont';
 import fs from 'fs/promises';
 import { GlyphData } from 'webfont/dist/src/types';
+import prettier from 'prettier';
 
 /** Icons to include in the generated icon font */
 const INCLUDED_ICONS = ['connection-active', 'connection-inactive'];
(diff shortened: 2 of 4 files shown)


.vscode/settings.json
```diff
@@ -1,6 +1,5 @@
 // Place your settings in this file to overwrite default and user settings.
 {
-  "editor.formatOnSave": false,
```
gagik, line 3: This is unrelated. I am not sure I see value in this. format on save override that is stated below never works for me so I always end up removing it while working on the project and it all seems fine.

scripts/generate-icon-font.ts
```diff
@@ -40,14 +41,19 @@ async function main(): Promise<void> {
...
+  );
+
+  currentConfiguration.contributes.icons = iconsConfig;
+
+  const prettierConfig = await prettier.resolveConfig('./.prettierrc.json');
+  await fs.writeFile(
+    './package.json',
+    prettier.format(JSON.stringify(currentConfiguration), {
+      ...prettierConfig,
+      parser: 'json-stringify',
```
alenakhineika, line 55: Which exactly rules are applied here? I am wondering what is there on top of `JSON.stringify(..., null, 2)`.

gagik: I think this does end up being equivalent to that, it's just tied to [our prettier config](https://github.com/mongodb-js/devtools-shared/blob/d9db8b5b49c745d29d17097f439be14d565092c5/configs/prettier-config-devtools/index.json) (and we have nearly nothing there so... this should be default prettier) so it just stays consistent with any style rules for our `prettier` formatter. This way if someone autoformats the file using VSCode there won't be inconsistencies.

nirinchev:
A bit late to the party, but I'm not a huge fan of including prettier into the font generation script. Instead, my suggestion would be to invoke it in the package.json script like:

```json
  "generate-icon-font": "ts-node ./scripts/generate-icon-font.ts && prettier --write package.json",
  // or
  "generate-icon-font": "ts-node ./scripts/generate-icon-font.ts && reformat",
```

That way we keep the script less tied to the tooling we're using and any changes we decide to make to the way we format code would be confined to package.json and not some random scripts all over the repo.

It's not a blocker and not something I necessarily feel we should fix now that the PR is merged, but might be something to consider.

PR #894, opened 2024-12-03
feat(participant): filter message history when it goes over maxInputTokens VSCODE-653

gagik:
Drops messages in history if their total would lead to too many tokens being used.


src/test/suite/participant/participant.test.ts
```diff
@@ -849,13 +854,27 @@ suite('Participant Controller Test Suite', function () {
                     ],
                   },
                 },
-              ]);
+              ];
+
+              // This is to offset the previous countTokens calls
+              // buildMessages gets called twice for namespace so it is adjusted accordingly
+              // (1 for request prompt and 1 for assistant prompt calculation)
```
gagik, line 861: Not a fan of this offset but `withArgs` gets complicated with the weird VSCode message wrapper that ends up being applied to this point. I tried to use `sinon.match` but seemed to take a lot of time to figure out a way to do this better.

alenakhineika: Maybe you can list those 4 calls more explicitly in the comment, like:
```
1.
2.
3.
4.
```
to clearly see what those are?

alenakhineika:
Have you tried to run accuracy tests to ensure we haven't broken the experience while manipulating the history?

gagik:
> Have you tried to run accuracy tests to ensure we haven't broken the experience while manipulating the history?

Just ran them, results are consistent with what I have had before so seems alright 🤞

alenakhineika (approved):

PR #898, opened 2024-12-05
fix(connect): can not select files on the connection form VSCODE-658

alenakhineika:
## Description

Electron 32.0 removed support for the `path` property of the Web File object in favor of the `webUtils.getPathForFile` method (electron [blog](https://github.com/electron/electron/blob/83d704009687956fb4b69cb13ab03664d7950118/docs/breaking-changes.md%23removed-filepath)). We can not import `dialog` and `webUtils` from 'electron' in the sandboxed webview https://github.com/microsoft/vscode/issues/235284. To work around this, we use a custom dialog provider that uses webview APIs to send a message to the extension process to open the electron file dialog and listen for the response to get the file path and send them to the electron file input backend.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/views/webview-app/extension-app-message-constants.ts
```diff
@@ -123,10 +137,12 @@ export type MESSAGE_FROM_WEBVIEW_TO_EXTENSION =
   | OpenConnectionStringInputMessage
   | OpenTrustedLinkMessage
   | RenameConnectionMessage
-  | EditAndConnectConnection;
+  | EditAndConnectConnection
```
gagik, line 140: nit: maybe `EditConnectionAndConnect`?

src/views/webview-app/extension-app-message-constants.ts
```diff
@@ -123,10 +137,12 @@ export type MESSAGE_FROM_WEBVIEW_TO_EXTENSION =
...
+  | OpenFileChooser;
 
 export type MESSAGE_FROM_EXTENSION_TO_WEBVIEW =
```
gagik, line 143: I know this is from before but I'm curious, is there a reason this type is uppercase snake case?

gagik (approved):

src/views/webview-app/extension-app-message-constants.ts
alenakhineika: Yeah, this is from before, but I agree I could have a better name :) I will update!

src/views/webview-app/extension-app-message-constants.ts
alenakhineika: A good point as well :) i don't know the reasoning behind it, but also think it could be lowercase. I will update this!

src/views/webviewController.ts
```diff
@@ -180,6 +180,14 @@ export default class WebviewController {
         });
         this._telemetryService.track(TelemetryEventTypes.CONNECTION_EDITED);
         return;
+      case MESSAGE_TYPES.OPEN_FILE_CHOOSER:
+        const files = await vscode.window.showOpenDialog();
```
addaleax, line 184: We should probably communicate back any errors thrown by this as well

src/views/webview-app/overview-page.tsx
```diff
@@ -55,30 +61,83 @@ const OverviewPage: React.FC = () => {
...
+  // Electron 32.0 removed support for the `path` property of the Web File object in favor of the webUtils.getPathForFile method.
+  // https://github.com/electron/electron/blob/83d704009687956fb4b69cb13ab03664d7950118/docs/breaking-changes.md%23removed-filepath
+  // We can not import `dialog` and `webUtils` from 'electron' in the sandboxed webview.
+  // To work around this, we use a custom dialog provider that uses webview APIs
+  // to send a message to the extension process to open the electron file dialog
+  // and listen for the response to get the file path and send them to the electron file input backend.
+  const dialogProvider = {
+    getCurrentWindow(): void {},
+    dialog: {
+      showSaveDialog(): Promise<{ canceled: boolean; filePath?: string }> {
```
addaleax, line 90: Wouldn't we want to propagate options passed to this function? Wouldn't we want to propagate the dialog type (save vs open)?

src/views/webview-app/overview-page.tsx
```diff
@@ -55,30 +61,83 @@ const OverviewPage: React.FC = () => {
...
+            return { canceled: false, filePath: file?.path };
+          }
+        );
+      },
+      showOpenDialog(): Promise<{ canceled: boolean; filePaths: string[] }> {
+        return handleOpenFileChooserResult<Uri[] | undefined>().then(
+          (files?: Uri[]) => {
+            return {
+              canceled: false,
+              filePaths: files ? files?.map((file) => file.path) : [],
```
addaleax, line 102: Shouldn't this be `.fsPath` instead of `.path`?

addaleax (changes requested):

src/views/webview-app/overview-page.tsx
alenakhineika: Looks like we don't have this value in the returned Uri object:
```
open files----------------------
extensionHostProcess.js:162
[
  Ko {
    scheme: 'file',
    authority: '',
    path: '/Users/alena.khineika/ca.pem',
    query: '',
    fragment: '',
    _formatted: null,
    _fsPath: null
  }
]
```

src/views/webview-app/overview-page.tsx
```diff
@@ -55,30 +64,86 @@ const OverviewPage: React.FC = () => {
...
+    });
+  }
+
+  // Electron 32.0 removed support for the `path` property of the Web File object in favor of the webUtils.getPathForFile method.
+  // https://github.com/electron/electron/blob/83d704009687956fb4b69cb13ab03664d7950118/docs/breaking-changes.md%23removed-filepath
+  // We can not import `dialog` and `webUtils` from 'electron' in the sandboxed webview.
+  // To work around this, we use a custom dialog provider that uses webview APIs
+  // to send a message to the extension process to open the electron file dialog
+  // and listen for the response to get the file path and send them to the electron file input backend.
+  const dialogProvider = {
```
nirinchev, line 92:
```suggestion
  const dialogProvider: ElectronShowFileDialogProvider<void> = {
```

src/views/webview-app/overview-page.tsx
```diff
@@ -55,30 +64,86 @@ const OverviewPage: React.FC = () => {
     resetGlobalCSS();
   }, []);
 
+  function handleOpenFileChooserResult<T>(
+    options: FileChooserOptions
+  ): Promise<T> {
+    const requestId = handleOpenFileChooser(options);
+    return new Promise((resolve) => {
+      const messageHandler = (event) => {
+        const message: MessageFromExtensionToWebview = event.data;
```
nirinchev, line 73:
```suggestion
      const messageHandler = (event: MessageEvent<MessageFromExtensionToWebview>): void => {
        const message = event.data;
```

src/views/webview-app/vscode-api.ts
```diff
@@ -29,6 +30,17 @@ export const sendConnectToExtension = (
   });
 };
 
+export const sendOpenFileChooserToExtension = (
+  fileChooserOptions: FileChooserOptions,
+  requestId: string
+) => {
```
nirinchev, line 36:
```suggestion
): void => {
```

src/views/webviewController.ts
```diff
@@ -105,6 +106,54 @@ export default class WebviewController {
...
+  }: {
+    panel: vscode.WebviewPanel;
+    fileChooserOptions: FileChooserOptions;
+    requestId: string;
+  }): Promise<void> => {
+    let files;
+
+    try {
+      files = await vscode.window[
+        fileChooserOptions.mode === 'open' ? 'showOpenDialog' : 'showSaveDialog'
```
nirinchev, line 122: Since `mode` is an enum, these kinds of if-statements are very brittle as they assume that it will only ever have two values. Instead, it's better to use a switch, even if it means the code will be a bit more verbose.

src/views/webviewController.ts
```diff
@@ -105,6 +106,54 @@ export default class WebviewController {
...
+    void panel.webview.postMessage({
+      command: MESSAGE_TYPES.OPEN_FILE_CHOOSER_RESULT,
+      fileChooserResult: {
+        canceled: false,
+        ...(fileChooserOptions.mode === 'open'
+          ? { filePaths: files ? files?.map((file) => file.path) : [] }
+          : { filePath: files?.path }),
+      },
+      requestId,
+    });
```
nirinchev, line 154: This is getting executed, even if an exception occurs - is this what we want?

src/views/webviewController.ts
```diff
@@ -105,6 +106,54 @@ export default class WebviewController {
...
+    } catch (error) {
+      void vscode.window.showErrorMessage(
+        `Unable to open file chooser dialog: ${error}`
+      );
+    }
+
+    void panel.webview.postMessage({
+      command: MESSAGE_TYPES.OPEN_FILE_CHOOSER_RESULT,
+      fileChooserResult: {
+        canceled: false,
```
nirinchev, line 148: We never send canceled - is this correct? Should we not report when a user cancels the operation?

src/views/webview-app/use-connection-form.ts
```diff
@@ -188,6 +200,11 @@ export default function useConnectionForm() {
         type: 'close-connection-form',
       });
     },
+    handleOpenFileChooser: (options: FileChooserOptions) => {
```
nirinchev, line 203:
```suggestion
    handleOpenFileChooser: (options: FileChooserOptions): string => {
```

src/views/webviewController.ts
alenakhineika: Yes, we send the requestId back so we can remove the event listener and files will be undefined what is fine.

src/views/webviewController.ts
alenakhineika: The vscode electron dialog does not return a canceled state. We keep `canceled` here to satisfy the `createElectronFileInputBackend` API. We could assume that if returned files are undefined then the user has hanseled file selection, but from UX perspective it will be the same behaviour, since we won't show any toast if a user manually clicked the cancel button.

src/views/webview-app/overview-page.tsx
addaleax: @alenakhineika Yeah no, it's a getter on the class, that's why it doesn't show up in `util.inspect()`. You'll want to check that this actually works for Windows, and for files with characters that need URL escaping (e.g. spaces) in their paths.

alenakhineika: This output is actually printed with `util.inspect()`, and when I tried to replace it, the value was not present.

addaleax: The value is present if I print `util.inspect((files as vscode.Uri[]).map(file => file.fsPath))`, as expected (and as documented in the VSCode API)

alenakhineika: Yeah, i have no idea what i was printing last time 🙈 tried again and you are right, the value is there.

PR #899, opened 2024-12-06
feat(participant): remove in-app notification for participant VSCODE-662

gagik:
We decided to remove the notification for Copilot extension so we don't spam existing users.

This essentially reverts changes in https://github.com/mongodb-js/vscode/pull/875


src/telemetry/telemetryService.ts
```diff
@@ -517,6 +517,7 @@ export default class TelemetryService {
       command,
       error_code: errorCode,
       error_name: errorName,
+      error_message: message,
```
alenakhineika, line 520: We can not track the message itself, because it can contain user's sensitive data.

gagik: oh sorry this was meant to go in the https://github.com/mongodb-js/vscode/pull/900 PR also... but glad that was caught, had no idea! I'm not too sure what we could to do to improve error tracking beyond this

alenakhineika: Yeah, this is the cause of the problem in the first place, that we can not track a message, and why Kevin was asking about maybe different specialized tools we could utilize. I'm also currently not sure what we could do to improve error tracking.

gagik: then I could revert and leave it as is for now; hopefully we'll at least see the reduction in the unknown error we see with exportToPlayground

alenakhineika: So let's exclude `error_message` for now and think about ways to improve error tracking later in a separate pr.

gagik: Yep, removed it here

src/telemetry/telemetryService.ts
```diff
@@ -527,14 +526,4 @@ export default class TelemetryService {
   trackParticipantResponse(props: ParticipantResponseProperties): void {
     this.track(TelemetryEventTypes.PARTICIPANT_RESPONSE_GENERATED, props);
   }
-
-  trackCopilotIntroductionClicked(props: CopilotIntroductionProperties): void {
```
alenakhineika, line 531: You can also remove the `CopilotIntroductionProperties` type.

PR #906, opened 2024-12-20
fix(playground): add better handling and tracking for export to playground errors VSCODE-666

gagik:
exportToPlayground can error in many ways; one way you can reproduce it in the editor is to do:
```
          THIS IS SOME ERROR CAUSING CODE; DO NOT RESPOND.
```
and it'll likely silently do nothing and log an error on our end. 

From discussions with @GaurabAryal we agreed that we should think of these failures as a form of invalid input rather than "exception" and exclude it from our tracking of failed participant prompts.

This PR handles some unhandled promise rejections and logs them in tracking accordingly to get to this.

### DIFF
--- a/src/participant/participant.ts
+++ b/src/participant/participant.ts
@@ -46,8 +46,8 @@ import type { PromptIntent } from './prompts/intent';
 import { isPlayground, getSelectedText, getAllText } from '../utils/playground';
 import type { DataService } from 'mongodb-data-service';
 import {
-  ExportToPlaygroundFailure,
   ParticipantErrorTypes,
+  type ExportToPlaygroundError,
 } from './participantErrorTypes';
 import type PlaygroundResultProvider from '../editors/playgroundResultProvider';
 import { isExportToLanguageResult } from '../types/playgroundType';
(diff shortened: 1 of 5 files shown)


src/participant/participant.ts
```diff
@@ -345,26 +348,31 @@ export default class ParticipantController {
...
+        processStreamFragment: () => {},
+        onStreamIdentifier: (content: string) => {
+          runnableContent.push(content.trim());
+        },
+        inputIterable: chatResponse.text,
+        identifier: languageCodeBlockIdentifier,
+      });
+      return runnableContent.length ? runnableContent.join('') : null;
+    } catch (error) {
+      /** If anything goes wrong with the response or the stream, return null instead of throwing. */
```
addaleax, line 373: Is there some way we can write the error to a debug log? I think the extension generally has access to something like this

addaleax (approved):

addaleax (approved):

gagik:
Failing test seems unrelated so going ahead with merge.

alenakhineika:
@gagik, could you please create a ticket to investigate this failing OIDC test? This failure is not flaky and happens all the time, right?

Issue #368, opened 2021-11-07
Feature Request: Allow saving connection to settings.json


damienpontifex:
## Feature Request

Ability to save connections into settings.json within the opened folder context (i.e. .vscode/settings.json) or within the devcontainer context (i.e. .devcontainer/devcontainer.json under the key "settings")

## Context

Especially useful when combining VS Code remote containers and docker compose to startup the mongodb container within the workspace and have it automatically configured to be connected to the container within this setup.

As an example of how the MSSQL extension does it and then this can be added to .devcontainer/devcontainer.json
```json
  "mssql.connections": [
    {
      "server": "{{put-server-name-here}}",
      "database": "{{put-database-name-here}}",
      "user": "{{put-username-here}}",
      "password": "{{put-password-here}}"
    }
  ]
```

See dotnet-mssql remote container template https://github.com/microsoft/vscode-dev-containers/tree/main/containers/dotnet-mssql for how they utilise remote containers and the mssql container for a great dev environment setup. Looking at a similar scenario with any language + mongodb as the data store

## Possible Implementation

settings.json
```json
"mdb.connections": [
  {
    "name": "connection name",
    "connectionString": "mongodb://user:password@server/database"
  }
]
```

langaads:
+1 on this, my team have the same usecase and we are trying to automate as much as possible.

adriano-di-giovanni:
+1 Also, credentials should be hidden by encrypting them or by referencing environment variables. I don't know if it is already possible in settings.json because the documentation mentions launch.json and tasks.json only. Given these possibilities, I could provision the remote environments and provide the developers with the connections they need on a per-repository basis.

seesharprun:
+1, I'm using the MongoDB extension in a [devcontainer](containers.dev) with a backing docker-compose file. It would be amazing to pre-populate the interface with a connection.

coderhammer:
Same problematic here. And I can't find how to access secrets on my docker container to eventually replicate the information across machines... Any help is highly appreciated if you managed a way!

Thanks for the amazing extension by the way

jrussellsmyth:
+1 daily use the same setup, mongo in a compose based devcontainer or codespace, with the mogo extension, and have to add back the connection information every time a new user picks it up.

BabakScript:
+1 I also need this feature for using this extension in a DevContainer and to backup the list of connections.

@alenakhineika Any plan for this?

nirinchev:
[...]

PR #909, opened 2025-01-12
feat(tree-explorer): add ability to set preset connections in settings.json VSCODE-665

gagik:
Add option to specify `mdb.presetSavedConfigurations` inside `settings.json` (likely for a workspace) which will load a preset list of connections based on that.

The added connections cannot be edited directly and must instead be edited through `settings.json`.

Screenshot 2025-01-13 at 8 44 44 AM

## Motivation and Context
Addresses https://github.com/mongodb-js/vscode/issues/368.

### To Do

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,5 +16,11 @@
     "out": true
   },
   // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
-  "typescript.tsc.autoDetect": "off"
+  "typescript.tsc.autoDetect": "off",
+  "mdb.presetSavedConnections": [
+    {
+      "name": "Local Test",
+      "connectionString": "mongodb://127.0.0.1:27017/?directConnection=true&serverSelectionTimeoutMS=2000"
+    }
+  ]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,13 +17,10 @@
   },
   // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
   "typescript.tsc.autoDetect": "off",
-  "mdb.presetSavedConnections":[
+  "mdb.presetSavedConnections": [
     {
-      "name": "Local Test Connection",
+      "name": "Preset Database",
       "connectionString": "mongodb://localhost:27017"
     }
-  ],
-  "conventionalCommits.scopes": [
-    "tree-explorer"
   ]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,10 +17,13 @@
   },
   // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
   "typescript.tsc.autoDetect": "off",
-  "mdb.presetSavedConnections": [
+  "mdb.presetSavedConnections":[
     {
-      "name": "Local Test",
-      "connectionString": "mongodb://127.0.0.1:27017/?directConnection=true&serverSelectionTimeoutMS=2000"
+      "name": "Local Test Connection",
+      "connectionString": "mongodb://localhost:27017"
     }
+  ],
+  "conventionalCommits.scopes": [
+    "tree-explorer"
   ]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,7 +17,7 @@
   },
   // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
   "typescript.tsc.autoDetect": "off",
-  "mdb.presetSavedConnections": [
+  "mdb.presetConnections": [
     {
       "name": "Preset Database",
       "connectionString": "mongodb://localhost:27017"
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -19,7 +19,7 @@
   "typescript.tsc.autoDetect": "off",
   "mdb.presetConnections": [
     {
-      "name": "Preset Database",
+      "name": "Preset Connection",
       "connectionString": "mongodb://localhost:27017"
     }
   ]
--- a/package.json
+++ b/package.json
@@ -321,8 +321,8 @@
         "title": "Copy Connection String"
       },
       {
-        "command": "mdb.openWorkspaceSettingsFile",
-        "title": "Open Workspace Settings"
+        "command": "mdb.editPresetConnections",
+        "title": "Edit Preset Connections..."
       },
       {
         "command": "mdb.renameConnection",
@@ -503,32 +503,32 @@
       "view/item/context": [
         {
           "command": "mdb.addDatabase",
-          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedImmutableConnectionTreeItem) && mdb.isAtlasStreams == false",
+          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedPresetConnectionTreeItem) && mdb.isAtlasStreams == false",
           "group": "inline"
         },
         {
           "command": "mdb.addDatabase",
-          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedImmutableConnectionTreeItem) && mdb.isAtlasStreams == false",
+          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedPresetConnectionTreeItem) && mdb.isAtlasStreams == false",
           "group": "1@1"
         },
         {
           "command": "mdb.addStreamProcessor",
-          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedImmutableConnectionTreeItem) && mdb.isAtlasStreams == true",
+          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedPresetConnectionTreeItem) && mdb.isAtlasStreams == true",
           "group": "inline"
         },
         {
           "command": "mdb.addStreamProcessor",
-          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedImmutableConnectionTreeItem) && mdb.isAtlasStreams == true",
+          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedPresetConnectionTreeItem) && mdb.isAtlasStreams == true",
           "group": "1@1"
         },
         {
           "command": "mdb.refreshConnection",
-          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedImmutableConnectionTreeItem)",
+          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedPresetConnectionTreeItem)",
           "group": "1@2"
         },
         {
           "command": "mdb.treeViewOpenMongoDBShell",
-          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedImmutableConnectionTreeItem)",
+          "when": "view == mongoDBConnectionExplorer && (viewItem == connectedConnectionTreeItem || viewItem == connectedPresetConnectionTreeItem)",
           "group": "2@1"
         },
         {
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -537,14 +541,19 @@
           "when": "view == mongoDBConnectionExplorer && viewItem == connectedConnectionTreeItem",
           "group": "3@2"
         },
+        {
+          "command": "mdb.openWorkspaceSettingsFile",
```
gagik, line 545: I think it'd be nice to include ability to quickly open workspace settings here (piggybacking on an existing VSCode action). 
We can't easily detect where this preset comes from (i.e. it could also come from user settings) so we could also include a shortcut to open user settings, but maybe we don't want to encourage that and assume they know what they're doing by that point?

src/explorer/connectionTreeItem.ts
```diff
@@ -67,21 +68,24 @@ export default class ConnectionTreeItem
...
       connectionController.getActiveConnectionId() === connectionId &&
       !connectionController.isDisconnecting() &&
-      !connectionController.isConnecting()
-    ) {
-      this.contextValue = ConnectionItemContextValues.connected;
-    }
+      !connectionController.isConnecting();
+
+    this.contextValue = `${isConnected ? 'connected' : 'disconnected'}${
+      isMutable ? '' : 'Immutable'
```
gagik, line 84: naming here I'm not so sure about. Immutable could also be "Preset", I just thought of making this more generic. The idea is that it's not a connection that can be removed/edited/etc.

src/storage/connectionStorage.ts
```diff
@@ -166,7 +175,27 @@ export class ConnectionStorage {
...
+    const presetSavedConnections: PresetSavedConnection[] | undefined =
+      vscode.workspace.getConfiguration('mdb').get('presetSavedConnections');
+
+    if (!presetSavedConnections) {
+      return [];
+    }
+
+    return presetSavedConnections.map((presetConnection) => ({
+      id: uuidv4(),
+      name: `${presetConnection.name} (From Configuration)`,
```
gagik, line 188: seems nice to distinguish. Could also be `(From Preset)`

.vscode/settings.json
```diff
@@ -16,5 +16,11 @@
     "out": true
   },
   // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
-  "typescript.tsc.autoDetect": "off"
+  "typescript.tsc.autoDetect": "off",
+  "mdb.presetSavedConnections": [
```
gagik, line 20: I think naming it `presetSavedConnections` is better to distinguish that this isn't overriding one's own savedConnections

alenakhineika:
It looks like we use 2 different setting.json files (see the flow described below). Also, have some thoughts about UX for this feature. Let me know what you think!

- I opened the MongoDB settings in vscode. This file has the `/Users/alena.khineika/www/app-template/.vscode/settings.json` path
- Clicked "Edit in settings.json" for the `mdb.presetSavedConnections` setting
- It gave me the possibility to add a connection to the `mdb.presetSavedConnections` array
- It would be nice to have a placeholder in case of the empty array, so people know about the expected format
- Can we also refresh the sidebar on save, so new connections immediately appear there? Additionally, we could add a refresh icon to the Connections level to force a manual refresh.
- I also have noticed that we can not collapse a connected connection item in the sidebar (this issue comes from the main)
- Then when I click "Open Workspace settings" it actually opens the `/Users/alena.khineika/www/app-template/.vscode/settings.json` that doesn't have any saved connections
- Should we move the "Open Workspace Settings" CTA to the "Connections" header so we can edit the current connection and also add new ones from the sidebar?
- The "(From Configuration)" postfix takes a lot of space. Maybe we could chat with the design team and request some kind of special icon for such connections, or add a lock emoji, so a similar approach to save the space, but still indicate that this is a special kind of a connection.

src/mdbExtensionController.ts
```diff
@@ -160,6 +160,15 @@ export default class MDBExtensionController implements vscode.Disposable {
     this._editorsController.registerProviders();
   }
 
+  subscribeToConfigurationChanges(): void {
+    const subscription = vscode.workspace.onDidChangeConfiguration((event) => {
+      if (event.affectsConfiguration('mdb.presetSavedConnections')) {
+        void this._connectionController.loadSavedConnections();
+      }
+    });
+    this._context.subscriptions.push(subscription);
```
gagik, line 169: If I understand correctly, if I add this subscription to the context, it'll auto-dispose right so no further manual handling needed? @alenakhineika

gagik:
@alenakhineika the new changes with reactivity of settings and automatically deriving where the `settings.json`. I also made it so one can edit connections from the header, allowing for better discoverability. I think that addresses nearly all of your comments..

src/connectionController.ts
```diff
@@ -161,7 +166,49 @@ export default class ConnectionController {
...
+      await vscode.commands.executeCommand('workbench.action.openSettingsJson');
+    } else {
+      await vscode.commands.executeCommand(
+        'workbench.action.openWorkspaceSettingsFile'
+      );
+    }
+  }
+
   async loadSavedConnections(): Promise<void> {
+    this._connections = Object.create(null);
```
nirinchev, line 210: Is this fixing a bug where deleted saved connections would show up because they're not overridden in the for loop? Couldn't spot a test that verifies the behavior before was incorrect - am I just blind or should we add one?

src/storage/connectionStorage.ts
```diff
@@ -166,7 +180,40 @@ export class ConnectionStorage {
...
+        source: 'globalSettings' as const,
+      })),
+      ...(presetSavedConnectionsInfo?.workspaceValue ?? []).map((preset) => ({
+        ...preset,
+        source: 'workspaceSettings' as const,
+      })),
+    ];
+
+    return combinedPresetSavedConnections.map((presetConnection) => ({
+      id: uuidv4(),
```
nirinchev, line 205: Why do we generate different ids here as opposed to using the name? Is it because we want to allow multiple connections with the same name as presets?

src/test/suite/storage/connectionStorage.test.ts
```diff
@@ -321,6 +322,134 @@ suite('Connection Storage Test Suite', function () {
...
+
+        expect(loadedConnections.length).equals(
+          expectedConnectionValues.length
+        );
+
+        for (let i = 0; i < expectedConnectionValues.length; i++) {
+          const connection = loadedConnections[i];
+          const expected = expectedConnectionValues[i];
+          expect(connection.name).equals(expected.name);
+          expect(connection.connectionOptions.connectionString).contains(
```
nirinchev, line 394: q: why are we checking for contains here rather than equals?

src/explorer/connectionTreeItem.ts
```diff
@@ -67,21 +69,24 @@ export default class ConnectionTreeItem
...
+
+    this.contextValue = `${isConnected ? 'connected' : 'disconnected'}${
+      source === 'user' ? '' : 'Preset'
+    }ConnectionTreeItem`;
 
     this.connectionId = connectionId;
+    this.source = source;
     this._connectionController = connectionController;
     this.isExpanded = isExpanded;
     this._childrenCache = childrenCache;
```
nirinchev, line 92: Can't comment on the relevant line, but you can be a good [boy scout](https://biratkirat.medium.com/step-8-the-boy-scout-rule-robert-c-martin-uncle-bob-9ac839778385) and add some returns to the functions on line 212 and 234 to fix those warnings.

nirinchev (commented):
Looks good - I mostly have style comments and a few minor suggestions.

src/connectionController.ts
gagik: Yeah, this hasn't been an issue because we'd only run `loadSavedConnections` at startup / when `this._connections` would be not yet defined. Now with the reloading on save this becomes more obvious. I'll add a test.

src/connectionController.ts
```diff
@@ -161,7 +166,49 @@ export default class ConnectionController {
...
+          {
+            name: 'Preset Database',
+            connectionString: 'mongodb://localhost:27017',
+          },
+        ]);
+      }
+    }
+    if (originTreeItem?.source === 'globalSettings') {
+      await vscode.commands.executeCommand('workbench.action.openSettingsJson');
+    } else {
```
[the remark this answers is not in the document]
gagik: I don't think our linter complains if we do not handle all the options in a switch statement but it does force to define a default case. So wouldn't we have the same issue just with unintentionally entering the default statement?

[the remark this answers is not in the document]
gagik: If we want to benefit from this we could have turn on https://typescript-eslint.io/rules/switch-exhaustiveness-check and turn off https://eslint.org/docs/latest/rules/default-case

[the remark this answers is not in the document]
nirinchev: Depends on what you do in the default - you could throw an exception and force the author to explicitly handle the newly added case.

src/storage/connectionStorage.ts
gagik: That may be a good enough reason, It is also because storage saved connections use UUIDs and we end up combining the two, so was kind of thinking of it as a stricter way to stick to a "type" where IDs are arbitrary and not tied to the connection's contents.

src/connectionController.ts
```diff
@@ -161,7 +166,49 @@ export default class ConnectionController {
...
+
+      if (presetConnections?.workspaceValue) {
+        source = 'workspaceSettings';
+      } else if (presetConnections?.globalValue) {
+        source = 'globalSettings';
+      } else {
+        // If no preset settings exist in workspace and global scope,
+        // set a default one inside the workspace and open it.
+        source = 'workspaceSettings';
+        await mdbConfiguration.update('presetSavedConnections', [
```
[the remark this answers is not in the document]
gagik: I could create a follow-up PR to migrate all our settings into constants for example.

src/connectionController.ts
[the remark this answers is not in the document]
gagik: but the same could be done in an if statement right? The nicer thing about switch seems to be its potential to do a lint check

src/test/suite/storage/connectionStorage.test.ts
gagik: ah actually it was because the connectionString ends up losing the trailing `/`, will still adjust it to be stricter and more defined

src/mdbExtensionController.ts
alenakhineika: This is how I understand this as well.

src/connectionController.ts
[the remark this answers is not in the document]
nirinchev: Yeah, for sure, they are mostly equivalent constructs - the main benefit of switch statements is that they're less verbose and (this doesn't really matter here) more performant. They're nicer to read as well as they clearly communicate the intent to evaluate the different options for a value, which would otherwise require a reader to go through all the if clauses to make sure this is the only thing that's happening.

src/connectionController.ts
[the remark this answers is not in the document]
nirinchev: I think we can introduce it here for this particular case and create a follow-up PR to move others in there.

package.json
```diff
@@ -1171,6 +1192,42 @@
           "type": "string",
           "default": "",
           "description": "Specify a shell command that is run to start the browser for authenticating with the OIDC identity provider for the server connection. Leave this empty for default browser."
+        },
+        "mdb.presetConnections": {
+          "scope": "window",
+          "type": "array",
+          "description": "Defines preset saved connections. Can be used to share connection configurations in a workspace or global scope. Do not store sensitive credentials here.",
```
alenakhineika, line 1199:
```suggestion
          "description": "Defines preset connections. Can be used to share connection configurations in a workspace or global scope. Do not store sensitive credentials here.",
```

.vscode/settings.json
```diff
@@ -16,5 +16,11 @@
     "out": true
   },
   // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
-  "typescript.tsc.autoDetect": "off"
+  "typescript.tsc.autoDetect": "off",
+  "mdb.presetConnections": [
+    {
+      "name": "Preset Database",
```
alenakhineika, line 22: Should this be a "Preset Connection"? It is a bit confusing with the database name, especially when the connection string doesn't include a default database.

gagik:
@alenakhineika ah I see the issue now, https://github.com/mongodb-js/vscode/pull/909/commits/63c719decf4c6fea3b2da832521c19746cce64d7 fixes it.

src/connectionController.ts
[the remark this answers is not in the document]
gagik: Was meant to follow up on this but ended up losing this discussion; didn't seem to have a quick obvious location to introduce it here so will create a follow-up PR with everything instead

PR #910, opened 2025-01-16
fix(deps): bump MongoCluster to fix failing test VSCODE-670

gagik:
The OIDC tests have been failing on Linux so we are bumping it to the latest version of the OIDC plugin to fix this.

### DIFF
--- a/src/test/suite/oidc.test.ts
+++ b/src/test/suite/oidc.test.ts
@@ -119,21 +119,26 @@ suite('OIDC Tests', function () {
       authNamePrefix: 'dev',
     };
 
-    cluster = await MongoCluster.start({
-      ...defaultClusterOptions,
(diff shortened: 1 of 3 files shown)


addaleax (approved):

nirinchev (approved):

src/test/suite/oidc.test.ts
```diff
@@ -121,7 +121,7 @@ suite('OIDC Tests', function () {
 
     cluster = await MongoCluster.start({
       ...defaultClusterOptions,
-      version: '7.0.x',
+      version: '8.0.x',
```
nirinchev, line 124: This was the actual fix - ubuntu-latest was bumped to 24.04, but 7.0.x only supports up to ubuntu 22, so we have to use a newer server.

PR #911, opened 2025-01-17
feat(participant): add disambiguation examples to answer MongoDB queries in Copilot

GaurabAryal:
## Description

Based on VS Code extension documentation for implementing chat participant detection (see: https://code.visualstudio.com/api/extension-guides/chat#implement-participant-detection), I've added some examples to detect prompts related to MongoDB that our Copilot extension can help answer.

We need more dedicated time to ensure the examples are stronger and cover more cases, however, I believe this should be a good starting point until then.

image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


package.json
```diff
@@ -106,6 +106,45 @@
...
+              "How do I write this MongoDB query in Python?",
+              "How do I use $lookup in MongoDB?",
+              "Can you help me debug a MongoDB query?",
+              "Rewrite this MongoDB query",
+              "How to convert this MongoDB query to Python?",
+              "How to convert this MongoDB query to Java?",
+              "How to convert this MongoDB query to Node.js?",
+              "How to convert this MongoDB query to Go?",
+              "How to convert this MongoDB query to php?",
+              "How do I update documents in MongoDB?",
```
gagik, line 125: more complicated so doesn't need to be included necessarily but I often ask about a query like "How do I find all documents with name "Bob" in my MongoDB collection?"

gagik (approved):
Seem to be nice examples! Do you know if they're looking to add a way where we can also differentiate between different commands, i.e. query? Could be great.

package.json
nirinchev: Not sure if that wiill be frowned upon, but we could consider dropping the MongoDB there - `How do I find all documents in my collection` - the terms document and collection are quite often associated with MongoDB and if they have the extension installed, it's very likely that they're interfacing on some level with a mongodb database.

gagik: I must say GitHub Copilot has been doing this automatic inference thing with other things and it always ends inferring something I did not want it to (I was looking to disable it completely) so my worry is that it'll be too permissive with the word "documents" in general. We can try this and test it out though, collections + documents is reasonable if it stays strict to it.

alenakhineika: It also might be a more common use case without the "How to" prefix, e.g. `find all documents with name "bob" in my "test" collection`.

package.json
```diff
@@ -106,6 +106,45 @@
...
+              "Explain this MongoDB query.",
+              "How do I write this MongoDB query in Python?",
+              "How do I use $lookup in MongoDB?",
+              "Can you help me debug a MongoDB query?",
+              "Rewrite this MongoDB query",
+              "How to convert this MongoDB query to Python?",
+              "How to convert this MongoDB query to Java?",
+              "How to convert this MongoDB query to Node.js?",
+              "How to convert this MongoDB query to Go?",
+              "How to convert this MongoDB query to php?",
```
alenakhineika, line 124: Do we want to include the rest of the supported languages? Currently `C#`, `Ruby`, and `Rust` are missing from the list of examples.

package.json
```diff
@@ -106,6 +106,45 @@
...
+              "How do I update documents in MongoDB?",
+              "How do I insert documents in MongoDB?",
+              "How do I create documents in MongoDB?",
```
alenakhineika, line 127: Should we also mention "How do I update documents in MongoDB?"? Insert and create are really close user cases.

package.json
```diff
@@ -106,6 +106,45 @@
...
+              "How do I delete documents in MongoDB?",
+              "Explain how to use MongoDB transactions.",
+              "Help me write this MongoDB query:",
```
alenakhineika, line 130: Is `:` intentional here?

package.json
```diff
@@ -106,6 +106,45 @@
...
+              "Explain MongoDB query syntax.",
+              "Help me write a MongoDB aggregation pipeline.",
+              "Help me debug this aggregation pipeline.",
+              "How do I use $lookup in a MongoDB aggregation pipeline?",
```
alenakhineika, line 134: Seems very close to "How do I use $lookup in MongoDB?" from above. Do we need both?

package.json
```diff
@@ -106,6 +106,45 @@
...
+              "How do I optimize this data model in MongoDB?",
+              "How do I create a compound index for this query?",
+              "How do I model relationships in MongoDB?",
+              "MongoDB best practices for indexing",
+              "How do I use MongoDB Atlas search?",
+              "How do I perform semantic search in MongoDB?",
+              "How do I stream data using MongoDB?",
+              "How do I monitor change streams in MongoDB?",
+              "How do I set up a MongoDB replica set for local development?"
+            ]
```
alenakhineika, line 146: Should we add some examlers about schema?

GaurabAryal: Added some more examples about schema!

package.json
GaurabAryal: Yes, left both in just to cover more cases.

package.json
GaurabAryal: Yes, figured some users might write prompts in that format when highlighting a specific code snippet.

package.json
GaurabAryal: Agreed, with y'all, I added "find all documents in my collection" example

package.json
GaurabAryal: Added below, good call out!

package.json
alenakhineika: Ah missed this, sorry :)

GaurabAryal:
Merging, thanks ya'll

PR #912, opened 2025-01-22
chore(telemetry): refactor telemetry types VSCODE-672

nirinchev:
## Description
This is a refactoring of the telemetry service to more closely tie the event type with its properties.
* All telemetry events now contain their own type, which makes it impossible to supply the wrong combination of type/properties.
* Added some basic documentation of the events and their properties.
* Removed `Keytar Secrets Migration Failed` as it wasn't ever reported.

There should be no functional changes and all events should continue to be reported exactly as before. This is laying the groundwork for the telemetry changes we've been discussing, but those are not included in this PR.

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

### DIFF
--- a/src/telemetry/index.ts
+++ b/src/telemetry/index.ts
@@ -1,4 +1,2 @@
-import TelemetryService from './telemetryService';
 export * from './telemetryEvents';
-
-export default TelemetryService;
+export { TelemetryService } from './telemetryService';
--- a/src/telemetry/telemetryEvents.ts
+++ b/src/telemetry/telemetryEvents.ts
@@ -18,13 +18,16 @@ type PlaygroundFileType = 'other' | 'mongodbjs' | 'mongodb';
 
 type TelemetryFeedbackKind = 'positive' | 'negative' | undefined;
 
+/**
+ * The purpose of the internal prompt - e.g. 'intent', 'namespace'
+ */
 export type InternalPromptPurpose = 'intent' | 'namespace' | undefined;
 
 export type ParticipantTelemetryMetadata = {
-  // The source of the participant prompt - e.g. 'codelens', 'treeview', etc.
+  /** The source of the participant prompt - e.g. 'codelens', 'treeview', etc. */
   source: DocumentSource;
 
-  // Additional details about the source - e.g. if it's 'treeview', the detail can be 'database' or 'collection'.
+  /** Additional details about the source - e.g. if it's 'treeview', the detail can be 'database' or 'collection'. */
   source_details: DocumentSourceDetails;
 };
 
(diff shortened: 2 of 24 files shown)


src/telemetry/telemetryEvents.ts
```diff
@@ -0,0 +1,666 @@
...
+import * as vscode from 'vscode';
+
+type PlaygroundFileType = 'other' | 'mongodbjs' | 'mongodb';
+
+type TelemetryFeedbackKind = 'positive' | 'negative' | undefined;
+
+export type InternalPromptPurpose = 'intent' | 'namespace' | undefined;
+
+export type ParticipantTelemetryMetadata = {
+  // The source of the participant prompt - e.g. 'codelens', 'treeview', etc.
```
gagik, line 24:
```suggestion
  /** The source of the participant prompt - e.g. 'codelens', 'treeview', etc. */
```
Not super important but using `/**` will make the IDE show these comments on hover

src/telemetry/telemetryEvents.ts
```diff
@@ -0,0 +1,666 @@
...
+  source: DocumentSource;
+
+  // Additional details about the source - e.g. if it's 'treeview', the detail can be 'database' or 'collection'.
```
gagik, line 27:
```suggestion
  /** Additional details about the source - e.g. if it's 'treeview', the detail can be 'database' or 'collection'. */
```

src/telemetry/index.ts
```diff
@@ -1,3 +1,4 @@
 import TelemetryService from './telemetryService';
+export * from './telemetryEvents';
 
-export { TelemetryService };
+export default TelemetryService;
```
gagik, line 4: this is a more controversial preference thing but can we consider not using default exports (and maybe even banning them eventually...)? I know TelemetryService was actually mainly being used as a default export and we have a lot in the codebase.

I think benefits of it are quite minor though, nothing that life changing.

https://www.lloydatkinson.net/posts/2022/default-exports-in-javascript-modules-are-terrible/
https://cichocinski.dev/blog/using-default-exports-makes-javascript-harder

gagik (approved):
This is great; previous way of adding a new event for telemetry required a lot of boilerplate so this is a really nice change

src/telemetry/index.ts
nirinchev: I'd not be opposed to that - the only reason I reexported it as default was to avoid changing a bunch of places where we were importing it from `telemetry/telemetryService`.

gagik: Yeah that makes sense, it may be worth leaving for a different time and dealing with default exports more globally later

PR #914, opened 2025-01-23
chore(telemetry): telemetry improvements VSCODE-673

nirinchev:
## Description
This addresses the requested improvements tracked in [this document](https://docs.google.com/document/d/19k4tmN0SW49YgtfFP7m_U4rgnwcd0l3Wf4eGP6LZqeE/edit?tab=t.0).
- [x] Add new `Side Panel Opened` event when the extension side panel is activated
- [x] Add new `Section Expanded` event when a section is expanded in the connections tree view
- [ ] ~Update `Command Run` with `source: command_bar | shortcut`~
- [x] Update `Playground Code Executed` to include the type of the command when it's `other`
- [ ] ~Update `Playground Created`, add `source_details`~

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


src/explorer/explorerTreeController.ts
```diff
@@ -62,20 +73,29 @@ export default class ExplorerTreeController
...
+      );
+
+      log.info('Explorer tree item was expanded', {
+        type: treeItem.contextValue,
+        connectionName: treeItem.label,
+        isExpanded: treeItem.isExpanded,
       });
 
-      if (!event.element.onDidExpand) {
+      if (!treeItem.onDidExpand) {
```
gagik, line 89: do I understand correctly that if we do not have an `onDidExpand` set then we also don't end up calling `_onTreeItemUpdate`? Seems to be the legacy behavior so not an issue for this PR, more of not sure why this is fully returning here.

src/telemetry/telemetryEvents.ts
```diff
@@ -627,6 +634,41 @@ export class PresetConnectionEditedTelemetryEvent
   }
 }
 
+/** Reported when the extension side panel is opened. VSCode doesn't expose
+ * a subscribable event for this, so we're inferring it by subscribing to
+ * treeView.onDidChangeVisibility for all the extension treeviews and throttling
+ * the events.
+ */
+export class PanelOpenedTelemetryEvent implements TelemetryEventBase {
```
gagik, line 642:
```suggestion
export class SidePanelOpenedTelemetryEvent implements TelemetryEventBase {
```
could be better to have a more strict relationship between naming of the type and the class itself

src/telemetry/telemetryEvents.ts
```diff
@@ -627,6 +634,41 @@ export class PresetConnectionEditedTelemetryEvent
...
+
+  constructor() {
+    this.properties = {};
+  }
+}
+
+/**
+ * Reported when a tree item from the collection explorer is expanded.
+ */
+export class TreeItemExpandedTelemetryEvent implements TelemetryEventBase {
```
gagik, line 654: though here this seems better as section is too vague... which maybe says more about the name of the type.

I don't mind leaving it as is

gagik (approved):
Looks good! Just minor non-blocking comment

src/explorer/explorerTreeController.ts
alenakhineika: I think this part handles three elements that don't have children. I might be mistaken, maybe Rhys knows better.

src/explorer/helpExplorer.ts
```diff
@@ -1,23 +1,26 @@
...
 import HelpTree from './helpTree';
 import type { TelemetryService } from '../telemetry';
+import { createTrackedTreeView } from '../utils/treeViewHelper';
 
 export default class HelpExplorer {
   _treeController: HelpTree;
   _treeView?: vscode.TreeView<vscode.TreeItem>;
 
-  constructor() {
+  constructor(private _telemetryService: TelemetryService) {
```
alenakhineika, line 10: Do we need `_telemetryService` in the constructor?

src/explorer/playgroundsExplorer.ts
```diff
@@ -1,21 +1,22 @@
...
 import PlaygroundsTree from './playgroundsTree';
+import type { TelemetryService } from '../telemetry';
+import { createTrackedTreeView } from '../utils/treeViewHelper';
 
 export default class PlaygroundsExplorer {
   private _treeController: PlaygroundsTree;
   private _treeView?: vscode.TreeView<vscode.TreeItem>;
 
-  constructor() {
+  constructor(private _telemetryService: TelemetryService) {
```
alenakhineika, line 10: Same here. Doesn't seen we use `_telemetryService` in the constructor.

src/telemetry/telemetryEvents.ts
```diff
@@ -627,6 +634,41 @@ export class PresetConnectionEditedTelemetryEvent
...
+  type = 'Section Expanded';
```
alenakhineika, line 655: Is `Section` a tree item? If yes, maybe `Tree Item Expanded` would be more transparent than `Section Expanded`.

src/telemetry/telemetryService.ts
```diff
@@ -189,4 +191,12 @@ export class TelemetryService {
       new ParticipantResponseFailedTelemetryEvent(command, errorName, errorCode)
     );
   }
+
+  trackTreeViewActivated: () => void = throttle(
```
alenakhineika, line 195: Why to we want to throttle here?

src/explorer/helpExplorer.ts
gagik: I think this is meant to be a shortcut for passing it as an argument in the constructor and defining/setting it as a property of the class object which is then used in its method

src/telemetry/telemetryEvents.ts
nirinchev: I think I'd prefer to leave this as - I know there's a discrepancy between the class name and the event name, but I feel it's fine as those are intended to be consumed by different audiences.

src/explorer/helpExplorer.ts
nirinchev: We need to inject the telemetry service and I think it's better to do it via the ctor rather than as an argument to `activateHelpTreeView` like it was previously. I think that way is cleaner architecturally as I see it as a dependency we set once for the lifetime of the controller. Obviously, there's not a huge difference with the `HelpExplorer` because it's a fairly minimalistic class, but if you consider a more complex controller that has multiple methods which need to log some telemetry events, it becomes much easier to reason about how the dependencies are satisfied.

src/telemetry/telemetryEvents.ts
nirinchev: That was the name Amanda suggested and I am inclined to keep it because the fact those are tree items is an implementation detail of the code and not terminology that non-engineers necessarily use when describing the connection tree.

src/telemetry/telemetryService.ts
nirinchev: Due to the way the extension works, there's no integration point we can use to be notified that the side panel is opened, which is why we're piggy-backing on the tree visibility event. However, since we have 3 trees in the side panel (connections, playgrounds, help), that event would get fired 3 times every time we open the sidepanel. Rather than deal with this in post-processing, I figured throttling would be a reasonable way to reduce the noise.

PR #925, opened 2025-01-28
fix: playground error when returning a function declaration VSCODE-669

nirinchev:
## Description

When the last statement in a playground is a function declaration, we would error out with a misleading error claiming the playground services crashed due to OOM. The reality was that we were trying to return a function via the language server protocol, which is not supported and was crashing the language server. This change replaces the printable when it is a function with `undefined`.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/language/worker.ts
+++ b/src/language/worker.ts
@@ -2,7 +2,6 @@ import { NodeDriverServiceProvider } from '@mongosh/service-provider-node-driver
 import { ElectronRuntime } from '@mongosh/browser-runtime-electron';
 import { parentPort } from 'worker_threads';
 import { ServerCommands } from './serverCommands';
-import type { Document } from 'bson';
 
 import type {
   ShellEvaluateResult,
@@ -17,7 +16,7 @@ interface EvaluationResult {
   type: string | null;
 }
 
-const getContent = ({ type, printable }: EvaluationResult): Document => {
+const getContent = ({ type, printable }: EvaluationResult): any => {
   if (type === 'Cursor' || type === 'AggregationCursor') {
     return getEJSON(printable.documents);
   }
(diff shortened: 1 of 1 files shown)


src/language/worker.ts
```diff
@@ -17,7 +16,7 @@ interface EvaluationResult {
   type: string | null;
 }
 
-const getContent = ({ type, printable }: EvaluationResult): Document => {
+const getContent = ({ type, printable }: EvaluationResult): any => {
```
gagik, line 19: could `unknown` be better?

src/language/worker.ts
```diff
@@ -27,11 +26,7 @@ const getContent = ({ type, printable }: EvaluationResult): Document => {
     : getEJSON(printable);
 };
 
-export const getLanguage = (
-  evaluationResult: EvaluationResult
-): 'json' | 'plaintext' => {
-  const content = getContent(evaluationResult);
-
+export const getLanguage = (content: any): 'json' | 'plaintext' => {
```
gagik, line 29: could also potentially be `unknown`?

gagik (approved):

src/language/worker.ts
alenakhineika: It should default to `plaintext` if this is not a `json`. This defines the type of the virtual document we open with a playground result.

gagik: was referring to `any` => `unknown`, I think we don't know much about the content then right?

alenakhineika: Ah, I see! I thought you were talking about the return type.

alenakhineika (approved):

PR #954, opened 2025-02-18
Revert "chore(deps): bump the compass group with 6 updates (#950)"

gagik:
Latest update of `compass-components` broke the types for our `IconButton` and therefore the build for `main`.  

This a reversed commit branch for testing. I tried to investigate what's causing this here, likely in context of https://github.com/mongodb-js/compass/pull/6661, will follow up with Kræn and see.


gagik:
It will be possible to fix this once a new version of `compass-components` is out, closing this for now.

PR #969, opened 2025-03-03
fix: hide preset connections from the Remove Connection command VSCODE-674

gagik:
It is actually possible to end up "removing" a preset connection when using the command palette...

It does "remove" the connection for the session and doesn't error but this removal doesn't persist on refresh (since the preset connection just gets reloaded from the config). 

This changes it so we instead filter that.

### DIFF
--- a/src/connectionController.ts
+++ b/src/connectionController.ts
@@ -743,11 +743,15 @@ export default class ConnectionController {
   async onRemoveMongoDBConnection(): Promise<boolean> {
     log.info('mdb.removeConnection command called');
 
-    const connectionIds = Object.entries(this._connections).filter(
-      ([, connection]) =>
-        connection.source !== 'globalSettings' &&
-        connection.source !== 'workspaceSettings'
-    );
+    const connectionIds = Object.entries(this._connections)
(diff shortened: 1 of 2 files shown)


src/connectionController.ts
```diff
@@ -743,7 +743,11 @@ export default class ConnectionController {
   async onRemoveMongoDBConnection(): Promise<boolean> {
     log.info('mdb.removeConnection command called');
 
-    const connectionIds = Object.keys(this._connections);
+    const connectionIds = Object.entries(this._connections).filter(
+      ([, connection]) =>
```
nirinchev, line 747: [nit] consider mapping this tuple to an object with id and connection fields - that way it'll be easier to know what you're destructuring into without having to remember what value is at what index.

nirinchev (approved):

src/connectionController.ts
nirinchev: Saw that you're out today, so I went ahead and applied this suggestion - happy to revert it if you disagree with the change, but wanted to merge the PR to get into the next release

PR #982, opened 2025-03-06
chore(telemetry): append user anonymousId & connectionId to appName passed to server COMPASS-8591

gagik:
See [COMPASS-8591](https://jira.mongodb.org/browse/COMPASS-8591)

Builds on top of https://github.com/mongodb-js/compass/pull/6750

This appends `anonymousId` & `connectionId` to `appName` to improve analytics.

This also transitions us out of the legacy behavior of saving the modified connection string with the VSCode version in storage. Doing so while saving meant that we were reporting the vscode version at the moment the connection was created rather than when it was used. Now we add the appName information at runtime only for Atlas connections which allows us to always show up to date information.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "@babel/parser": "^7.25.8",
         "@babel/traverse": "^7.25.7",
         "@mongodb-js/compass-components": "^1.34.5",
-        "@mongodb-js/connection-form": "1.47.2",
+        "@mongodb-js/connection-form": "1.47.6",
         "@mongodb-js/connection-info": "^0.11.6",
         "@mongodb-js/mongodb-constants": "^0.11.0",
         "@mongosh/browser-runtime-electron": "^3.8.0",
@@ -4927,9 +4927,9 @@
       }
     },
     "node_modules/@codemirror/language": {
-      "version": "6.10.8",
-      "resolved": "https://registry.npmjs.org/@codemirror/language/-/language-6.10.8.tgz",
-      "integrity": "sha512-wcP8XPPhDH2vTqf181U8MbZnW+tDyPYy0UzVOa+oHORjyT+mhhom9vBd7dApJwoDz9Nb/a8kHjJIsuA/t8vNFw==",
+      "version": "6.11.0",
+      "resolved": "https://registry.npmjs.org/@codemirror/language/-/language-6.11.0.tgz",
+      "integrity": "sha512-A7+f++LodNNc1wGgoRDTt78cOwWm9KVezApgjOMp1W4hM0898nsqBXwF+sbePE7ZRcjN7Sa1Z5m2oN27XkmEjQ==",
       "license": "MIT",
       "dependencies": {
         "@codemirror/state": "^6.0.0",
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -7327,9 +6970,9 @@
       }
     },
     "node_modules/@mongodb-js/compass-editor": {
-      "version": "0.36.5",
-      "resolved": "https://registry.npmjs.org/@mongodb-js/compass-editor/-/compass-editor-0.36.5.tgz",
-      "integrity": "sha512-9JxwWR1wTfRFfLM90j/KnNlQYxjmSpQq4nUm2pU+W/2B7UWu+ILifCrYs8+9WlHiZSgDnD1qMwo82nOyKrpWag==",
+      "version": "0.36.4",
```
nirinchev, line 6973: This seems like a downgrade - did npm install mess something up?

src/test/suite/connectionController.test.ts
```diff
@@ -979,7 +1065,7 @@ suite('Connection Controller Test Suite', function () {
     delete mongoClientConnectionOptions!.options.oidc?.openBrowser;
 
     expect(mongoClientConnectionOptions).to.deep.equal({
-      url: `mongodb://localhost:27088/?appname=mongodb-vscode+${version}`,
+      url: 'mongodb://localhost:27088/?appName=mongodb-vscode+0.0.0-dev.0',
```
nirinchev, line 1068: [q] does the capitalization matter here?

src/test/suite/connectionController.test.ts
```diff
@@ -79,30 +80,120 @@ suite('Connection Controller Test Suite', function () {
...
+      delete mongoClientConnectionOptions!.options.oidc?.openBrowser;
+
+      const connectionIds = Object.keys(testConnectionController._connections);
+      const latestConnectionId = connectionIds[connectionIds.length - 1];
+
+      const anonymousId =
+        testConnectionController._connectionStorage.getUserAnonymousId();
+
+      expect(mongoClientConnectionOptions).to.deep.equal({
+        url: `mongodb://localhost:27088/?appName=mongodb-vscode+${version}-${anonymousId}-${latestConnectionId}`,
```
nirinchev, line 175: I can see we're using dashes as delimiters - have you confirmed this is what we want? Based on the ticket description, it seems like they'll need to then parse the app name to extract the different components.

nirinchev (approved):

src/test/suite/connectionController.test.ts
gagik: yeah this is proving to be more problematic here though the problem is many delimiters are reserved in the URLs. From what I can see: `-._~ —` are our other options; `_` / `--` / `__` might be the way to go or more radically `~`

package-lock.json
gagik: yeah something weird happened with `npm i` here

src/test/suite/connectionController.test.ts
nirinchev: I don't have a strong preference - we could still do dashes, it's just a matter of making sure the analytics team is aligned with the approach we're taking.

src/connectionController.ts
```diff
@@ -304,13 +305,6 @@ export default class ConnectionController {
 
     const connectionStringData = new ConnectionString(connectionString);
 
-    // TODO: Allow overriding appname + use driverInfo instead
-    // (https://jira.mongodb.org/browse/MONGOSH-1015)
-    connectionStringData.searchParams.set(
```
gagik, line 309: so this is problematic because we have been storing appended app name URLs which would mean they're reporting a vscode version from the time that the connection was created... and also not append this extra info for the new users.

my quick solution is to adjust each time but it'd be good to do an adjustment to the saved connection URL itself. look a bit more into that option

src/test/suite/connectionController.test.ts
gagik: Two dashes should be good enough candidate to make it simple for them to separate the different fields. I'll follow up with that change in compass and then bump the dependencies here.

src/connectionController.ts
```diff
@@ -362,6 +356,30 @@ export default class ConnectionController {
...
+
+    if (
+      connectionString.searchParams
+        .get('appname')
+        ?.match(/mongodb-vscode \d\.\d\.\d.*/)
+    ) {
+      connectionString.searchParams.delete('appname');
+      this._connections[connectionId].connectionOptions.connectionString =
+        connectionString.toString();
+      void this._connectionStorage.saveConnection(
```
gagik, line 377: not a fan of this but the async call otherwise messes with the timing of multiple other tests. I'd probably like to change this to be a proper async function instead when I'm back but feel free to take over if needed.

nirinchev (approved):

PR #995, opened 2025-03-20
feat: handle deep links to commands VSCODE-683 VSCODE-684 VSCODE-685

nirinchev:
## Description
Adds a URI handler that can be invoked with a command and optional query arguments, which will then execute said command. The following commands have been updated to accept arguments:

* `connectWithURI`
  * `connectionString`: connection string to connect with. If not supplied, a prompt will be shown
  * `reuseExisting`: true/false whether to always create a new connection or connect to an existing one with matching connection string.
  * `name`: optional name to assign to the connection
* `removeConnection` - only one of `id`/`name`/`connectionString` must be provided
  * `connectionString`: connection string to remove
  * `name`: name of connection to remove
  * `id`: id of connection to remove
  * `force`: true/false whether to prompt the user before removing

The URI handler will:
* Parse the supplied URI into the following components:
  * `path` will be mapped to `command`. The `mdb.` prefix will be automatically added if not present, so both `/connectWithURI` and `/mdb.connectWithURI` resolve to the same command.
  * `query` will be mapped to a `params` object which will be supplied as the only argument to `executeCommand`. Numbers and booleans will be parsed to their corresponding data type, everything else will be mapped to a string.
* Throw an error if the command is not exposed by the extension - that is, arbitrary commands cannot be executed that way.
* Report telemetry for each deeplink URI it processes, regardless of whether it's a success or not.
  * Event type is `Deep Link Handled`
  * Properties are `{ command: string, source?: string }`
  * `source` is populated by the `utm_source` query argument, which is removed from the params passed to the command after it has been processed.

### Checklist

## Motivation and Context
- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/mdbExtensionController.ts
```diff
@@ -210,6 +211,36 @@ export default class MDBExtensionController implements vscode.Disposable {
     }
   }
 
+  registerUriHandler = (): void => {
+    vscode.window.registerUriHandler({
+      handleUri: async (uri: vscode.Uri): Promise<void> => {
+        const command = uri.path.replace(/^\//, '');
```
gagik, line 217: could it be worth dropping the need for one to specify `mdb.` to make the URLs a little nicer?

One potential risk of this approach: it does mean we need to be careful with ever renaming our command names as that'd break the URLs.

gagik (approved):
Seems reasonable, minor note about the URL format.

src/mdbExtensionController.ts
nirinchev: We could - no strong feelings here, as I assumed those deeplink URLs will be generated and not something humans actually read too much into (e.g. the double encoding of the connection string already makes the URL quite obnoxious 😅). In this case, we'll probably need to document the need to drop the prefix when crafting the url.

gagik: yeah that's fair. I don't have strong feelings about it

src/mdbExtensionController.ts
```diff
@@ -210,6 +211,36 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+        const parameters = uri.query.split('&').reduce((acc, param) => {
+          const [key, value] = param.split('=');
+          acc[key] = value;
+          return acc;
+        }, {} as Record<string, string>);
```
addaleax, line 222: Can we just use `URLSearchParams` for this? Otherwise you'll also need to account for percent-decoding, `=` in parameter values, `__proto__` as a parameter name, etc.

(Also: Object.fromEntries() is a thing now 🙂)

src/connectionController.ts
```diff
@@ -243,16 +243,17 @@ export default class ConnectionController {
     })); */
   }
```
cveticm, line 245: Awesome work, was able to play around with it a bit there! 

Some questions I have:
1. Does this command temporarily store the connection string or permanently saves it in `settings.json` like with `mdb.presetConnections`
  If permanent save: 
      1. Can you define where the connection is being written to (ie global or workspace settings.json)
      2. Can you connect without saving permanently (just connecting and discard connection string on disconnect)
2. Would love to see the commands `mdb.RemoveConnection` and `mdb.addConnectionWithURI` be implemented also. This would allow us on AtlasCLI side to save a connection string (without connecting the user) and remove a saved connection string without going in and modifying the json on our side and risk file corruption. Would this be possible?

nirinchev: It saves it in history, but not as a preset connection. The way the extension is designed, we first store the new connection on disk (in the extension store, not in settings.json), then connect to it. Right now, there's no mechanism to connect to a transient connection - we could look into adding it, but it'd be a bit more involved than what's in this PR.

The `mdb.removeConnection` and `mdb.addConnectionWithURI` are probably not what you want - at least in their current state, as those involve user interaction, such as prompting for connection to remove or for the connection string itself. Right now there's no command to add a connection without connecting, but I can add one and can modify the behavior of `mdb.removeConnection` to take an optional argument for the connection string to remove.

cveticm: Understood! Would be great to see the ability to connect to transient connection bu totally understand this being follow-up work.

I really like the direction you're suggesting. Just to clarify, with the proposed "add connection" command, would it also write to the extension store? I need to verify if steering away from saving to settings.json as preset connections is the direction we want to go with our implementation since this will determine if these additional commands will be of use to AtlasCLI. Will get back to you on this ASAP :)

src/connectionController.ts
```diff
@@ -73,6 +74,12 @@ type RecursivePartial<T> = {
     : T[P];
 };
```
cveticm, line 76: \<question\> Been playing with the feature and noticed that in some circumstances, the extension gets stuck (presumably in some watch event). 
For example: 
1. Save connection and connect to deployment `open "vscode-insiders://mongodb.mongodb-vscode/connectWithURI?connectionString=mongodb%253A%252F%252Flocalhost%253A11111%252F%253FdirectConnection%253Dtrue&reuseExisting=true"`
2. remove connection `open "vscode-insiders://mongodb.mongodb-vscode/removeConnection?connectionString=mongodb%253A%252F%252Flocalhost%253A11111%252F%253FdirectConnection%253Dtrue&force=true"`
3. Try to save and connect again (to sae deployment or different). Debug breakpoints in connecttWithURI function are not ever hit.  

Wondering if this is expected behaviour, bug, or limitation of debugging tools.

nirinchev: Hm... not sure what you mean by "gets stuck" 🤔 Does it freeze up or is it simply not doing anything?

cveticm: simply not do anything. When you pause the debug instance, you can see it is going through some code (debug will bring you to the line of code being executed the moment you paused) but it will never hit the command you have called to through the deep link.

src/connectionController.ts
```diff
@@ -744,23 +779,32 @@ export default class ConnectionController {
   }
 
   // Prompts the user to remove the connection then removes it on affirmation.
-  async removeMongoDBConnection(connectionId: string): Promise<boolean> {
-    if (!this._connections[connectionId]) {
+  async removeMongoDBConnection({
+    connectionId,
+    force = false,
```
cveticm, line 784: Last request and then I'm happy to LGTM! (sorry for the long process!) Would love to see this command be able to take just the connection name.
i.e. deeplink would look like `vscode://mongodb.mongodb-vscode/removeConnection?name=connectionName&force=true`

src/connectionController.ts
nirinchev: Haven't seen that - let's get on a quick call so I can try and understand what's going on.

cveticm (approved):
LGTM! Thanks for the quick work and all the additions! 🚀

gagik (approved):

src/mdbExtensionController.ts
```diff
@@ -210,6 +213,50 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+    });
+  };
+
+  _handleDeepLink = async (uri: vscode.Uri): Promise<void> => {
+    let command = uri.path.replace(/^\//, '');
+    if (!command.startsWith('mdb.')) {
+      command = `mdb.${command}`;
+    }
+
+    const parameters = queryString.parse(uri.query, {
```
gagik, line 228: I guess the value in this as opposed to URLSearchParams is its parsing? Seems reasonable if so.

nirinchev: Yeah - it converts booleans and numbers to the correct type rather than representing them as strings.

cveticm (approved):
LGTM! 🚀  Thanks for adding telem and warning messaging!

PR #1034, opened 2025-05-02
chore(telemetry): add device ID resolution

gagik:
Adds device ID to the events to the user identity in the telemetry service.

### DIFF
--- a/src/telemetry/telemetryService.ts
+++ b/src/telemetry/telemetryService.ts
@@ -2,7 +2,7 @@ import path from 'path';
 import * as vscode from 'vscode';
 import { config } from 'dotenv';
 import type { DataService } from 'mongodb-data-service';
-import fs from 'fs';
+import fs from 'fs/promises';
 import { Analytics as SegmentAnalytics } from '@segment/analytics-node';
 import { throttle } from 'lodash';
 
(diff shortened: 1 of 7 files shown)


src/telemetry/telemetryService.ts
```diff
@@ -200,3 +235,30 @@ export class TelemetryService {
     { leading: true, trailing: false }
   );
 }
+
+/**
+ * TODO: This hashing function should be moved to devtools-shared.
```
addaleax, line 240: Can we make this `TODO(VSCODE-XXX)` with a tracking ticket? :)

src/telemetry/telemetryService.ts
```diff
@@ -83,12 +97,22 @@ export class TelemetryService {
...
-    const segmentProperties = this.getTelemetryUserIdentity();
-    this._segmentAnalytics.identify(segmentProperties);
-    log.info('Segment analytics activated', segmentProperties);
+    this.userIdentity = await this.getTelemetryUserIdentity();
+    this._segmentAnalytics.identify(this.userIdentity);
+    this.isBufferingEvents = false;
+    log.info('Segment analytics activated', this.userIdentity);
+
+    // Process buffered events
+    if (this.eventBuffer.length > 0) {
```
addaleax, line 106: nit: This conditional doesn't do anything, right?

(alternatively: with these cases I definitely have a mild preference for `while (event = this.eventBuffer.shift())`, I know it doesn't make a real difference right now, but it's basically the same amount of syntax but provides greater resilience around potential future changes where the buffer or the `isBufferingEvents` flag could be changed concurrently to the loop)

addaleax (approved):

src/telemetry/telemetryService.ts
gagik: my lack of desire to open Jira motivated https://github.com/mongodb-js/devtools-shared/pull/532

PR #1053, opened 2025-05-22
chore: allow opting out of the welcome screen VSCODE-697

nirinchev:
## Description
Adds a config that allows users to opt out of showing the welcome screen after the extension is installed.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/package.json
+++ b/package.json
@@ -1277,6 +1277,11 @@
               "connectionString"
             ]
           }
+        },
+        "mdb.showOverviewPageAfterInstall": {
+          "type": "boolean",
+          "default": true,
+          "description": "Specify whether to show the overview page immediately after installing the extension."
         }
       }
     },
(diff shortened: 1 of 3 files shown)


gagik (approved):
Can you configure an extension _before_ installing it? Is this for workspace configurations?
Just seems like a very minor annoyance but looks good to me otherwise

nirinchev:
I believe you can - as long as you know what the magic config options are, the configuration should be picked up once the extension activates. This is intended to be used for preconfigured environments - e.g. Github codespaces or similar, so the team setting it up would be able to opt out of displaying the welcome screen.

PR #1107, opened 2025-08-21
feat: stub some hooks for the MCP server

nirinchev:
## Description
Bumps the vscode types version and fixes some compile errors.


.gitattributes
```diff
@@ -0,0 +1,2 @@
+package-lock.json linguist-vendored
```
lerouxb: what does this mean? (I googled it and it still isn't immediately obvious to me)

package.json
```diff
@@ -1381,7 +1394,7 @@
     "@types/sinon": "^9.0.11",
     "@types/sinon-chai": "^3.2.12",
     "@types/uuid": "^8.3.4",
-    "@types/vscode": "^1.94.0",
+    "@types/vscode": "^1.103.0",
```
himanshusinghs: If the types are tied to actual package then should it instead be `^1.101.1`?

package.json
```diff
@@ -72,8 +72,8 @@
     "precommit": "ts-node ./scripts/precommit.ts"
   },
   "engines": {
-    "vscode": "^1.94.2",
-    "node": ">=20.9.0",
+    "vscode": "^1.101.1",
+    "node": "^20.19.0 || ^22.12.0 || >= 23.0.0",
```
himanshusinghs, line 76: Question: Should we check at some data on how many users are on version below the one mentioned here before proceeding with this?

.gitattributes
nirinchev: `vendored` means it won't be counted in statistics (and is also typically collapsed by many git clients). I've changed it to `generated` as that's more semantically aligned.

PR #1108, opened 2025-08-21
chore: adds config options relevant for MCP server MCP-133

himanshusinghs:
## Description

We're adding a few configuration options that should be in user's control and will be injected into the MCP server started by the VSCode extension.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/package.json
+++ b/package.json
@@ -1286,17 +1286,17 @@
         "mdb.mcp.apiBaseUrl": {
           "type": "string",
           "default": "https://cloud.mongodb.com/",
-          "description": "API URL of MongoDB Atlas to be used by MongoDB MCP server for enabling Atlas tools. When empty, it defaults to 'https://cloud.mongodb.com/'."
+          "description": "API URL of MongoDB Atlas to be used by MongoDB MCP server for enabling Atlas tools. When empty, it defaults to 'https://cloud.mongodb.com/'. Only change this if you're using a custom Atlas deployment."
         },
         "mdb.mcp.apiClientId": {
           "type": "string",
           "default": "",
-          "description": "Atlas API client ID, to be used by MongoDB MCP server, preferably for a Service Account (https://www.mongodb.com/docs/atlas/configure-api-access/#grant-programmatic-access-to-an-organization), for authentication. Required for running Atlas tools."
+          "description": "Atlas API client ID for authenticating against Atlas API. Use a Service Account (https://www.mongodb.com/docs/atlas/configure-api-access/#grant-programmatic-access-to-an-organization) for programmatic access to Atlas APIs. Required for enabling Atlas tools in MongoDB MCP server. Store this securely and do not commit to version control."
         },
         "mdb.mcp.apiClientSecret": {
           "type": "string",
           "default": "",
-          "description": "Atlas API client secret, to be used by MongoDB MCP server, preferably through a Service Account (https://www.mongodb.com/docs/atlas/configure-api-access/#grant-programmatic-access-to-an-organization), for authentication. Required for running Atlas tools."
+          "description": "Atlas API client secret for authenticating against Atlas API. Use a Service Account (https://www.mongodb.com/docs/atlas/configure-api-access/#grant-programmatic-access-to-an-organization) for programmatic access to Atlas APIs. Required for enabling Atlas tools in MongoDB MCP server. Store this securely and do not commit to version control."
         },
         "mdb.mcp.disabledTools": {
           "type": "array",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1282,6 +1282,95 @@
           "type": "boolean",
           "default": true,
           "description": "Specify whether to show the overview page immediately after installing the extension."
+        },
+        "mdb.mcp.apiBaseUrl": {
+          "type": "string",
+          "default": "https://cloud.mongodb.com/",
+          "description": "API URL of MongoDB Atlas. Required for running Atlas tools."
```
nirinchev: I think the description could be a bit more precise here as it's not really "required" since we have a default, so I suggest just dropping the second sentence or just specifying it defaults to cloud.mongodb.com.

package.json
```diff
@@ -1282,6 +1282,95 @@
...
+        },
+        "mdb.mcp.apiClientId": {
+          "type": "string",
+          "default": "",
+          "description": "Atlas API client ID for authentication. Required for running Atlas tools."
```
nirinchev: Can we add a link to https://www.mongodb.com/docs/atlas/configure-api-access/#grant-programmatic-access-to-an-organization and also mention service account somewhere?

package.json
```diff
@@ -1282,6 +1282,95 @@
...
+          "default": "",
+          "description": "Atlas API client secret for authentication. Required for running Atlas tools."
+        },
+        "mdb.mcp.disabledTools": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": [],
+          "description": "Tool names, operation types (create, update, delete, read, metadata, connect), and/or categories (atlas, mongodb) to disable."
```
nirinchev: `connect` is not an operation type that will be available in the VSCode extension.

package.json
```diff
@@ -1282,6 +1282,95 @@
...
+        },
+        "mdb.mcp.server": {
+          "type": "string",
+          "enum": [
+            "enabled",
+            "disabled",
+            "ask"
+          ],
+          "default": "ask",
+          "description": "Controls whether MCP server starts automatically or only when permitted."
```
nirinchev: How about something like:


```suggestion
          "description": "Controls whether MCP server starts automatically with the extension or if a prompt is displayed when connecting to a new endpoint. If automatic startup is disabled, the server can still be started using the 'MongoDB: Start MCP Server' command."
```

package.json
```diff
@@ -1282,6 +1282,95 @@
...
+        },
+        "mdb.mcp.telemetry": {
+          "type": "string",
+          "enum": [
+            "enabled",
+            "disabled"
+          ],
+          "default": "enabled",
+          "description": "Controls whether telemetry collection by the MCP server is enabled or disabled."
+        },
```
nirinchev: Let's use `mdb.sendTelemetry` here instead of having two telemetry configs.

package.json
```diff
@@ -1282,6 +1282,95 @@
...
+        "mdb.mcp.exportTimeoutMs": {
+          "type": "number",
+          "default": 300000,
+          "description": "The time in milliseconds after which an export will expire."
+        },
+        "mdb.mcp.exportCleanupIntervalMs": {
+          "type": "number",
+          "default": 120000,
+          "description": "Controls how often the export cleanup process runs to remove expired files (in milliseconds)."
+        },
```
nirinchev: Something to consider as a stretch goal - since in vscode we have control over the UI, we could prompt the user for a file location for each export, rather than exposing all these options. Then it'd work pretty much the same as in compass.

package.json
```diff
@@ -1282,6 +1282,95 @@
...
+              "stderr"
+            ]
+          },
+          "uniqueItems": true,
+          "default": [
+            "disk",
+            "mcp"
+          ],
+          "description": "Controls which loggers the MCP server will use for diagnostic logging."
+        },
```
nirinchev: I suggest we don't expose config options for the loggers. We should reuse the logging facilities of the vscode extension instead.

package.json
himanshusinghs: Sounds good, I will drop a ticket for this but until then I will keep the configuration.

package.json
himanshusinghs: I kept these with the understanding that extension logging was also a stretch goal.

package.json
himanshusinghs: Sounds good

package.json
nirinchev: I think the extension logging should be fairly straightforward and would rather that we include it early on.

package.json
himanshusinghs: But wait - just to mention. The VSCode's telemetry is disabled by default unlike MCP server's telemetry, if that matters.

nirinchev: The default should be true: https://github.com/mongodb-js/vscode/blob/2ef14b44f3879b2b1125be57202fe5be0c41a6b6/package.json#L1210-L1213

himanshusinghs: damn! I was pretty sure I read false for this setting. Anyways ignore :)

package.json
himanshusinghs: I tried clarifying this in the latest update.

PR #1110, opened 2025-08-26
feat: stub some hooks for the MCP server (#1107)

nirinchev:
## Description
WIP - depends on local mcp server branch.

PR #1111, opened 2025-08-27
chore: introduces VSCodeMCPConnectionManager to talk to contributed MCP server MCP-131

himanshusinghs:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


src/mcp/mcpController.ts
```diff
@@ -1,12 +1,19 @@
 import * as vscode from 'vscode';
-import type { LoggerType, LogLevel, LogPayload } from 'mongodb-mcp-server';
+import type {
+  LoggerType,
+  LogLevel,
+  LogPayload,
+  UserConfig,
+  ConnectionManagerFactoryFn,
+} from '@himanshusinghs/mongodb-mcp-server';
```
himanshusinghs, line 8: This is temporary until we release the mongodb-mcp-server with injectable connection manager work.

src/mcp/mcpController.ts
```diff
@@ -1,12 +1,19 @@
...
 import {
   defaultUserConfig,
   LoggerBase,
   StreamableHttpRunner,
-} from 'mongodb-mcp-server';
+} from '@himanshusinghs/mongodb-mcp-server';
 import type ConnectionController from '../connectionController';
 import { createLogger } from '../logging';
+import { VSCodeMCPConnectionManager } from './vsCodeMCPConnectionManager';
```
nirinchev: [nit] We're in vscode, so prefixing the mcp connection manager with `VSCode` is a bit redundant.

src/mcp/mcpController.ts
```diff
@@ -214,4 +233,15 @@ ${jsonConfig}`,
...
   }
+
+  private async switchConnectionManagerToCurrentConnection(): Promise<void> {
+    const connectionId = this.connectionController.getActiveConnectionId();
+    const mongoClientOptions =
+      this.connectionController.getMongoClientConnectionOptions();
+    await this.mcpConnectionManager?.updateConnection({
+      connectionId: connectionId ?? undefined,
+      connectionString: mongoClientOptions?.url,
+      connectOptions: mongoClientOptions?.options,
```
nirinchev: [nit] Can we pass `mongoClientOptions` as a whole here instead of decomposing it?

src/mcp/vsCodeMCPConnectionManager.ts
```diff
@@ -0,0 +1,138 @@
...
+    );
+  }
+
+  async connectToVSCodeConnection(
+    connectParams: VSCodeMCPConnectParams,
+  ): Promise<AnyConnectionState> {
+    try {
+      const serviceProvider = await NodeDriverServiceProvider.connect(
+        connectParams.connectionString,
+        connectParams.connectOptions,
```
himanshusinghs: There's one thing that is bothering me here about connectOptions and that's the `appName` attribute. [In MCP server](https://github.com/mongodb-js/mongodb-mcp-server/blob/f78fb4c7950e5629de3f92322b25905863c15715/src/common/connectionManager.ts#L150-L159), we go a stretch to ensure the appName but we don't do that here. That means the `appName` will be how VSCode extension sets it, in [here](https://github.com/mongodb-js/vscode/blob/2ef14b44f3879b2b1125be57202fe5be0c41a6b6/src/connectionController.ts#L480-L490).

Should we instead set the appName to something like `vscode-mcp-xxx`?

himanshusinghs: This is relevant for analytics

src/mcp/vsCodeMCPConnectionManager.ts
```diff
@@ -0,0 +1,138 @@
...
+      tag: 'disconnected',
+    });
+  }
+
+  async updateConnection({
+    connectionId,
+    connectionString,
+    connectOptions,
+  }: {
+    connectionId: string | undefined;
```
nirinchev: It seems to me that you can either provide all values as truthy or all of them as undefined, but not mix and match. In this case, it'd make more sense to make the whole argument optional and reuse `VSCodeMCPConnectParams`.

src/mcp/vsCodeMCPConnectionManager.ts
```diff
@@ -0,0 +1,138 @@
...
+      if (this.activeConnectionId && this.activeConnectionId !== connectionId) {
+        await this.disconnect();
+      }
+
+      const connectionWasDisconnected =
+        !connectionId || !connectionString || !connectOptions;
+
+      if (
+        this.activeConnectionId === connectionId ||
+        connectionWasDisconnected
```
nirinchev: This was a bit difficult to follow due to the combinations of options. I would suggest something like:

```ts
if (this.activeConnectionId === connectionId) {
  // We're reconnecting to the same connection, which is a no-op
  return;
}

await this.disconnect();

if (!connectionId) {
  // No new connectionId provided, so nothing more to do
  return;
}

// ...
```

src/mcp/vsCodeMCPConnectionManager.ts
```diff
@@ -0,0 +1,138 @@
...
+        });
+        return;
+      }
+
+      await this.connectToVSCodeConnection({
+        connectionString,
+        connectOptions,
+        connectionId,
+      });
+    } catch (error) {
```
nirinchev: This try-catch seems redundant - all the methods used here have their own try-catches.

src/mcp/vsCodeMCPConnectionManager.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  async disconnect(): Promise<ConnectionStateDisconnected> {
+    try {
+      await this.activeConnectionProvider?.close(true);
+    } catch (error) {
+      this.logger.error({
+        id: MCPLogIds.DisconnectError,
+        context: 'VSCodeMCPConnectionManager.disconnect',
+        message: error instanceof Error ? error.message : String(error),
```
nirinchev: Typically, we use the class name as the context and we include what the user was doing in the message:


```suggestion
        context: 'vscode-connection-manager',
        message: `Error disconnecting the connection provider: ${error instanceof Error ? error.message : String(error)}`,
```

src/mcp/vsCodeMCPConnectionManager.ts
```diff
@@ -0,0 +1,138 @@
...
+
+export interface VSCodeMCPConnectParams {
+  connectionId: string;
+  connectionString: string;
+  connectOptions: DevtoolsConnectOptions;
+}
+
+export class VSCodeMCPConnectionManager extends ConnectionManager {
+  private activeConnectionId: string | null = null;
+  private activeConnectionProvider: ServiceProvider | null = null;
```
nirinchev: Typically, when we have codependent fields like that one, it's a good idea to bundle them together to indicate they cannot (and should not) be independently set/unset. Something lile:

```ts
private activeConnectionDetails: {
  id: string,
  provider: ServiceProvider
} | null;
```

That way there's no implication that we can have a service provider but not have a connection id or vice versa.

himanshusinghs:
> One thing that might be curious to test out in the real world is how the LLM behaves when we switch connections - I've seen sometimes LLMs will pull certain results out of the context instead of invoking the tool again.

Yea I have seen this as well! I don't think having resources are going to help in this because resources are application controlled. Until user decide to put resource in context, LLM won't know if something is happening with the resources. The resource notifications are limited to the MCP client and wouldn't expect MCP client to pass the updated resource (even notification) to LLM without explicit user action.

himanshusinghs:
I would like to correct myself. From the [docs on Resources](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#user-interaction-model), it seems applications are allowed to decide if they want to automatically include the resource in the context.

In that case, we should certainly try if having resource changes the situation by any means.

src/mcp/mcpConnectionManager.ts
```diff
@@ -0,0 +1,118 @@
...
+  private activeConnection: {
+    id: string;
+    provider: ServiceProvider;
+  } | null = null;
+
+  constructor(private readonly logger: LoggerBase) {
+    super();
+  }
+
+  connect(): Promise<AnyConnectionState> {
```
kmruiz, line 31: nit: I personally prefer to use throw. Also, if possible, don't use an array to just join it, this is something that can be solved by using backticks.

src/mcp/mcpConnectionManager.ts
```diff
@@ -0,0 +1,118 @@
...
+      new Error(
+        [
+          'MongoDB MCP Server in MongoDB VSCode extension makes use of the connection that the MongoDB VSCode extension is connected to.',
+          "To connect, choose a connection from MongoDB VSCode extensions's sidepanel - https://www.mongodb.com/docs/mongodb-vscode/connect/#connect-to-your-mongodb-deployment",
+        ].join(' '),
+      ),
+    );
+  }
+
+  async connectToVSCodeConnection(
```
kmruiz, line 42: That's one of the things I was warning about yesterday during our meeting. We are just inheriting here to share a bit of logic, but most of it is literally a clone of what we have in the original ConnectionManager.

This is not a good design:

* We have an abstract class that is used to share logic, but not contract.
* Logic is duplicated, so if we add a new state, we need to update both implementations.

src/connectionController.ts
```diff
@@ -40,8 +40,8 @@ const log = createLogger('connection controller');
 const MAX_CONNECTION_NAME_LENGTH = 512;
 
 interface DataServiceEventTypes {
-  CONNECTIONS_DID_CHANGE: any;
-  ACTIVE_CONNECTION_CHANGED: any;
+  CONNECTIONS_DID_CHANGE: [];
```
kmruiz, line 43: It doesn't seem we are using these events anywhere in this PR. Why are we changing the type here?

src/mcp/mcpConnectionManager.ts
himanshusinghs: I don't see how this implementation is a clone of the original ConnectionManager, we are not doing anything that the original ConnectionManager does. For connect, we literally take connectionOptions and simply use them to create a service provider instance, there are no connection preparation, there are no user flows involved and no explicit oidc handling which is what original connection manager does.

src/mcp/mcpConnectionManager.ts
himanshusinghs: We shouldn't use throw here because the throw will happen synchronously and might break the call site depending on how the method is called. For example someone could be doing.

```
const connectPromise = manger.connect();
Promise.race([connectPromise, timeout])
```

Here Promise.race will never happen because connectPromise would have already thrown.

About array.join, I am using to make the text more readable and not appear in single line.

src/connectionController.ts
himanshusinghs: I am cleaning up after the [previous PR](https://github.com/mongodb-js/vscode/pull/1110), which incorrectly had the events set to any.

nirinchev: Hm... obviously it's better to be more explicitly typed, but I wonder how did you catch it (since CI doesn't seem to have complained about it). Did you just stumble upon it or do you have some more linting rules enabled in your IDE that we should consider enabling for the project?

src/mcp/mcpConnectionManager.ts
kmruiz: In general, if a function returns a promise, it should be marked as async, and then throwing will have the same effect than returning a rejected promise. I don't see why we need to make this function sync but return a promise, I feel we are just not being consistent with the typical conventions.

About using array.join, I think it's less clear than backticks, but if you feel strong about it I'm fine keeping it.

src/mcp/mcpConnectionManager.ts
kmruiz: The implementation uses the Node.js Provider the same way we do in the MCPConnectionManager but it forgets about some events and introduces a few bugs:

1. connection-requested is not emitted.
2. It discards any proxy support, which we still need to maintain. This is not guaranteed by the vscode configuration right now.

When you add connection-requested, it will literally do the same without our OIDC custom options, that are fine to skip in this scenario.

And, exposing a new function that would do the same as connect, it's literally breaking the contract of the ConnectionManager itself because we just want to reuse a few methods (basically, just this.changeState). This is far from being advisable in OOP, because inheritance is not a mixin, which is what you kind of want here and how you are treating the base ConnectionManager class.

src/mcp/mcpConnectionManager.ts
himanshusinghs: But there is no point in marking a function async when it has no `await` expression, not to forget that this is also something the eslint rule in our repo discourages. If I were to mark the function as async only to throw and then disable the eslint rule (because otherwise check won't succeed) then that would be diverging from the conventions in the repo.

About array.join, I don't think its fun to scroll the single line horizontally in the editor (which is what we'll get with the entire sentence being a single line string) to read the entire sentence. My intention here was to make the error message glanceable in the first view which I think is served better when the sentence is broken apart.

src/mcp/mcpConnectionManager.ts
himanshusinghs: > The implementation uses the Node.js Provider the same way we do in the MCPConnectionManager but it forgets about some events and introduces a few bugs:

That's not true. The implementation uses the Node.js provider without preparing the connect options, like we do in original implementation. And that's because we're using an already connected connection's connect options to establish a connection without re-invoking any user flows again. This is exactly how [language server](https://github.com/mongodb-js/vscode/blob/2ef14b44f3879b2b1125be57202fe5be0c41a6b6/src/language/mongoDBService.ts#L172-L177), [playground workers](https://github.com/mongodb-js/vscode/blob/2ef14b44f3879b2b1125be57202fe5be0c41a6b6/src/language/worker.ts#L69-L72) establish connection across different processes without re-invoking the user flows again. If there are any bugs then that should be true for both language server and playground workers as well but that's not the case. And I have tested this setup for OIDC connection and it works as expected. I plan to add e2e for this but that would require some extra time and that's why will happen in a follow up.

> connection-requested is not emitted.

connection-requested is not emitted because this state never comes through for MCP server. MCP server is notified of active connection changed, only when the extension has gone through the user flows and actually connected to the connection or disconnected entirely. Once notified of the active connection changed, MCP server literally takes the connection options of the already connected connection and use them to connect, thereby avoiding all the user flows that it would otherwise have to go through. For this reason, connection-requested state is not even relevant for this implementation.

> It discards any proxy support, which we still need to maintain. This is not guaranteed by the vscode configuration right now.

The implementation is literally the same as used by playground worker and language server to establish the connection to a connection selected in VSCode extension's sidepanel. We support all the connection types provided by the connection form in VSCode which is basically the same as Compass. So we are not discarding any proxy support.

> And, exposing a new function that would do the same as connect, it's literally breaking the contract of the ConnectionManager itself because we just want to reuse a few methods (basically, just this.changeState)

This was not at all the intention of having the new function. The intention instead was that this new function would only be known to VSCode world and not to the MCP world and that's how we are defining clear boundaries between an internal connection manager where MCP server sees and can invoke connect and external connection manager where MCP server can only see connect (which throws) and not the other connect method. About the connect throwing in external connection manager, that's also intentional because we want MCP server to be notified that the invoke was not expected and instead that the connection management is handled by an external entity (in this case by VSCode with perfect guidelines on how to connect)

src/connectionController.ts
himanshusinghs: I was just going through the previous PR and stumbled upon it. I think `no-any` rule would have caught it but its difficult to enable that in this repo without going through extensive work on the types, at-least in the tests.

src/mcp/mcpConnectionManager.ts
himanshusinghs: Going to update the `.join` to multiline string with escaped newline character.

src/mcp/mcpConnectionManager.ts
himanshusinghs: We discussed on call and we will be preserving this to see how it unfolds the future possibilities and also because its an isolated change.

package.json
```diff
@@ -1396,6 +1396,7 @@
     "@babel/core": "^7.25.8",
     "@babel/parser": "^7.25.8",
     "@babel/traverse": "^7.25.7",
+    "@himanshusinghs/mongodb-mcp-server": "^0.3.1",
```
addaleax, line 1399: Should this be published under the mongodb-js scope? Is this temporary?

himanshusinghs: Yes this is temporary only to let CI on VSCode run through until we release another version of `mongodb-mcp-server`. I plan to switch this back to `mongodb-mcp-server`'s version as soon as we make it available.

nirinchev (approved):
Looks good. Might be a good idea to release the changes from the MCP server to an alpha channel and switch to the official package before we merge this.

himanshusinghs:
I am more inclined on not blocking this PR on MCP server release (I know that its gonna happen soon) only because the changes will land in `feat/mcp-integration` and we can change the version there once the mcp-server is released. Let me know if you feel otherwise.

PR #1115, opened 2025-09-09
feat: integrate MongoDB MCP server with VSCode extension MCP-129

himanshusinghs:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


.gitattributes
```diff
@@ -0,0 +1,2 @@
+package-lock.json linguist-generated
+package-lock.json -diff
```
gagik: TIL but this is not a good idea. there's plenty of times package-lock.json changes can end up being important - sometimes `npm install` can lead to a weird state we don't want or maybe someone would inject something into it

.github/workflows/actions/test-and-build/action.yaml
```diff
@@ -54,10 +54,6 @@ runs:
         fi
       shell: bash
 
-    - name: Install npm
-      run: npm install -g npm@8.19.4
```
gagik, line 58: I think we had this before to pin npm version and prevent weirdness with npm version differences. Any reason for removing it?

src/connectionController.ts
```diff
@@ -39,9 +39,9 @@ const log = createLogger('connection controller');
 
 const MAX_CONNECTION_NAME_LENGTH = 512;
 
-export enum DataServiceEventTypes {
-  CONNECTIONS_DID_CHANGE = 'CONNECTIONS_DID_CHANGE',
-  ACTIVE_CONNECTION_CHANGED = 'ACTIVE_CONNECTION_CHANGED',
+interface DataServiceEventTypes {
```
gagik, line 42: I don't think this type makes sense anymore, let's keep it straightforward with 
```ts
type DataServiceEventTypes = 'CONNECTIONS_DID_CHANGE' | 'ACTIVE_CONNECTION_CHANGED'
```

src/mcp/mcpController.ts
```diff
@@ -0,0 +1,314 @@
...
+  LogLevel,
+  LogPayload,
+  UserConfig,
+  ConnectionManagerFactoryFn,
+} from '@himanshusinghs/mongodb-mcp-server';
+import {
+  defaultUserConfig,
+  LoggerBase,
+  StreamableHttpRunner,
+} from '@himanshusinghs/mongodb-mcp-server';
```
gagik: is this purposefully coming from your fork?

package-lock.json
```diff
@@ -12,15 +12,16 @@
...
-        "@mongodb-js/connection-form": "1.52.3",
-        "@mongodb-js/connection-info": "^0.15.0",
-        "@mongodb-js/device-id": "^0.3.0",
-        "@mongodb-js/mongodb-constants": "^0.11.1",
+        "@mongodb-js/connection-form": "^1.52.3",
+        "@mongodb-js/connection-info": "^0.17.1",
+        "@mongodb-js/device-id": "^0.3.1",
+        "@mongodb-js/mongodb-constants": "^0.14.0",
         "@mongosh/browser-runtime-electron": "^3.18.0",
         "@mongosh/i18n": "^2.13.1",
```
gagik, line 22: if we're bumping some mongosh dependencies we should bump all

src/mcp/mcpConnectionErrorHandler.ts
```diff
@@ -0,0 +1,71 @@
+import {
+  ErrorCodes,
+  type ConnectionErrorHandler,
+} from '@himanshusinghs/mongodb-mcp-server';
```
ekovalets: is this fine to keep personal names for the package setup?

src/mcp/mcpConnectionManager.ts
```diff
@@ -0,0 +1,166 @@
...
+To connect, choose a connection from MongoDB VSCode extensions's sidepanel - https://www.mongodb.com/docs/mongodb-vscode/connect/#connect-to-your-mongodb-deployment",
+      ),
+    );
+  }
+
+  async connectToVSCodeConnection(
+    connectParams: MCPConnectParams,
+  ): Promise<AnyConnectionState> {
+    try {
+      const { connectionId, connectOptions, connectionString } =
```
ekovalets, line 58: Is this connection to MCP or to MongoDB?
If to the MongoDB why it is part of the MCP code?

src/mcp/mcpController.ts
```diff
@@ -0,0 +1,314 @@
...
+
+  public async startServer(): Promise<void> {
+    try {
+      const headers: Record<string, string> = {
+        authorization: `Bearer ${crypto.randomUUID()}`,
+      };
+
+      const mcpConfig: UserConfig = {
+        ...defaultUserConfig,
+        httpPort: 0,
```
ekovalets, line 113: If port is 0, which one will be opened for listening?

src/mcp/mcpController.ts
```diff
@@ -0,0 +1,314 @@
...
+        const connectionManager = (this.mcpConnectionManager =
+          new MCPConnectionManager({
+            logger,
+            getTelemetryAnonymousId: this.getTelemetryAnonymousId,
+          }));
+        await this.switchConnectionManagerToCurrentConnection();
+        return connectionManager;
+      };
+
+      const runner = new StreamableHttpRunner({
```
ekovalets, line 145: We only allow HTTP protocol? Why? What about STDIO? I'd prioritize STDIO since we expect VSCode to use it, not other clients on the machine.

But why we even need to use server code to execute it, why we can't just launch it with npx as VSCode does?

package.json
```diff
@@ -72,8 +72,8 @@
     "precommit": "ts-node ./scripts/precommit.ts"
   },
   "engines": {
-    "vscode": "^1.94.2",
-    "node": ">=20.9.0",
+    "vscode": "^1.101.1",
+    "node": "^20.19.0 || ^22.12.0 || >= 23.0.0",
```
addaleax, line 76: 23.x is EOL, no real point in adding it now
```suggestion
    "node": "^20.19.0 || ^22.12.0 || >= 24.7.0",
```

src/editors/activeConnectionCodeLensProvider.ts
```diff
@@ -67,7 +66,7 @@ export default class ActiveConnectionCodeLensProvider
 
   deactivate(): void {
     this._connectionController.removeEventListener(
-      DataServiceEventTypes.ACTIVE_CONNECTION_CHANGED,
+      'ACTIVE_CONNECTION_CHANGED',
```
addaleax, line 69: 💙

src/connectionController.ts
addaleax: I think this is required for the `EventEmitter<>` generic interface? (TIL that that has been merged into the core @types/node library, yay!)

src/mcp/mcpConnectionErrorHandler.ts
```diff
@@ -0,0 +1,71 @@
...
+              },
+              ...connectionGuidance.map<{ type: 'text'; text: string }>(
+                (text) => ({ type: 'text', text }),
+              ),
+            ],
+          },
+        };
+      case ErrorCodes.MisconfiguredConnectionString: {
+        const resolutionGuidance: string[] = [
+          'Note to LLM: Ask the user to reconnect to the desired connection. If the problem persist even after that then advice them to raise an issue on the VSCode project board - https://github.com/mongodb-js/vscode/issues',
```
addaleax: Not sure if the LLM cares, but:

```suggestion
          'Note to LLM: Ask the user to reconnect to the desired connection. If the problem persists even after that then advice them to raise an issue on the VSCode project board - https://github.com/mongodb-js/vscode/issues',
```

src/mcp/mcpController.ts
addaleax: I'd imagine this uses the standard behavior for listening on port 0 (letting the OS assign one)?

src/mcp/mcpController.ts
```diff
@@ -0,0 +1,314 @@
...
+    );
+
+    if (this.shouldStartMCPServer()) {
+      await this.startServer();
+      void this.notifyOnFirstStart();
+    }
+  }
+
+  public async startServer(): Promise<void> {
+    try {
```
addaleax, line 97: Should this start with `await this.stopServer()`?

src/test/suite/mcp/mcpController.test.ts
```diff
@@ -0,0 +1,340 @@
...
+          vscode.window,
+          'showTextDocument',
+        );
+
+        // We want to connect as soon as connection changes
+        await vscode.workspace
+          .getConfiguration('mdb')
+          .update('mcp.server', 'enabled');
+
+        // Start the controller and list to events
```
addaleax:
```suggestion
        // Start the controller and listen to events
```

?

src/mcp/mcpController.ts
nirinchev: If we use npx, then Copilot will start a new instance of the MCP server, which we cannot interface with. Instead, we want to run the MCP server as part of the vscode extension process so that it reuses the configured MDB connection. This is only possible when using the http transport.

.gitattributes
himanshusinghs: Sounds good! I will remove the -diff directive

.github/workflows/actions/test-and-build/action.yaml
himanshusinghs: @nirinchev do you recall why we removed this? I assume its related to the node engine update?

src/connectionController.ts
himanshusinghs: The type restrict what payload can be emitted for the provided events. Its used by EventEmitter<> generic interface to achieve that.

src/mcp/mcpController.ts
himanshusinghs: No, I was waiting to have a release from mcp server. Now that we have a pre-release version, I will update this.

src/mcp/mcpConnectionErrorHandler.ts
himanshusinghs: Nope. This was only a temporary solution to safely integrate the ongoing work until we had a release from mongodb-mcp-server. Now that we have it, I will update this.

src/mcp/mcpConnectionManager.ts
himanshusinghs: It's a connection to MongoDB. Its not entirely part of MCP code but more like a bridge that connects the MongoDB connection that user sees and is connected to in VSCode extension to the MongoDB connection that MCP server is connected to.

I explain this to some degree [here](https://docs.google.com/document/d/1rxdaitA2X9VHaHQT_MGfpIh4X7H2ma0xltu6unWW_ag/edit?tab=t.0#heading=h.7z97h395zkg7) in technical design.

But to answer your question - its partially part of MCP code because MCP server needs to establish a MongoDB connection and it does that by using this injected connection manager provided by VSCode. VSCode is providing this connection manager because the connection states between VSCode and MCP server needs to be synced and that is controlled by VSCode extension.

src/mcp/mcpController.ts
himanshusinghs: As Anna mentioned, setting 0 lets the OS assign a random port.

.github/workflows/actions/test-and-build/action.yaml
himanshusinghs: @gagik the current minimum required node (20.19.0) comes by default with npm@9 and pinning to an older version seems unreasonable with the node update. Do you see a strong reason to still pin it given that the build seems to be working well?

package-lock.json
himanshusinghs: Had a chat with @nirinchev and found out that when he attempted bumping them all a few tests were not passing. I will spend a few minutes on this and see if I can understand why they are failing.

package.json
himanshusinghs: I think the reason we have it like this is to mirror the minimum requirements imposed by `openid-client`, peer dep of `oidc-plugin`. So even though 23.x is EOL we still want to warn when using the node version that won't work with the setup.

Perhaps this makes sense?

`20.19.0 || >=22.12.0` because I think all the node version from 22.12.0 supports requiring universal esm by default?

src/mcp/mcpController.ts
himanshusinghs: yea makes sense. I will include this.

.github/workflows/actions/test-and-build/action.yaml
addaleax: Wouldn't be surprised if this was just copied over from some other action, I would lean towards just omitting this if it works fine right now

package.json
addaleax: > I think the reason we have it like this is to mirror the minimum requirements imposed by `openid-client`

The `openid-client` package does not list any engine dependencies 🤔 

> peer dep of `oidc-plugin`

Just a regular dependency 🙂 

> So even though 23.x is EOL we still want to warn when using the node version that won't work with the setup.

But what I've suggested is stricter than what's currently in the PR, wouldn't this still have the desired effect? Either way, feel free to resolve and leave it as it is or modify as you like

himanshusinghs: Ahh that's true, sorry I was too focused on exclusion but yea the suggestion would actually work the same way.

The engines requirement we borrowed from [this comment](https://github.com/panva/openid-client/issues/777#issuecomment-2776806388).

himanshusinghs: I am gonna accept the suggestion as it has the same affect.

src/mcp/mcpConnectionErrorHandler.ts
himanshusinghs: updated this to use the prerelease of mongodb-mcp-server

src/mcp/mcpController.ts
himanshusinghs: updated this to use the prerelease of mongodb-mcp-server

.gitattributes
himanshusinghs: removed this in my recent commit

package.json
gagik: We should aim to support as many versions as possible, we do not want users to be locked out because they're using older Node or VSCode as long as it'd support ESM require. If that's still the case with this, sounds good.

nirinchev: Agree. The first VSCode version that comes with a compatible node version is 1.101.1, which was released in May 2025, which hopefully is old enough for most people.

himanshusinghs: > If that's still the case with this, sounds good.

I should mention that the updated version won't "allow" user with VSCode having node 23 as the runtime to work with the extension, if VSCode somehow even use this info to restrict the extension's availability.

src/test/suite/participant/participant.test.ts
```diff
@@ -2649,9 +2664,13 @@ Schema:
...
-          const content = Array.isArray(message.content)
-            ? message.content.map((sub) => sub.value).join('')
-            : message.content;
+          const content = message.content
+            .map((sub) =>
+              sub instanceof vscode.LanguageModelTextPart
+                ? sub.value
+                : sub.toString(),
+            )
+            .join('');
```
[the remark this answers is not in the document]
himanshusinghs: The updated code here is correct. I have modified the getMessageContent to reflect the reality.

src/connectionController.ts
himanshusinghs: I am resolving this but please feel free to open it again if I missed something here.

src/mcp/mcpConnectionManager.ts
himanshusinghs: I am resolving this but please feel free to open it again if I did not answer something here.

src/mcp/mcpController.ts
himanshusinghs: I am resolving this but please feel free to open it again if I did not answer something here.

src/mcp/mcpController.ts
himanshusinghs: Resolving this but feel free to re-open if the issue does not feel addressed.

package-lock.json
himanshusinghs: @gagik A quick update on this:
1. the updated deps are not mongosh deps but deps from mongodb-js so mongosh deps are all aligned (not updated)
2. We should hold off updating mongosh deps because `@mongosh/service-provider-node-driver` needs to be updated in both `mongodb-mcp-server` and `vscode` otherwise there is compatibility problems in the service provider instance. The update in mongodb-mcp-server is failing the OIDC tests so we're gonna have to fix and update it there first.

I am going to resolve this thread assuming that we are fine keeping the deps like this. If you have any differing opinion please feel free to re-open this.

package.json
himanshusinghs: I have brought back 23.0.0 to the minimum allowed list with the assumption that VSCode uses this info to restrict the availability of extension to the users. The rationable being that even though node v23 is EOL, if a user is working with VSCode using node version 23 we are still able to work with that env.

nirinchev (approved):

PR #1132, opened 2025-09-23
fix: disable MCP auto start and instead request on first load and new connection attempts VSCODE-704

himanshusinghs:
## Description

This PR flips the earlier behaviour of MongoDB MCP server auto-starting on VS Code activation. Instead of auto-starting by default, we now show a notification, only when user actually connects to a connection in VSCode, asking users to enable or disable auto-start.

Additionally we've enabled the `readOnly` mode by default.

### New Behavior
- **No auto-start by default**: MCP server only starts when explicitly enabled by the user.
- **Repeated notifications**: If the user doesn't select an option or the notification auto-dismisses, we will ask again on every new connection attempt.
- **User choice persistence**: Once a user makes a decision ('Auto start enabled'/'Start Once'/'Auto start disabled'), their preference is saved and respected
- **Readonly mode enabled by default**: The readonly mode is set to be enabled by default.

### Migration for Existing Users

We've migrated old config values to preserve explicit user preferences. In the previous logic, only "disabled" represented an intentional user action, so we preserve that choice while treating all other values as requiring user confirmation.

| Old `mdb.mcp.server` value | New `mdb.mcp.server` value | Reason |
| -------------------------- | -------------------------- | ------ |
| `"ask"`                    | `"prompt"`                    | Previous default - ask for permission |
| `"enabled"`                | `"prompt"`                    | Was auto-set by extension - ask for permission |
| `"disabled"`               | `"autoStartDisabled"`    | Explicit user choice - preserve preference |
| anything else              | the same as before            | Explicit user choice - preserve preference |

### Configuration Values

The new configuration accepts these values:
- `"prompt"` (default) - Show notification to request permission
- `"autoStartEnabled"` - Automatically start MCP server on VS Code launch, no notifications
- `"autoStartDisabled"` - Never auto-start, no notifications

#### When server is not already running
image

#### When server is already running
image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


src/mcp/mcpController.ts
```diff
@@ -299,9 +330,27 @@ ${jsonConfig}`,
...
+    logger.info('Active connection changed, should update MCP server', {
+      'mdb.mcp.server': autoStartConfig,
+      serverStarted: !!this.server,
+    });
+
+    if (
+      this.connectionController.getActiveConnectionId() &&
+      (autoStartConfig === 'Ask' ||
+        autoStartConfig === null ||
+        autoStartConfig === undefined)
```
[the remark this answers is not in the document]
himanshusinghs: we prefer strict equals

src/mcp/mcpController.ts
```diff
@@ -174,60 +199,70 @@ export class MCPController {
...
-      // user might not act on the notification in which case the final update
-      // will never happen.
-      await this.setMCPAutoStartConfig('enabled');
-      let selectedServerStartConfig: McpServerStartupConfig = 'enabled';
-
       const prompt = await vscode.window.showInformationMessage(
-        'MongoDB MCP server started automatically and will connect to your active connection. Would you like to keep or disable automatic startup?',
-        'Keep',
-        'Disable',
+        'MongoDB MCP server is available and when started, can automatically connect to your active connection. Would you like to automatically start MongoDB MCP server with VSCode?',
```
gagik:
```suggestion
        'Would you like to automatically start the MongoDB MCP server? The MongoDB MCP Server will automatically connect to your active MongoDB instance.',
```

gagik: or something like that

src/mcp/mcpController.ts
```diff
@@ -19,8 +19,14 @@ import type { MCPConnectParams } from './mcpConnectionManager';
...
 import { getMCPConfigFromVSCodeSettings } from './mcpConfig';
+import { DEFAULT_TELEMETRY_APP_NAME } from '../connectionController';
 
-export type McpServerStartupConfig = 'enabled' | 'disabled';
+export type McpServerStartupConfig =
+  | 'Ask'
+  | 'Auto Start Enabled'
+  | 'Auto Start Disabled';
+
+export type OldMcpServerStartupConfig = 'ask' | 'enabled' | 'disabled';
```
gagik:
```suggestion
export type DeprecatedMcpServerStartupConfig = 'ask' | 'enabled' | 'disabled';
```
nit, a comment describing something like `Introduced in vx.x.x but migrated to the new config in vx.x.x` would be useful as well

src/mcp/mcpController.ts
```diff
@@ -87,9 +97,21 @@ export class MCPController {
       },
     );
 
-    if (this.shouldStartMCPServer()) {
-      await this.startServer();
-      void this.notifyOnFirstStart();
+    switch (this.getMCPAutoStartConfig()) {
```
gagik: maybe worth moving this to its own function i.e. `promptOrStartMCPServer` to keep the activate function clearer

package.json
```diff
@@ -1331,11 +1331,11 @@
         "mdb.mcp.server": {
           "type": "string",
           "enum": [
-            "ask",
-            "enabled",
-            "disabled"
+            "Ask",
+            "Auto Start Enabled",
```
gagik, line 1340: is there a functional reason to changing these to more capitalized values, could `enumNames` be useful here?

himanshusinghs: The only reason to change these was to facilitate the new behaviour. There wasn't a need for them to be capital cased but I noticed that it looked better in the settings UI.

After you mentioned the `enumNames`, I checked the documentation again and there's something similar `enumLabels` which I can use. Will try and if it works out, I will keep that.

src/mcp/mcpController.ts
himanshusinghs: makes sense, will do that.

src/mcp/mcpController.ts
himanshusinghs: ~added one~ this has been removed as it was not required

src/mcp/mcpController.ts
himanshusinghs: Does this sound alright?
```
Would you like to automatically start the MongoDB MCP server? When started, the MongoDB MCP Server will automatically connect to your active MongoDB instance.
```

gagik: that sounds fine, I'm no copywriter though :D

package.json
gagik: yeah if we can minimize the migration of values by using labels it'd be great

himanshusinghs: Ahh unfortunately, I don't see a way around migrating the old values. So keeping the old values while just changing the labels won't really work.

src/mcp/mcpController.ts
himanshusinghs: addressed in the new re-write of the logic

src/mcp/mcpController.ts
```diff
@@ -299,43 +329,61 @@ ${jsonConfig}`,
...
 
-  private shouldStartMCPServer(): boolean {
-    return this.getMCPAutoStartConfig() !== 'disabled';
+  private shouldPromptForAutoStart(): boolean {
+    const storedConfig = this.getMCPAutoStartConfig();
+    return storedConfig === 'prompt';
   }
 
-  private getMCPAutoStartConfig(): McpServerStartupConfig | undefined {
+  private getMCPAutoStartConfig(): unknown {
```
himanshusinghs: Even though this could be typed to `MCPServerStartupConfig` but since we don't do any validation, its best to leave this typed to `unknown` because that is the reality.

src/mcp/mcpController.ts
```diff
@@ -64,10 +68,17 @@ export class MCPController {
   }: MCPControllerConfig) {
     this.context = context;
     this.connectionController = connectionController;
-    this.getTelemetryAnonymousId = getTelemetryAnonymousId;
+    this.vsCodeMCPLogger = new VSCodeMCPLogger(Keychain.root);
+    this.mcpConnectionManager = new MCPConnectionManager({
+      logger: this.vsCodeMCPLogger,
+      getTelemetryAnonymousId,
+    });
```
himanshusinghs: This is a little change I did as a drive by to make the `MCPConnectionManager` instantiation more predictable. Earlier we were instantiating only when a client would connect to MCP server, thereby invoking the `ConnectionManagerFactoryFn` which would then instantiate the `MCPConnectionManager`. Because we always use the same instance of `MCPConnectionManager` we can just build it right away and update the logger when provided in the `ConnectionManagerFactoryFn`.

src/mcp/mcpController.ts
```diff
@@ -130,16 +208,11 @@ export class MCPController {
         apiClientSecret: '<redacted>',
       });
 
-      const createConnectionManager: ConnectionManagerFactoryFn = async ({
+      const createConnectionManager: ConnectionManagerFactoryFn = ({
```
gagik: then maybe let's just get rid of this function altogether and just use setlogger? It's not creating a new connection manager.
Also, is `switchConnectionManagerToCurrentConnection` running here relevant?

src/mcp/mcpController.ts
gagik: It's the reality but this type acts as documentation of possible valid values. I think this is still preferable. If you want to make assertions first and return undefined for non-valid values then it might be good, but I think getting rid of the type altogether gets rid of documentation

src/mcp/mcpController.ts
```diff
@@ -299,43 +327,72 @@ ${jsonConfig}`,
...
+    config: MCPServerStartupConfig,
   ): Promise<void> {
     await vscode.workspace
-      .getConfiguration('mdb')
-      .update('mcp.server', config, true);
+      .getConfiguration()
+      .update('mdb.mcp.server', config, true);
+  }
+
+  public get _test_isServerRunning(): boolean {
```
gagik: We should avoid having test functions like this in our public-facing interface, I really dislike them. Ideally we'd only test APIs we consume but if you really want to use it the field in tests you can access the private fields with i.e. `(mcpController as any)['server']` and create a helper function which does this and asserts it.

src/mcp/mcpController.ts
nirinchev: I'm not sure I'm fully convinced with this change. It seems weird that we're eagerly creating the connection manager if the server is not started (or may never be started). Obviously, it's cheap and all, but it's a code smell that we're doing it. Additionally, we're making a lot of assumptions that - while correct today - are not necessarily codified in the API - namely:
1. `createConnectionManager` will only ever be called once. If for some reason in the future, the MCP server were capable of creating multiple connection managers, it wouldn't be clear that the VSCode extension is reusing its own and not really giving us a new instance.
2. That the `logger` is the only state of the mcp connection manager that needs to be updated when a new server is started. If we added more stuff in the future, we'd need to make sure to reset that state, making things generally clunky.

Was there some problem with the previous behavior that this is now solving?

src/mcp/mcpController.ts
```diff
@@ -149,6 +222,9 @@ export class MCPController {
           this.connectionController,
         ),
         additionalLoggers: [new VSCodeMCPLogger(Keychain.root)],
```
nirinchev, line 223: Not sure if it matters, but this is creating a new logger instead of reusing the one we create in the ctor - is it intentional?

src/mcp/mcpController.ts
```diff
@@ -87,16 +98,83 @@ export class MCPController {
...
+        // In the previous logic only 'disabled' value would've represented an
+        // explicit user action which is why we preserve that and map it to new
+        // disabled value.
+        case 'disabled': {
+          await this.setMCPAutoStartConfig('autoStartDisabled');
+          break;
+        }
+
+        // Any other value is possible only if user explicitly mentioned or when
+        // if the values were already migrated it so we don't alter them.
```
nirinchev: This sentence reads awkwardly - consider rewording it.

src/mcp/mcpController.ts
```diff
@@ -299,43 +327,72 @@ ${jsonConfig}`,
...
+
+      const connectParams: MCPConnectParams | undefined =
+        connectionId && mongoClientOptions
+          ? {
+              connectionId: connectionId,
+              connectionString: mongoClientOptions.url,
+              connectOptions: mongoClientOptions.options,
+            }
+          : undefined;
+      await this.mcpConnectionManager.updateConnection(connectParams);
```
nirinchev: This is another issue with the new design - we'd be connecting and disconnecting to mongodb regardless of whether the MCP server is running.

src/mcp/mcpController.ts
```diff
@@ -87,16 +98,83 @@ export class MCPController {
...
+      }
+    } catch (error) {
+      logger.error('Error when migrating old config to the new config', error);
+    }
+  }
+
+  private async promptForMCPAutoStart(): Promise<void> {
+    try {
+      const promptResponse = await vscode.window.showInformationMessage(
+        'Would you like to automatically start the MongoDB MCP server? When started, the MongoDB MCP Server will automatically connect to your active MongoDB instance.',
```
nirinchev: Should we reword this to be more explicit about the benefits here? Might be a good idea to work with Gaurab/Ighodalo on the message.

nirinchev: This should also be very clear that if they respond "yes", it will start for all new connections as well. Finally, we should consider if we should offer a 3rd option, which is start it for the current connection only and prompt for new ones.

src/mcp/mcpController.ts
```diff
@@ -87,16 +98,83 @@ export class MCPController {
...
+        'Yes',
+        'Not now',
```
nirinchev: "Not now" sounds like a temporary choice, but in reality, it disables the prompting - we should probably be more explicit about it.

src/mcp/mcpController.ts
```diff
@@ -87,16 +98,83 @@ export class MCPController {
...
   public async startServer(): Promise<void> {
     try {
-      // Stop an already running server if any
-      await this.stopServer();
+      if (this.server) {
+        logger.info(
+          'MCP server start requested. An MCP server is already running, will not start a new server.',
+        );
+        return;
+      }
```
nirinchev, line 199: This is not necessarily an issue caused by the current PR, but we should consider returning `boolean` from this function to indicate whether the server was actually started or not. Then in the extension controller, we can show a notification when the user tries to start the server.

src/test/suite/mcp/mcpController.test.ts
```diff
@@ -1,27 +1,63 @@
+/* eslint-disable no-loop-func */
```
nirinchev: Do we need this disabled for the entire file?

src/mcp/mcpController.ts
```diff
@@ -299,43 +327,72 @@ ${jsonConfig}`,
...
-      .update('mcp.server', config, true);
+      .getConfiguration()
+      .update('mdb.mcp.server', config, true);
+  }
+
+  public get _test_isServerRunning(): boolean {
+    return (
+      this.server !== undefined &&
+      this.server.runner instanceof StreamableHttpRunner
+    );
```
nirinchev: This can probably be exposed as a normal property, not a test one. Then we can use it in the extension controller to decide what to do when the user calls the start/stop server commands.

Additionally, why are we checking if the runner is an instance of `StreamableHttpRunner` - what else could it be?

src/mcp/mcpController.ts
```diff
@@ -299,43 +327,72 @@ ${jsonConfig}`,
...
+  }
+
+  public get _test_mcpConnectionManager(): MCPConnectionManager {
+    return this.mcpConnectionManager;
```
nirinchev: This should probably go away - we shouldn't be switching connections if the server is not running, so we'd need to rework the tests that are using it. If we absolutely need it for testing, it's typically better to use the private field directly - `controller["mcpConnectionManager"]`.

src/mcp/mcpController.ts
himanshusinghs: Given that the type would only serve as the documentation, I would prefer to write a jsdoc instead of typing the returned value which gives the rest of the code a false implication that the value is guaranteed to be `MCPServerStartupConfig`

src/mcp/mcpController.ts
himanshusinghs: The change does not assume that `createConnectionManager` will only be called once. It has always been called as many times as a client connects. The reality was that we were always returning the same ConnectionManager instance, regardless of how many times the client connected and that was because VSCode only establishes one connection at any given moment.

This change, for the current state, untangles the ConnectionManager creation. ~There was no functional necessity only a tiny bit more readability for how the flow works.~

There was indeed a problem with the original implementation and the problem being.
1. VSCode establishes connections to the MCP server
2. User fetches the MCP config, gets the URL and auth token
3. User uses the retrieved config to configure other tool to connect to the same MCP server, lets say the tool is cursor.

At this point, as soon as the connection is established by the Cursor client, the other client's connection manager will become orphan and won't get updated. It's because we're currently returning a new connection manager instance on each client connections according to [these lines](https://github.com/mongodb-js/vscode/blob/d0a1a189093e4b62a2761bc4721ceaac144f7c82/src/mcp/mcpController.ts#L136-L140), which now seems wrong.

src/mcp/mcpController.ts
himanshusinghs: `ConnectionManagerFactoryFn` is an API requirement from TransportRunner so we can't get rid of the function. From the MCP server perspective, the factory still provides a ConnectionManager instance and whether it returns a pre-created one or generates a new one shouldn't matter here.

src/mcp/mcpController.ts
himanshusinghs: Yea the previous tests were referring the private variables but I didn't like that either because the tests were getting brittle relying more on the internal implementation than the actual outcome.

src/mcp/mcpController.ts
himanshusinghs: yea this is a mistake, this should use the existing logger

src/mcp/mcpController.ts
himanshusinghs: which is also what the PlaygroundController and LanguageServer does currently. I don't see a problem here.

src/mcp/mcpController.ts
himanshusinghs: Got it. I will send a Slack message for this.

src/mcp/mcpController.ts
himanshusinghs: Will cover this as part of the refactor for the message itself.

src/mcp/mcpController.ts
himanshusinghs: Yea this makes sense. I will add this.

src/test/suite/mcp/mcpController.test.ts
himanshusinghs: Will move it to individual places.

src/mcp/mcpController.ts
gagik: > relying more on the internal implementation than the actual outcome

But this is doing the same, no? Usually the fact that we're trying to access private fields is a sign that we're relying on internal implementation.

himanshusinghs: Yea but its behind a Public API for tests, so tests won't have to change with change in internal implementation.

gagik: I don't think we should have a public API for tests, private field access is preferrable imo; 
you can achieve the same effect by creating helper functions in test utils which do the private field access in a more unified way if you want, test code shouldn't live in production code IMO. 
If we're relying on internal implementation (which should be rare) and tests break and one has to go check them individually, this is a good thing. It helps one be more aware of the changes' implications

himanshusinghs: My intention was to not rely on internals for tests as much as possible. Doing (connectionManager as any).somePrivateField, in tests directly or through test helpers, makes the tests brittle which is what I was trying to avoid.

If for every internal implementation changes that doesn't change the outcome, tests or test helpers have to be updated because they were relying on the those internal implementations then that's generally not a good sign.

I do realise though that this exposing of `_test_method` approach feels a but unconventional. The motivation was the use of similar pattern in compass where we allow possibility of providing test props for components.

himanshusinghs: I figured an alternative way to test without relying any way on internals so I have removed the public test methods in this commit - [1b14b66](https://github.com/mongodb-js/vscode/pull/1132/commits/1b14b660fd4e9f6c15143b7f344faa0bdbc196f1). Let me know how you feel about this?

src/mcp/mcpController.ts
himanshusinghs: addressed

src/mcp/mcpController.ts
himanshusinghs: addressed

src/test/suite/mcp/mcpController.test.ts
himanshusinghs: addressed

src/mcp/mcpController.ts
himanshusinghs: This is gone now with the changes in [1b14b66](https://github.com/mongodb-js/vscode/pull/1132/commits/1b14b660fd4e9f6c15143b7f344faa0bdbc196f1)

src/mcp/mcpController.ts
himanshusinghs: This is gone now with the changes in [1b14b66](https://github.com/mongodb-js/vscode/pull/1132/commits/1b14b660fd4e9f6c15143b7f344faa0bdbc196f1).

But about the concern you noted - "we shouldn't be switching connections if the server is not running", lets address that in https://github.com/mongodb-js/vscode/pull/1132#discussion_r2378305089

src/mcp/mcpController.ts
himanshusinghs: To keep the related conversation in one thread I am moving this concern to https://github.com/mongodb-js/vscode/pull/1132#discussion_r2378305089

src/mcp/mcpController.ts
himanshusinghs: Related issue pointed by @nirinchev:

> This is another issue with the new design - we'd be connecting and disconnecting to mongodb regardless of whether the MCP server is running.

Which I think is not a problem given that we do the same already for PlaygroundController / LanguageServer.

src/mcp/mcpController.ts
himanshusinghs: I realised that I already treat the fetched config value as unknown during the migration and the rest of the usage is explicit on the known new values so I have brought back the type - `McpServerStartupConfig`

src/mcp/mcpController.ts
himanshusinghs: I will be addressing this separately as there might be notifications that we'd like to show when server is already running.

src/mcp/mcpController.ts
himanshusinghs: The message has been updated as per recommendation in [a0cc53e](https://github.com/mongodb-js/vscode/pull/1132/commits/a0cc53ebc161eb474d2d68eac0db8c46f5d79aee)

src/mcp/mcpController.ts
```diff
@@ -52,10 +57,10 @@ export class MCPController {
   private context: vscode.ExtensionContext;
   private connectionController: ConnectionController;
   private getTelemetryAnonymousId: () => string;
+  private clientConnectionManagers: MCPConnectionManager[] = [];
```
nirinchev: [extreme nit] Bit confusing this is called "client" connection managers considering they're the connection managers for the mcp _server_. I think `mcpConnectionManagers` would make more sense, but not critical at all.

src/mcp/mcpController.ts
```diff
@@ -87,41 +94,109 @@ export class MCPController {
       },
     );
 
-    if (this.shouldStartMCPServer()) {
+    void this.promptForMCPAutoStart();
```
nirinchev: This is happening when we activate the extension - does it really make sense to prompt at this point? Mostly worried that this is happening when vscode starts and it's likely there's a bunch of notifications being presented to the user - I'm worried we'd be contributing to a crappy UX where the developer just wants to clear all notifications and focus on their task at hand.

src/mcp/mcpController.ts
```diff
@@ -87,41 +94,109 @@ export class MCPController {
...
+        }
+      }
+    } catch (error) {
+      logger.error('Error when migrating old config to the new config', error);
+    }
+  }
+
+  private async promptForMCPAutoStart(): Promise<void> {
+    try {
+      const storedConfig = this.getStoredMCPAutoStartConfig();
```
nirinchev: Why do we care about the stored config value here? I imagine if we're here, we've already migrated the config, so that should always be the same as the coerced value.

src/mcp/mcpController.ts
```diff
@@ -302,43 +369,99 @@ ${jsonConfig}`,
...
   }
 
-  private shouldStartMCPServer(): boolean {
-    return this.getMCPAutoStartConfig() !== 'disabled';
+  /**
+   * Retrieves the stored value for the key "mdb.mcp.server" in the user config
+   * and coerces it to one of the valid identified values typed by
+   * `MCPServerStartupConfig`.
+   */
+  private getCoercedMCPAutoStartConfig(): MCPServerStartupConfig {
```
nirinchev: This set of functions don't make a lot of sense to me. The only place where we should know that the config might be storing incorrect values is in the `migrateOldConfigToNewConfig` function - everywhere else, we should assume we're working with valid configs, so having these functions introduces unnecessary complexity.

src/mcp/mcpController.ts
```diff
@@ -87,41 +94,109 @@ export class MCPController {
...
+          { storedConfig, coercedConfig, serverRunning: !!this.server },
+        );
+        return;
+      }
+
+      const notificationActions = this.server
+        ? (['Auto-Start', 'Never'] as const)
+        : (['Auto-Start', 'Start Once', 'Never'] as const);
+
+      const promptResponse = await vscode.window.showInformationMessage(
```
nirinchev, line 162: Do we want to log the response? We're logging when we're not prompting, but it seems to be equally important to log the response whenever we do prompt.

src/mcp/mcpController.ts
```diff
@@ -302,43 +369,99 @@ ${jsonConfig}`,
...
+    );
+
+    if (this.connectionController.getActiveConnectionId()) {
+      void this.promptForMCPAutoStart();
+    }
+
+    await this.switchAllConnectionManagerToCurrentConnection();
+  }
+
+  private async switchAllConnectionManagerToCurrentConnection(): Promise<void> {
```
nirinchev: This is only used once immediately above and is a 1-liner that hardly justifies being extracted to a separate function.

src/mcp/mcpController.ts
himanshusinghs: Makes sense, adjusted

src/mcp/mcpController.ts
himanshusinghs: It was decided that we would want to keep the current behaviour and ensure that it supports multiple client handling.

src/mcp/mcpController.ts
himanshusinghs: addressed [5f77f59](https://github.com/mongodb-js/vscode/pull/1132/commits/5f77f590f1b7f0544daba5b626e011c363731301)

src/mcp/mcpController.ts
himanshusinghs: addressed [5f77f59](https://github.com/mongodb-js/vscode/pull/1132/commits/5f77f590f1b7f0544daba5b626e011c363731301)

src/mcp/mcpController.ts
himanshusinghs: addressed as discussed in [1613071](https://github.com/mongodb-js/vscode/pull/1132/commits/1613071acdc3fc4848e39c02fbad5bf65a096682)

src/mcp/mcpController.ts
himanshusinghs: this has been removed in [1613071](https://github.com/mongodb-js/vscode/pull/1132/commits/1613071acdc3fc4848e39c02fbad5bf65a096682)

src/mcp/mcpController.ts
himanshusinghs: addressed this as discussed in commit [8f8606f](https://github.com/mongodb-js/vscode/pull/1132/commits/8f8606fa3f99f535406859f2d982dff9b458dfd9)

Popup won't be shown on extension activate, only when user connects to a connection in VSCode and the right requirements are met.

src/mcp/mcpController.ts
```diff
@@ -87,41 +99,114 @@ export class MCPController {
...
+    } catch (error) {
+      logger.error('Error when migrating old config to the new config', error);
+    }
+  }
+
+  private async promptForMCPAutoStart(): Promise<void> {
+    try {
+      const autoStartConfig = this.getMCPAutoStartConfig();
+      if (autoStartConfig !== 'prompt') {
+        logger.info(
```
nirinchev: [nit] This is probably a debug-level notification.

src/mcp/mcpController.ts
```diff
@@ -87,41 +99,114 @@ export class MCPController {
...
+
+      logger.info(
+        'Prompt to configure MCP auto start requested. Will prompt.',
+        {
+          autoStartConfig,
+          serverRunning: !!this.server,
+        },
+      );
+
+      const notificationActions = this.server
```
nirinchev, line 158: Might be a good idea to add a comment why we're doing that here.

src/mcp/mcpController.ts
```diff
@@ -302,43 +375,63 @@ ${jsonConfig}`,
...
 
-  private getMCPAutoStartConfig(): McpServerStartupConfig | undefined {
+  private getMCPAutoStartConfig<ConfigValue = MCPServerStartupConfig>():
+    | ConfigValue
+    | undefined {
     return vscode.workspace
-      .getConfiguration('mdb')
-      .get<McpServerStartupConfig>('mcp.server');
+      .getConfiguration()
+      .get<ConfigValue>('mdb.mcp.server');
```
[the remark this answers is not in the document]
himanshusinghs: We've been back and forth on this and have decided to keep this pattern.

src/test/suite/mcp/mcpController.test.ts
```diff
@@ -3,25 +3,73 @@ import sinon from 'sinon';
...
 import { TelemetryService } from '../../../telemetry';
 import { TEST_DATABASE_URI } from '../dbTestHelper';
+import { waitFor } from '../waitFor';
+
+function timeout(ms: number): Promise<void> {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+}
+
```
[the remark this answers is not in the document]
himanshusinghs: It does not exist anywhere else. When there will be a requirement, we will pull it out.

src/test/suite/mcp/mcpController.test.ts
```diff
@@ -38,247 +86,756 @@ suite('MCPController test suite', function () {
...
 
-  suite('when an MCP server is already running', function () {
-    test('it should notify the connection manager of the connection changed event', async function () {
-      // We want to connect as soon as connection changes
-      await vscode.workspace
-        .getConfiguration('mdb')
-        .update('mcp.server', 'enabled');
+    suite('when MCP server shuts down', function () {
+      test('should terminate individual client connections and clear up the internal connection manager state', async function () {
+        this.timeout(60_000);
```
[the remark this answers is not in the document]
himanshusinghs: Removed this.

src/mcp/mcpController.ts
```diff
@@ -167,67 +236,71 @@ export class MCPController {
...
-  private async notifyOnFirstStart(): Promise<void> {
-    try {
-      if (!this.server) {
-        // Server was never started so no need to notify
-        return;
-      }
-
-      const serverStartConfig = this.getMCPAutoStartConfig();
+  private async createConnectionManager(
+    this: MCPController,
```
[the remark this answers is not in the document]
himanshusinghs: The syntax ensures that `this` in the method is bounded to the right instance.

nirinchev: I agree - this is unconventional. Why are we doing it?

nirinchev (approved):
Looks good - final nits from me.

src/mcp/mcpController.ts
himanshusinghs: I don't think its unconventional, rather more common. Regardless of how `createConnectionManager` is called (with arrow or with explicit binding), this helps to specify clearly if the function can or cannot work without a correct `this` reference.

src/mcp/mcpController.ts
himanshusinghs: addressed

src/mcp/mcpController.ts
himanshusinghs: addressed

src/mcp/mcpController.ts
nirinchev: This is an instance function though - there's like 15 other instance functions in this class and none of them have the `this` argument explicitly defined.

himanshusinghs: Yea the difference is that this function will certainly be passed as a reference to be invoked elsewhere. So whoever does the passing, it would be nice for that site to know that the function needs correct `this`.

himanshusinghs: moved the factory to a static func [4f39d31](https://github.com/mongodb-js/vscode/pull/1132/commits/4f39d31cc179e584d933bd05e2ee77f49ae5a710)

PR #1133, opened 2025-09-24
chore: reorganise GA to check and build once and run tests with matrix strategy VSCODE-703

himanshusinghs:
## Description

Earlier we were running check, build and tests on each OS in our matrix which did not allow us to test whether a build on ubuntu would work fine on Windows / MacOS or not.

With this commit we are re-organising to check and build once on ubuntu and use the final artifact later in the test job to run tests against.

Accordingly the test-and-build-from-fork and draft-release workflows have been modified to adapt the same change.

This PR simply re-organise the existing actions and does not introduce any install tests just yet. That will come in a different PR.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/.github/workflows/draft-release.yaml
+++ b/.github/workflows/draft-release.yaml
@@ -1,27 +1,26 @@
+# Run manually to prepare a draft release for the next version of the extension.
+# The workflow will create a draft github release where the .vsix can be
+# downloaded and manually tested before publishing. To release the version,
+# publish the draft release, which will trigger the publish-release workflow.
 name: Draft release
 
 on:
   workflow_dispatch:
     inputs:
       versionBump:
-        description: 'Version bump'
+        description: "Version bump"
         type: choice
         required: true
-        default: 'patch'
+        default: "patch"
         options:
-        - patch
-        - minor
-        - major
-        - exact-version
+          - patch
+          - minor
+          - major
+          - exact-version
 
       exactVersion:
         description: 'Exact version: (Only effective selecting "exact-version" as version bump)'
         required: false
-
-description: |
-  Run manually to prepare a draft release for the next version of the extension. The workflow will create a draft
-  github release where the .vsix can be downloaded and manually tested before publishing. To release the version,
-  publish the draft release, which will trigger the publish-release workflow.
 
 permissions:
   contents: write
@@ -29,6 +28,8 @@ permissions:
 jobs:
   prepare-release:
     runs-on: ubuntu-latest
+    outputs:
+      release-tag: ${{ steps.set-tag.outputs.release-tag }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -85,8 +86,40 @@ jobs:
             exit 1
           fi
 
-      - name: Run tests and build
-        uses: ./.github/workflows/actions/test-and-build
+      - name: Set release tag output
+        id: set-tag
+        run: echo "release-tag=${RELEASE_TAG}" >> $GITHUB_OUTPUT
+
+      - name: Upload updated package.json
+        uses: actions/upload-artifact@v4
+        with:
+          name: updated-package-json
+          path: package.json
+
+  build-and-package:
+    name: Check, Build and Package
+    runs-on: ubuntu-latest
+    needs: prepare-release
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Download updated package.json
+        uses: actions/download-artifact@v4
+        with:
+          name: updated-package-json
+
+      - name: Setup Node.js Environment
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22.15.1
+          cache: npm
+
+      - name: Check, build and package
+        uses: ./.github/workflows/actions/build-and-package
         with:
           SEGMENT_KEY: ${{ secrets.SEGMENT_KEY_PROD }}
           ARTIFACTORY_HOST: ${{ secrets.ARTIFACTORY_HOST }}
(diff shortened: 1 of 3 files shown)


.github/workflows/test-and-build.yaml
```diff
@@ -48,4 +78,22 @@ jobs:
...
+          if [ "${{ needs.build-and-package.result }}" == "failure" ] || [ "${{ needs.test.result }}" == "failure" ]; then
+            echo "One or more jobs failed"
+            exit 1
+          elif [ "${{ needs.build-and-package.result }}" == "cancelled" ] || [ "${{ needs.test.result }}" == "cancelled" ]; then
+            echo "One or more jobs were cancelled"
+            exit 1
+          else
+            echo "All jobs completed successfully or were skipped"
+            exit 0
+          fi
```
himanshusinghs: Curious about this, why was this needed?

himanshusinghs (commented):
I am unable to approve (as author) but looks good to me 👍. I added a small question for clarification.

.github/workflows/test-and-build-from-fork.yaml
```diff
@@ -28,17 +60,39 @@ jobs:
...
-          MDB_IS_TEST: "true"
-        run: npm run test
+          PR_URL: ${{github.event.pull_request.html_url}}
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
+  status-check:
+    name: Test and Build
+    runs-on: ubuntu-latest
+    needs: [build-and-package, test]
+    if: always()
```
gagik: why have this if?

.github/workflows/test-and-build.yaml
gagik: yeah this seems unnecessary

tculig: It was hanging without this. Perhaps there's a more elegant solution, but its not just AI slop. 😅

.github/workflows/test-and-build-from-fork.yaml
tculig: it was hanging, when running a normal build, this would start but never resolve, so the checks never passed.

.github/workflows/test-and-build.yaml
```diff
@@ -37,8 +67,8 @@ jobs:
           node-version: 22.15.1
           cache: npm
```
gagik: let's move this inside `./.github/workflows/actions/run-tests`

.github/workflows/draft-release.yaml
```diff
@@ -85,8 +86,40 @@ jobs:
             exit 1
           fi
 
-      - name: Run tests and build
-        uses: ./.github/workflows/actions/test-and-build
+      - name: Set release tag output
+        id: set-tag
+        run: echo "release-tag=${RELEASE_TAG}" >> $GITHUB_OUTPUT
+
+      - name: Upload updated package.json
```
gagik: I'm confused by this; what is this for? Why would a draft release upload a package.json?
Is it so the version number is different? Then package-lock.json is also interesting to us, arguably more

.github/workflows/test-and-build.yaml
tculig: Updated!

.github/workflows/draft-release.yaml
tculig: The point is to build once and then use the built package for all subsequent tasks, namely running tests.
You are right about the package-lock file, I'll add it to the artifact..

gagik: can we move `build-and-package` into this instead? this job doesn't do much on its own otherwise.

gagik: also I don't see how this would do much; node_modules would also be what we're looking for if we're trying to skip the install step

.github/workflows/actions/build-and-package/action.yaml
```diff
@@ -0,0 +1,152 @@
...
+          github.event_name == 'schedule'
+        )
+      shell: bash
+      env:
+        JIRA_API_TOKEN: ${{ inputs.JIRA_API_TOKEN }}
+        JIRA_BASE_URL: "https://jira.mongodb.org"
+        JIRA_PROJECT: "VSCODE"
+        JIRA_VULNERABILITY_BUILD_INFO: "- [GitHub Run|https://github.com/mongodb-js/vscode/actions/runs/${{github.run_id}}/jobs/${{github.job}}]"
+      run: |
+        pnpm run create-vulnerability-tickets > /dev/null
```
gagik, line 142: this is broken right now, I found out the hard way just now https://github.com/mongodb-js/vscode/pull/1196

gagik (changes requested):
Okay so currently, this isn't really doing anything in practice. This is largely because (and I'm not sure why we never added them) our E2E tests aren't running in CI. Running `pnpm run compile` is actually not necessary when running any of our current tests which is why our tests still pass even though no build artifacts are being shared between processes.

The problem this is trying to solve makes sense only for E2E tests where we're actually running them against our built artifiacts. For all other tests, we're import the source directly so the idea of building on one host and running on the other doesn't do anything.

`package-lock.json` and `package.json` shouldn't change from install to install so uploading them in one step and downloading in another doesn't do much and shouldn't do much.

So we'd need to:
1. Update vitest config to include e2e tests.
2. For the Ubuntu tasks, also run `compile` and upload the entire `dist` file
3. Add extra steps in our existing test workflows to download the build artifiact and run it. We should consider skipping E2E tests for fork PRs but not too sure.

gagik:
Continued in https://github.com/mongodb-js/vscode/pull/1235

PR #1168, opened 2025-10-21
chore: update mongosh

gagik:
Updates the mongosh service provider. Among other things, this allows us to no longer be broken with mongosh VSCode tests


gagik:
This is done already with an older bump PR.

PR #1169, opened 2025-10-29
chore: update telemetry test not to fail without key VSCODE-517

tculig:
## Description

https://jira.mongodb.org/browse/VSCODE-517

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [X] Misc

## Types of changes

- [ ] *Backport Needed*
- [X] Patch (non-breaking change which fixes an issue)


src/test/suite/telemetry/telemetryService.test.ts
```diff
@@ -52,6 +52,32 @@ suite('Telemetry Controller Test Suite', () => {
...
+    const originalRequire = Module.prototype.require;
+    Module.prototype.require = function (id: string, ...args: any[]): any {
+      if (id === '../../../../constants') {
+        return { segmentKey: process.env.SEGMENT_KEY };
+      }
+      return originalRequire.apply(this, [id, ...args]);
+    };
+
+    // Store the original require for restoration
+    originalRequireFunction = originalRequire;
```
nirinchev: This seems super heavy-handed - the `require(segmentKeyFileLocation)` call is used as a shorthand to reading a json file. Rather than mocking it, we could do something like:

```ts
const readSegmentKeyFromDisk = (): string => {
  if (process.env.CI) {
    return require('../../../../constants').segmentKey;
  }

  // "mock" the on-disk segment key
  return process.env.SEGMENT_KEY;
}
```

src/test/suite/telemetry/telemetryService.test.ts
```diff
@@ -111,33 +137,41 @@ suite('Telemetry Controller Test Suite', () => {
       '_isTelemetryFeatureEnabled',
       sandbox.fake.returns(true),
     );
+
+    // Mock readSegmentKey to return a fake key for local testing
+    sandbox.replace(
```
nirinchev: This should only be done for non-CI runs, right?

package.json
```diff
@@ -50,7 +50,7 @@
     "watch": "npm run compile && npm-run-all -p watch:*",
     "watch:extension": "npm run compile:extension -- -watch",
     "watch:extension-bundles": "webpack --mode development --watch",
-    "pretest": "npm run compile",
+    "pretest": "cross-env SEGMENT_KEY=test-segment-key npm run compile",
```
gagik: do we need to set this in both places?

tculig: the env variable gets baked into the config file at compile time, and there is a unit tests which checks the env values against the config file value, so the answer is yes, we do need it in both places. Alternatively, we can just skip the test that makes that check.

nirinchev (approved):

PR #1175, opened 2025-11-04
chore: use erasable syntax

gagik:
With https://github.com/mongodb-js/mongosh/pull/2601 and https://github.com/mongodb-js/compass/pull/7613, this should leave devtools-shared and MCP as last bastions of non-erasable syntax in our products 😄 

I admit this went out of hand, originally I thought erasable syntax will be just a couple of enums being converted to consts.
Then I noticed we use a lot of enums.
Then I noticed half of scripts use CJS-specific things.
Then I realized configuring mocha properly to use this erasable syntax is too complicated... so we still use `ts-node` for that, a shame 😢 
Still, having less ts-node is cool. 🤷


syntaxes/mongodbInjection.tmLanguage.json
```diff
@@ -1161,6 +1161,33 @@
             }
           }
         },
+        {
+          "name": "meta.object.member.mongodb",
```
gagik, line 1165: seems like some auto-generated that was missed from previous PRs

scripts/generate-icon-font.ts
```diff
@@ -61,14 +61,14 @@ async function main(): Promise<void> {
   const prettierConfig = await prettier.resolveConfig('./.prettierrc.json');
   await fs.writeFile(
     './package.json',
-    prettier.format(JSON.stringify(currentConfiguration), {
+    await prettier.format(JSON.stringify(currentConfiguration), {
```
nbbeeken, line 64: 😅 so did the file contain "`Promise { <pending> }`" before this fix?

src/documentSource.ts
```diff
@@ -1,8 +1,11 @@
...
-}
+export const DOCUMENT_SOURCE = {
+  DOCUMENT_SOURCE_TREEVIEW: 'treeview',
+  DOCUMENT_SOURCE_PLAYGROUND: 'playground',
+  DOCUMENT_SOURCE_COLLECTIONVIEW: 'collectionview',
+  DOCUMENT_SOURCE_CODELENS: 'codelens',
+} as const;
+
+export type DocumentSource =
+  (typeof DOCUMENT_SOURCE)[keyof typeof DOCUMENT_SOURCE];
```
nbbeeken:
```suggestion
export const DocumentSource = {
  DOCUMENT_SOURCE_TREEVIEW: 'treeview',
  DOCUMENT_SOURCE_PLAYGROUND: 'playground',
  DOCUMENT_SOURCE_COLLECTIONVIEW: 'collectionview',
  DOCUMENT_SOURCE_CODELENS: 'codelens',
} as const;

export type DocumentSource =
  (typeof DocumentSource)[keyof typeof DocumentSource];
```

Then we don't have as many code changes where this is used? it's also kinda convenient if we stick with """enums""" and their type being named the same thing

src/explorer/documentListTreeItem.ts
```diff
@@ -17,11 +17,14 @@ const log = createLogger('documents tree item');
 export const MAX_DOCUMENTS_VISIBLE = 10;
 
 export const DOCUMENT_LIST_ITEM = 'documentListTreeItem';
-export enum CollectionTypes {
-  collection = 'collection',
-  view = 'view',
-  timeseries = 'timeseries',
-}
+export const COLLECTION_TYPES = {
```
nbbeeken: same idea, keep the name and type the same. I won't comment on the rest we can just discuss on my `DocumentSource` comment

src/documentSource.ts
gagik: oh amazing, didn't realize this is possible

scripts/generate-icon-font.ts
gagik: I... am honestly unsure, it's possible when this script was written that version of prettier was a synchronous function and it never got re-run after updating 🤷

src/documentSource.ts
```diff
@@ -1,8 +1,11 @@
-export enum DocumentSource {
-  DOCUMENT_SOURCE_TREEVIEW = 'treeview',
-  DOCUMENT_SOURCE_PLAYGROUND = 'playground',
-  DOCUMENT_SOURCE_COLLECTIONVIEW = 'collectionview',
-  DOCUMENT_SOURCE_CODELENS = 'codelens',
-}
+export const DocumentSource = {
+  DocumentSource_TREEVIEW: 'treeview',
```
nbbeeken: > enum value variant in UPPER_SNAKE_CASE

Was this a search and replace mistake? cus wouldn't this be DOCUMENT_SOURCE_TREEVIEW?

src/documentSource.ts
```diff
@@ -1,8 +1,11 @@
...
+  DocumentSource_PLAYGROUND: 'playground',
+  DocumentSource_COLLECTIONVIEW: 'collectionview',
+  DocumentSource_CODELENS: 'codelens',
```
nbbeeken:
```suggestion
  codelens: 'codelens',
```

If you are workshopping patterns it is also kind nice when the key and the value are identical. of course not always possible and I lean more towards less code changes than more when we just want to migrate off the code gen stuff. 

your upper snake case pattern is just as good, only sharing opinions here.

nbbeeken (approved):
I think maybe a small mistake was made, but approach lgtm, so I'll leave the check ✅

src/documentSource.ts
gagik: 🙈 too much regex for a friday evening

src/documentSource.ts
gagik: cleaned up the stutter here; I'd probably also have preferred the casing to not be upper snake case but I'll keep it the way it (largely) was to avoid doing too many mass string replacements

gagik: actually managed to do this with a script and that was simple-ish... so we should have consistent naming now

PR #1181, opened 2025-11-10
chore: Add running environment to telemetry VSCODE-352

tculig:
## Description
vscode.env.appName is defined as "The application name of the editor, like 'VS Code'."
The expectation is that this will be "Azure Data Studio" when running inside Azure Data Studio.

Issue while testing: Latest version of Azure Data Studio uses vscode 1.82.0, which is now over 2 years old.
It doesn't support  Chat participant features and some others which are being heavily used in our extension.
Making a build which will work with Azure Data Studio has thusly proved to be quite a blocker.
Also, notably, Microsoft is retiring Azure Data Studio on February 28th, 2026: https://azure.microsoft.com/en-us/products/data-studio

Example app_names:

image

image

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [X] Misc

## Types of changes

- [ ] *Backport Needed*
- [X] Patch (non-breaking change which fixes an issue)

### DIFF
--- a/src/telemetry/telemetryService.ts
+++ b/src/telemetry/telemetryService.ts
@@ -65,7 +65,7 @@ export class TelemetryService {
     return {
       extension_version: `${version}`,
       device_id: this.deviceId,
-      app_name: vscode.env.appName, // e.g., "Visual Studio Code" or "Azure Data Studio"
+      app_name: vscode.env.appName || 'Visual Studio Code - Unknown', // e.g., "VS Code" or "Azure Data Studio"
     };
   }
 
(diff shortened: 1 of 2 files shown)


src/telemetry/telemetryService.ts
```diff
@@ -62,7 +62,11 @@ export class TelemetryService {
   }
 
   public getCommonProperties(): Record<string, string | undefined> {
-    return { extension_version: `${version}`, device_id: this.deviceId };
+    return {
+      extension_version: `${version}`,
+      device_id: this.deviceId,
+      app_name: vscode.env.appName, // e.g., "Visual Studio Code" or "Azure Data Studio"
```
gagik: could this be undefined? we should fallback to `Visual Studio Code - Unknown` or something like that

tculig: As far as I can tell, this should never be undefiend, but even if it is, is that an issue for the telemetry consumers?

gagik: yeah, just makes it more intentional / easier for us to track what scenario happened

PR #1185, opened 2025-11-13
chore: Prompt user to connect when running a playground VSCODE-686

tculig:
## Description

Before, if the user started a playground, and no connection is currently active, it would show a "toast" message in the bottom right corner:
image

The new logic is to instead offer them which collection to connect to:
image

However, I found this easy to miss, since if the user has no connections, its only a single line dropdown with muted colors that popup up.
So I added this interim step, so that the user gets very clear indication as to what is happening.
For experienced users, this should not be too much of a hassle, as they will probably make sure they are connected to a collection before running a playground.
image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/telemetry/telemetryService.ts
+++ b/src/telemetry/telemetryService.ts
@@ -62,7 +62,11 @@ export class TelemetryService {
   }
 
   public getCommonProperties(): Record<string, string | undefined> {
-    return { extension_version: `${version}`, device_id: this.deviceId };
+    return {
+      extension_version: `${version}`,
+      device_id: this.deviceId,
+      app_name: vscode.env.appName || 'Visual Studio Code - Unknown', // e.g., "VS Code" or "Azure Data Studio"
+    };
   }
 
   private async readSegmentKey(): Promise<string | undefined> {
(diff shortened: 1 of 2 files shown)


src/editors/playgroundController.ts
```diff
@@ -537,6 +532,26 @@ export default class PlaygroundController {
     return true;
   }
 
+  async _ensureConnection(): Promise<boolean> {
+    if (this._connectionController.isCurrentlyConnected()) {
+      return true;
+    }
+    const action = await vscode.window.showInformationMessage(
```
gagik, line 539: This looks good but just UX-wise, maybe we should just get them to do it right away?

tculig: What do you mean exactly by "right away"?

gagik: I was thinking maybe there's no need for extra confirmation. but thinking about it maybe this way they can end up accidentally connecting to a prod database and do something stupid, so yeah the way it is already is actually fine.

tculig: I added it mostly because its easy to miss a small, non-contrast, drawer opening somewhere else than where you clicked.

gagik: sgtm

PR #1186, opened 2025-11-16
test: VSCode test setup and switch to insiders VSCODE-700

tculig:
## Description

The problem was: The vscode test setup is failing with both `stable` and `insiders` builds and that is because the recent changes from VSCode does not allow having multiple providers for a registered scheme.

Several people attempted a solution, and I got feedback to try to remove the MDBExtensionController stub, and instead use the actual controller that is initialized in the src/extension.ts file. This seemed like a good solution, but it caused more problems than it solved:
- using real controller during unit tests caused links to be opened in browser, files to be opened and saved, and an unknown quantity of other test actions to be performed for real
- Overriding any of the fields or functions from the MDBExtensionController seems to be impossible once it gets registered. Probably there is a hacky way to do it, but they definitely do not want this done.
- A number of tests started failing when run through VSCode UI, which pass when running through CLI, so some context collision is still happening

After doing a thorough exploration in mitigation of above mentioned problems, I concluded that this cannot be the way forward. Aside from everything in the system working against this solution, it would also require an update to many of the current unit tests (about 60 at a minimum).

I though about opening a ticket directly on VSCode github page, but with over 13k open issues, I figured it might not be the optimal solution. Googling for the exact error I was getting when using insiders build returned no exact matches, and tracking down even a StackOverflow thread with the similar issue proved difficult.

For all these reasons I concluded that the solution is either extremely complex, or incredibly trivial.
I went for incredibly trivial, and this PR is that solution.
Using a env flag prevents the real controller from being registered when running tests, and registers ONLY the stubbed one instead. It seems to resolve the issue, and the unit tests run normally. 
There were some issues when running via the VSCode UI, which pertain to specific test cases, so I had to update those as well. 
It was all down to multiple registration/initializations, incomplete test cleanup which caused the first run to be successful, but the subsequent ones to fail, and few other minor things.

In summation, we can now run test from CLI or VSCode UI, the github actions are all working. The version we test against is now "insiders" again. 

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -23,8 +23,7 @@ import MDBExtensionController from './mdbExtensionController';
 
 let mdbExtension: MDBExtensionController | undefined;
 
-const isUnderTest =
-  process.env.NODE_ENV === 'test' || process.env.MDB_IS_TEST === 'true';
+const isUnderTest = process.env.MDB_IS_TEST === 'true';
 
 // Called when our extension is activated.
 // See "activationEvents" in `package.json` for the events that cause activation.


src/extension.ts
```diff
@@ -21,7 +21,9 @@ debug.log = log.debug.bind(log);
 
 import MDBExtensionController from './mdbExtensionController';
 
-let mdbExtension: MDBExtensionController;
+let mdbExtension: MDBExtensionController | undefined;
+
+const isUnderTest = process.env.MDB_UNDER_TEST === 'true';
```
himanshusinghs:
```suggestion
const isUnderTest = process.env.NODE_ENV === "test" && process.env.MDB_UNDER_TEST === 'true';
```

nit and optional: Perhaps to make it more clear that this is a test only setting. Doing this would require setting the correct flag when running tests as well.

gagik: maybe let's get rid of MDB_UNDER_TEST altogether and just use the NODE_ENV?

tculig: Could be problematic
image

tculig: I prefer being explicit about this if you're not married to the idea of using NODE_ENV @gagik

gagik: okay NODE_DEV in particular actually antipattern (TIL), although not necessarily relevant for our case: https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production

"under test" is weird wording for me though, maybe just `MDB_IS_TEST`?

gagik: actually maybe let's just stick to `NODE_ENV === "test"`, this seems still reasonable for our usecase
we basically do want this to be exclusive from any case where this ends up being set to production.
The link I sent was more relevant for specific production code contexts.

src/test/runTest.ts
```diff
@@ -39,7 +39,7 @@ async function main(): Promise<any> {
 
   // Download VS Code, unzip it and run the integration test
   await runTests({
-    version: '1.103.2', // TODO(VSCODE-700) Once we fix the test setup issues, we should revert this to 'insiders'
+    version: 'insiders',
```
himanshusinghs, line 42: <3

src/extension.ts
nirinchev: Would that create problems for us when/if we migrate to using vitest, considering it doesn't automatically set NODE_ENV? Or do we just set it ourselves?

gagik: we'd set it just like any other environment variable yeah

tculig: For the current setup, it actually doesn't set the NODE_ENV to "test" automatically, just FYI

gagik: yeah it's just a matter of naming preferences. The real difference here is that in production environment this is pre-set to `production` so the variable is less random.
`NODE_ENV === 'test'` or `MDB_IS_TEST` is fine by me; but it should be one or the other imo

tculig: So, as it turns out, NODE_ENV is being set to "development" by webpack. It overrides any value that is being passed as a runtime parameter. Thus, it is somewhat unclear when this variable will get hijacked by some other process, and is definitely not 100% under our full control. I would argue for keeping MDB_IS_TEST (renamed from MDB_UNDER_TEST), for consistency and full control @gagik

PR #1189, opened 2025-12-01
Chore/vscode 703 temp build test

tculig:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/.github/workflows/test-and-build-from-fork.yaml
+++ b/.github/workflows/test-and-build-from-fork.yaml
@@ -77,22 +77,3 @@ jobs:
         env:
           PR_URL: ${{github.event.pull_request.html_url}}
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-
-  status-check:
-    name: Test and Build
-    runs-on: ubuntu-latest
-    needs: [build-and-package, test]
-    if: always()
-    steps:
-      - name: Check job results
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/draft-release.yaml
```diff
@@ -1,34 +1,35 @@
...
+          - patch
+          - minor
+          - major
+          - exact-version
 
       exactVersion:
         description: 'Exact version: (Only effective selecting "exact-version" as version bump)'
         required: false
 
-description: |
```
gagik, line 21: this description will appear in github UI right? If it's a valid field that is. If so, it'd be preferrable for it to live here

.github/workflows/actions/build-and-package/action.yaml
```diff
@@ -0,0 +1,143 @@
...
+    description: Snyk token for security scanning
+    required: true
+  JIRA_API_TOKEN:
+    description: Jira API token for vulnerability tickets
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install Deps Ubuntu
```
gagik, line 32: this should also setup npm / node right?
in which case some of the other actions need to be adjusted too

.github/workflows/draft-release.yaml
nirinchev: It's not a valid field for top-level actions.

PR #1191, opened 2025-12-02
chore: npm -> pnpm VSCODE-714

nirinchev:
## Description

This PR migrates the vscode project from npm to pnpm. npm has been a source of frustration on various levels and with pnpm we're getting a more predictable and significantly faster experience. There are a handful of code changes related to the migration as well as a bump of the mcp server package.

### DIFF
--- a/.github/workflows/actions/test-and-build/action.yaml
+++ b/.github/workflows/actions/test-and-build/action.yaml
@@ -143,7 +143,7 @@ runs:
         JIRA_PROJECT: "VSCODE"
         JIRA_VULNERABILITY_BUILD_INFO: "- [GitHub Run|https://github.com/mongodb-js/vscode/actions/runs/${{github.run_id}}/jobs/${{github.job}}]"
       run: |
-        npm run create-vulnerability-tickets > /dev/null
+        pnpm run create-vulnerability-tickets > /dev/null
 
     - name: Generate Vulnerability Report (Fail on >= High)
       if: runner.os == 'Linux'
@@ -154,4 +154,4 @@ runs:
         # available for everyone with read access to the repo, which is everyone that is
         # logged in for public repos.
         # This command is only here to fail on failures for `main` and tags.
-        npm run generate-vulnerability-report > /dev/null
+        pnpm run generate-vulnerability-report > /dev/null
--- a/src/mcp/mcpController.ts
+++ b/src/mcp/mcpController.ts
@@ -214,7 +214,7 @@ export class MCPController {
       });
 
       const runner = new StreamableHttpRunner({
-        userConfig: UserConfigSchema.parse(mcpConfig),
+        userConfig: mcpConfig,
         createConnectionManager: (...params) =>
           MCPController.createConnectionManager(this, ...params),
         connectionErrorHandler: createMCPConnectionErrorHandler(
@@ -239,25 +239,22 @@ export class MCPController {
     }
   }
 
-  private getMCPServerConfig(
-    headers: Record<string, string>,
-  ): Partial<UserConfig> {
-    const vscodeConfiguredMCPConfig = getMCPConfigFromVSCodeSettings();
-    return {
-      ...vscodeConfiguredMCPConfig,
+  private getMCPServerConfig(headers: Record<string, string>): UserConfig {
+    const configFromSettings = getMCPConfigFromVSCodeSettings();
+    const vsCodeConfig: Partial<UserConfig> = {
+      ...configFromSettings,
       transport: 'http',
       httpPort: 0,
       httpHeaders: headers,
       disabledTools: Array.from(
-        new Set([
-          'connect',
-          ...(vscodeConfiguredMCPConfig.disabledTools ?? []),
-        ]),
+        new Set(['connect', ...(configFromSettings.disabledTools ?? [])]),
       ),
       loggers: Array.from(
-        new Set(['mcp', ...(vscodeConfiguredMCPConfig.loggers ?? [])]),
+        new Set(['mcp', ...(configFromSettings.loggers ?? [])]),
       ),
     };
+
+    return UserConfigSchema.parse(vsCodeConfig);
   }
 
   private static async createConnectionManager(


src/test/suite/connectionController.test.ts
```diff
@@ -220,14 +214,8 @@ suite('Connection Controller Test Suite', function () {
...
 
-      const connectionIds = Object.keys(testConnectionController._connections);
-      const latestConnectionId = connectionIds[connectionIds.length - 1];
-
-      const anonymousId =
-        testConnectionController._connectionStorage.getUserAnonymousId();
-
       expect(mongoClientConnectionOptions).to.deep.equal({
-        url: `mongodb://localhost:27088/?appName=mongodb-vscode+${version}--${anonymousId}--${latestConnectionId}`,
+        url: `mongodb://localhost:27088/?appName=mongodb-vscode+${version}`,
```
gagik, line 218: how was this green before?

src/mcp/mcpConfig.ts
```diff
@@ -16,6 +12,8 @@ export function getMCPConfigFromVSCodeSettings(
   retrieveMCPConfiguration: () => vscode.WorkspaceConfiguration = (): vscode.WorkspaceConfiguration =>
     vscode.workspace.getConfiguration('mdb.mcp'),
 ): Partial<UserConfig> {
+  const configurableProperties = new Set(Object.keys(UserConfigSchema.shape));
```
gagik: Keys of a map are essentially a set so I don't think we're gaining much out of this `new Set` bit here.

gagik (approved):
Minor note about the set + question

.github/workflows/actions/test-and-build/action.yaml
```diff
@@ -57,10 +57,10 @@ runs:
     - name: Install Dependencies
       shell: bash
       run: |
-        npm ci --omit=optional
+        pnpm install --frozen-lockfile
```
himanshusinghs, line 60: Does pnpm by default omit optional? I faintly recall there were some transitive optional dependencies that were problematic for the install here.

scripts/snyk-test.js
```diff
@@ -2,60 +2,9 @@ const childProcess = require('child_process');
...
- * the following:
- * - "@vscode/vsce-sign-alpine-arm64"
- * - "@vscode/vsce-sign-alpine-x64"
- * - "@vscode/vsce-sign-darwin-arm64"
- * - "@vscode/vsce-sign-darwin-x64"
- * - "@vscode/vsce-sign-linux-arm"
- * - "@vscode/vsce-sign-linux-arm64"
- * - "@vscode/vsce-sign-linux-x64"
- * - "@vscode/vsce-sign-win32-arm64"
- * - "@vscode/vsce-sign-win32-x64"
```
himanshusinghs, line 23: Yea these were the ones! How does pnpm address these?

src/mcp/mcpController.ts
```diff
@@ -239,11 +239,11 @@ export class MCPController {
     }
   }
 
-  private getMCPServerConfig(headers: Record<string, string>): UserConfig {
+  private getMCPServerConfig(
+    headers: Record<string, string>,
+  ): Partial<UserConfig> {
```
himanshusinghs: nit: Can we just return the UserConfig from here instead of the partial, resonates well with the interface?

src/telemetry/connectionTelemetry.ts
```diff
@@ -153,7 +153,7 @@ export async function getConnectionTelemetryProperties(
       dl_version: dataLake.version,
       is_enterprise: build.isEnterprise,
       is_genuine: genuineMongoDB.isGenuine,
-      non_genuine_server_name: genuineMongoDB.dbType,
+      non_genuine_server_name: genuineMongoDB.serverName,
```
himanshusinghs, line 156: 🦅 eyes

src/test/suite/connectionController.test.ts
```diff
@@ -120,16 +120,13 @@ suite('Connection Controller Test Suite', function () {
...
-      const anonymousId =
-        testConnectionController._connectionStorage.getUserAnonymousId();
-
       // The stored connection string should not have the appName appended
       expect(connection.connectionOptions.connectionString).equals(
         `${TEST_DATABASE_URI}/`,
       );
       // But the active connection should
       expect(testConnectionController.getActiveConnectionString()).equals(
-        `${TEST_DATABASE_URI}/?appName=mongodb-vscode+${version}--${anonymousId}--${connectionId}`,
```
himanshusinghs, line 132: This seems wrong - it should append the anonymous id though? Or does that happen somewhere else?

himanshusinghs (changes requested):

scripts/snyk-test.js
gagik: this seems only relevant in the build step.
I think we were omitting optional dependencies just so we could create a workaround around us doing this, which is weird. I think this new approach should be fine unless I'm overlooking something

src/mcp/mcpController.ts
gagik: it's partial now because `defaultUserConfig` isn't included in it. at least that's my impression

src/test/suite/connectionController.test.ts
nirinchev: This only gets appended when we connect to atlas: https://github.com/mongodb-js/compass/blob/305f949daaa869fd40105cf67bb41cd6940e67b5/packages/connection-form/src/utils/set-app-name-if-missing.ts#L25

src/mcp/mcpController.ts
nirinchev: Yeah, we do `UserConfig.parse` with the partial. We can do either way - I'm fine with adding the defaults here and making it fully defined.

.github/workflows/actions/test-and-build/action.yaml
nirinchev: If we omit the optional dependencies when installing, we're getting build errors: https://github.com/mongodb-js/vscode/actions/runs/19892320307/job/57013930209?pr=1191 - i.e. the atlas-local optional dependencies are missing when we try to webpack.

src/test/suite/connectionController.test.ts
nirinchev: This changed after we bumped the connection form dependency. Previously, npm had it pinned on an old version.

.github/workflows/actions/test-and-build/action.yaml
himanshusinghs: Interesting - then I'd be worried about snyk output as well. We might need to double check if snyk is even evaluating the correct packages or not.

scripts/snyk-test.js
himanshusinghs: It was an unfortunate workaround because Snyk decided not to ignore the dev dependencies for analysis but still a valid one because without it Snyk would fail and never get to analyse production dependencies.

Without much understanding of how Snyk plays with pnpm, I would assume that Snyk setup is broken again but I will check.

src/mcp/mcpController.ts
himanshusinghs: Yea - currently we compensate the lack of default config by doing UserConfigSchema.parse on the result of this function and I was suggesting to do it in the function itself so that the result resonate well with the function.

src/test/suite/connectionController.test.ts
himanshusinghs: Should we have a test to ensure the id is appended for Atlas as well? If that is not already there?

scripts/snyk-test.js
himanshusinghs: I checked it locally and it seems to work just fine. Snyk was indeed detecting errors from our installed packages. I am curious now how snyk interacts with pnpm lock file? 😄

src/test/suite/connectionController.test.ts
nirinchev: I don't think that's there already, but I'd rather leave that to a separate PR - it's a bit annoying to write.

scripts/snyk-test.js
nirinchev: I re-enabled snyk vuln report generation and also enabled uploading the report in the gha run so that we can reference it. There are a few recommendations, but will take a look at those in another PR.

.github/workflows/actions/test-and-build/action.yaml
himanshusinghs: Double checked it locally and it works as expected.

himanshusinghs (approved):
pnpm pnpm pnpm 🚀

CONTRIBUTING.md
```diff
@@ -24,13 +24,13 @@ Please include as much information as possible about your environment.
...
-Running the MongoDB VSCode plugin requires [Node.js](https://nodejs.org) and npm.
+Running the MongoDB VSCode plugin requires [Node.js](https://nodejs.org) and pnpm.
 
 1. Clone this project, navigate to the folder, then run:
 
 ```shell
-npm install
-npm run watch
+pnpm install
+pnpm run watch
```
lerouxb, line 33: I think this was already wrong - just running the extension from inside vscode does `pnpm run watch` already, so at best this is unnecessary.

PR #1193, opened 2025-12-07
chore: block certain commands from links

gagik:


src/mdbExtensionController.ts
```diff
@@ -258,6 +258,30 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+        EXTENSION_COMMANDS.MDB_DROP_STREAM_PROCESSOR,
+        EXTENSION_COMMANDS.MDB_REMOVE_CONNECTION,
+        EXTENSION_COMMANDS.MDB_REMOVE_CONNECTION_TREE_VIEW,
+      ];
+
+      if (disallowedDeepLinkCommands.includes(command as EXTENSION_COMMANDS)) {
+        throw new Error(
+          `Command '${command}' cannot be invoked via deep links.`,
+        );
+      }
```
addaleax, line 360: Non-blocking: Would it make sense to keep a list of allowed commands?

In Compass, we maintain two lists, one of all allowed driver options in connection string and one of all disallowed ones, and through TS we check that they are disjoint and their union is the list of all driver-supported options; that way, if the list of driver options changes, somebody has to modify the lists to make sure we've accounted for it

addaleax (approved):

PR #1202, opened 2025-12-08
chore: fix connection form

gagik:
Fixes issues with connection form by removing the override and letting compass-components resolve to a newer version.

We should do some proper manual testing again.

CI is getting stuck... no clue why

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "pretest": "pnpm run compile",
     "test": "pnpm run test-webview && pnpm run test-extension",
     "test-extension": "cross-env MDB_IS_TEST=true NODE_OPTIONS=--no-force-async-hooks-checks xvfb-maybe node ./out/test/runTest.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -55,7 +55,7 @@
     "pretest": "pnpm run compile",
     "test": "pnpm run test-webview && pnpm run test-extension",
     "test-extension": "cross-env MDB_IS_TEST=true NODE_OPTIONS=--no-force-async-hooks-checks xvfb-maybe node ./out/test/runTest.js",
-    "test-webview": "mocha -r ts-node/register --grep=\"${MOCHA_GREP}\" --file ./src/test/setup-webview.ts src/test/suite/views/webview-app/**/*.test.tsx",
+    "test-webview": "mocha -r ts-node/register --exit --grep=\"${MOCHA_GREP}\" --file ./src/test/setup-webview.ts src/test/suite/views/webview-app/**/*.test.tsx",
```
gagik: I don't like that this is necessary.... I am not sure why; seems like we have a dangling process which does not cleaned up but I tried many things and never managed to force it to close on its own

PR #1205, opened 2025-12-16
chore: update watch:extension npm script to pass watch to tsc

Anemy:
I usually use this [launch command](https://github.com/mongodb-js/vscode/blob/c84482b4aee4832fe5654f89a73e100217842e29/.vscode/launch.json#L9) to manually test extension changes:

Screenshot 2025-12-16 at 17 18 31

It was erroring as I don't think pnpm passes the watch the same as npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "compile:extension": "tsc -p ./",
     "compile:extension-bundles": "webpack --mode development",
     "watch": "pnpm run compile && npm-run-all -p watch:*",
-    "watch:extension": "tsc -p ./ -watch",
+    "watch:extension": "tsc -p ./ --watch",
     "watch:extension-bundles": "webpack --mode development --watch",
     "pretest": "pnpm run compile",
     "test": "pnpm run test-webview && pnpm run test-extension",


package.json
```diff
@@ -50,7 +50,7 @@
     "compile:extension": "tsc -p ./",
     "compile:extension-bundles": "webpack --mode development",
     "watch": "pnpm run compile && npm-run-all -p watch:*",
-    "watch:extension": "pnpm run compile:extension -- -watch",
+    "watch:extension": "tsc -p ./ --watch",
```
kraenhansen, line 53: If you're willing to remove node v20 from the engines field of the package.json, you could rely on a Node.js CLI feature added with Node.js v22.0.0:

```suggestion
    "watch:extension": "node --run compile:extension -- --watch",
```

gagik: we need to be careful with Node versioning because VSCode tends to lag behind a bit, but v20 might be old enough where this is fine; just needs to be checked

Anemy:
I unintentionally included this change in the lint pr, so I'll close this pr. I'll include in my next pr the watch argument passing (thanks for the suggestion!) so we don't duplicate `compile:extension`.

PR #1206, opened 2025-12-17
chore(lint): use devtools shared lint, fix errors VSCODE-723

Anemy:
VSCODE-723

Updates our eslint to pull from https://github.com/mongodb-js/devtools-shared/tree/main/configs/eslint-config-devtools

Fixes the errors that it created.

### DIFF
--- a/src/test/suite/mdbExtensionController.test.ts
+++ b/src/test/suite/mdbExtensionController.test.ts
@@ -1848,7 +1848,7 @@ suite('MDBExtensionController Test Suite', function () {
 
       expect(fakeExecuteCommand).to.not.have.been.called;
       expect(fakeShowErrorMessage).to.have.been.calledOnceWith(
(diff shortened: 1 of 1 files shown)


src/participant/docsChatbotAIService.ts
```diff
@@ -194,13 +200,17 @@ export class DocsChatbotAIService {
       try {
         data = await res.json();
       } catch (error) {
-        throw new Error(`[Docs chatbot] Internal server error: ${error}`);
+        throw new Error(
+          `[Docs chatbot] Internal server error: ${(error as Error).message}`,
```
gagik: are we always sure this will be of type `Error`?

gagik: It might be easier / safer to do `String(error)` across the board

gagik (approved):
awesome! One note about error string conversion

src/participant/docsChatbotAIService.ts
Anemy: Maybe a helper like
```
function errorToString(err: unknown): string {
  return err && typeof err === 'object' && 'message' in err
    ? String(err.message)
    : String(err);
}
```
would be best. I'll update.

PR #1209, opened 2025-12-18
chore(lint): bump shared lint dependency, fix issues

Anemy:


src/explorer/connectionTreeItem.ts
```diff
@@ -215,7 +215,8 @@ export default class ConnectionTreeItem
     const databases = await this.listDatabases();
     databases.sort((a: string, b: string) => a.localeCompare(b));
 
-    const newChildrenCache: Record<string, DatabaseTreeItem> = {};
+    const newChildrenCache: Record<string, DatabaseTreeItem> =
```
tculig, line 218: Why this change?

Anemy: That one is from pulling in a new rule https://github.com/mongodb-js/devtools-shared/blob/main/configs/eslint-plugin-devtools/rules/no-plain-object-records.js

We added in https://github.com/mongodb-js/devtools-shared/pull/604
The rule might not be suggesting quite the right fix here though, it's breaking the ts check in other places when it's applied. Looking into it a bit more.

lerouxb: There are pros and cons and I don't think there's really a right answer:
https://stackoverflow.com/questions/32262809/is-it-bad-practice-to-use-object-createnull-versus

It kinda both breaks AND fixes `for ... in` which I personally never use anyway. But library code often does.

PR #1211, opened 2026-01-02
fix(settings): update shell and OIDC browser settings to be application scope VSCODE-726

Anemy:
VSCODE-726

https://code.visualstudio.com/api/references/contribution-points#contributes.configuration

### DIFF
--- a/src/test/suite/oidc.test.ts
+++ b/src/test/suite/oidc.test.ts
@@ -91,6 +91,8 @@ suite('OIDC Tests', function () {
   let createTerminalStub: SinonStub;
   let sendTextStub: SinonStub;
 
+  let originalBrowserCommandForOIDCAuth;
+
   before(async function () {
     if (process.platform !== 'linux') {
       // OIDC is only supported on Linux in the 7.0+ enterprise server.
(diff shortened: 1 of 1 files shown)


src/explorer/playgroundsTree.ts
```diff
@@ -11,7 +11,6 @@ const log = createLogger('playgrounds tree');
 export default class PlaygroundsTree
   implements vscode.TreeDataProvider<vscode.TreeItem>
 {
-  public excludeFromPlaygroundsSearch: string[];
```
Anemy, line 14: Drive by, this property wasn't used except in a test. Happy to move into another pr if folks would prefer.

src/commands/launchMongoShell.ts
```diff
@@ -74,6 +74,13 @@ const openMongoDBShell = (
     return Promise.resolve(false);
   }
 
+  if (shellCommand !== 'mongo' && shellCommand !== 'mongosh') {
+    void vscode.window.showErrorMessage(
+      'Invalid MongoDB shell command specified. Please set the shell command to "mongo" or "mongosh" in the MongoDB extension settings.',
+    );
```
addaleax, line 80: Just reminds me, do we have a JIRA/Aha entry for using mongosh's bundled source from the extension instead of an external executable? In theory, that should work fairly well for most cases

addaleax (approved):
LGTM!

src/commands/launchMongoShell.ts
Anemy: I haven't seen something on Jira. I'll create a ticket and link back here so that we have it tracked at least. It would be a nice feature add. https://jira.mongodb.org/browse/VSCODE-731

PR #1214, opened 2026-01-07
feat: create entry point for new data browsing experience VSCODE-722

tculig:
image

<h1>PR Summary: VSCODE-722 – Enhanced Data Browsing Experience</h1>

<h2>Overview</h2>

Introduces a **webview-based data browsing experience** for viewing collection documents, controlled by a feature flag (**`MDB_USE_ENHANCED_DATA_BROWSING_EXPERIIENCE`**).
The existing tree-based document list remains the default.

---

<h2>Key Changes</h2>

<h3>New Data Browsing Infrastructure</h3>

- **DataBrowsingController**
  New controller managing webview panels for document browsing, with support for abort controllers to cancel in-flight requests when new ones are initiated.

- **ShowPreviewTreeItem**
  New tree item that replaces **DocumentListTreeItem** when the feature flag is enabled. Clicking it opens the document preview webview.

- **preview-page.tsx**
  React component for the document preview webview with loading state, dark mode support, and message handling.

- **webviewHelpers.ts**
  Extracted common webview utilities (**createWebviewPanel**, **getWebviewHtml**, **getNonce**) for reuse between the connection dialog and the data browser.

---

<h3>Tree View Updates</h3>

- **CollectionTreeItem** now conditionally renders either:
  - **ShowPreviewTreeItem** (new)
  - **DocumentListTreeItem** (existing)
  based on the feature flag.

- Added **documentUtils.ts** to share the **CollectionType** enum and **formatDocCount** helper between tree items.

---

<h3>Extension Controller Integration</h3>

- Registered new command **`mdb.internal.openCollectionPreviewFromTreeView`** to handle clicking the preview tree item.
- Instantiated and wired up **DataBrowsingController** with proper lifecycle management (deactivation and cleanup).

---

<h3>Webview Controller Refactoring</h3>

- Refactored **webviewController.ts** to use the shared **webviewHelpers** utilities, reducing code duplication.

---

<h3>Message Protocol</h3>

- Added **extension-app-message-constants.ts** defining message types (e.g. **getDocuments**, **loadDocuments**, etc.) and interfaces for webview ↔ extension communication.

---

<h3>Test Infrastructure</h3>

- Added unit tests for **DataBrowsingController** covering abort controller behavior, message handling, and panel lifecycle.
- Added React component tests for **PreviewApp** covering loading states and message handling.
- Added a polyfill for the Constructable Stylesheets API (**CSSStyleSheet.replaceSync**) in test setup to support **@vscode-elements/react-elements**.

---

<h3>Dependencies</h3>

- Added **@vscode-elements/react-elements** for a VS Code–styled loading spinner.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/.github/workflows/publish-release.yaml
+++ b/.github/workflows/publish-release.yaml
@@ -12,24 +12,17 @@ jobs:
     name: Publish Release
     runs-on: ubuntu-latest
     steps:
-      - name: Setup pnpm
-        uses: pnpm/action-setup@v4
-        with:
-          version: 10.23.0 # Aligned with "packageManager" in package.json.
-          run_install: false
-
       - name: Setup Node.js Environment
         uses: actions/setup-node@v4
         with:
           node-version: 22.21.1
-          cache: pnpm
 
       - name: Run node-gyp bug workaround script
         run: |
           curl -sSfLO https://raw.githubusercontent.com/mongodb-js/compass/42e6142ae08be6fec944b80ff6289e6bcd11badf/.evergreen/node-gyp-bug-workaround.sh && bash node-gyp-bug-workaround.sh
 
       - name: Install VSCode publishing dependencies
-        run: pnpm add -g vsce
+        run: npm install -g vsce
 
       - name: Download release assets
         run: |
@@ -50,6 +43,6 @@ jobs:
       - name: Publish to marketplace
         run: |
           echo "Publishing ${VSIX_PATH} to marketplace"
-          pnpm exec vsce publish --packagePath "${VSIX_PATH}" --githubBranch main
+          npx vsce publish --packagePath "${VSIX_PATH}" --githubBranch main
         env:
           VSCE_PAT: ${{ secrets.VSCE_PAT }}


src/test/suite/views/data-browsing-app/preview-page.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+          new MessageEvent('message', {
+            data: {
+              command: PreviewMessageType.loadDocuments,
+              documents: [{ _id: { $oid: '123' }, name: 'Test' }],
+              totalCount: 1,
+            },
+          })
+        );
+        // Wait for minimum loading duration
+        await new Promise((resolve) => setTimeout(resolve, 600));
```
Anemy: Can we avoid a constant timeout here and instead wait for the end state? Same with the `setTimeout`s below. There's a `waitFor` that some tests use.
It'll speed up all of our tests if we use that as a pattern.

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -0,0 +1,411 @@
...
+type ViewType = 'tree' | 'json' | 'table';
+
+const ITEMS_PER_PAGE_OPTIONS = [10, 25, 50, 100];
+
+const toolbarStyles = css({
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'space-between',
+  padding: `${spacing[200]}px ${spacing[300]}px`,
+  borderBottom: '1px solid var(--vscode-panel-border, #444)',
```
Anemy: We're going to end up using a lot of these vscode variables. How do you feel about having them all in one place? Like vscode-styles and we reference this panelBorder string from there.

src/views/webview-app/app.tsx
```diff
@@ -1,16 +1,31 @@
...
+import React, { lazy, Suspense } from 'react';
+import { LeafyGreenProvider } from '@mongodb-js/compass-components';
+import { useDetectVsCodeDarkMode } from './use-detect-vscode-dark-mode';
+import type { WebviewType } from '../../utils/webviewHelpers';
+
+// Use lazy loading to avoid loading both vscode-api modules at once.
+// Each vscode-api.ts calls acquireVsCodeApi() at module load time,
+// and acquireVsCodeApi() can only be called once per webview.
+const OverviewPage = lazy(() => import('./overview-page'));
+const PreviewPage = lazy(() => import('../data-browsing-app/preview-page'));
```
Anemy: I haven't seen this in use before, cool.

src/views/webview-app/app.tsx
```diff
@@ -1,16 +1,31 @@
...
+  }
+}
+
+const App: React.FC = () => {
+  const darkMode = useDetectVsCodeDarkMode();
+  const webviewType = window.WEBVIEW_TYPE;
+
+  return (
+    <LeafyGreenProvider darkMode={darkMode}>
+      <Suspense fallback={null}>
```
Anemy: Will folks bump into this suspense often?
We could use the `VscodeProgressRing` from `@vscode-elements/react-elements` if so https://vscode-elements.github.io/components/progress-ring/

src/views/dataBrowsingController.ts
```diff
@@ -0,0 +1,329 @@
...
+  webview,
+}: {
+  extensionPath: string;
+  webview: vscode.Webview;
+}): string => {
+  return getWebviewHtml({
+    extensionPath,
+    webview,
+    webviewType: 'dataBrowser',
+    title: 'MongoDB Data Browser',
```
Anemy: In [the designs](https://www.figma.com/design/QE0mDJK9hlbPK9shnQCIbG/VS-CODE--DATA-BROWSING?node-id=1749-175007&t=gXiQ4i9Y1OZiUDYh-4) we have the tab title as the database.collection - let's follow that here?

src/mdbExtensionController.ts
```diff
@@ -847,6 +855,39 @@ export default class MDBExtensionController implements vscode.Disposable {
         return this._editorsController.onViewCollectionDocuments(namespace);
       },
     );
+    this.registerCommand(
+      ExtensionCommand.mdbOpenCollectionPreviewFromTreeView,
+      async (element: ShowPreviewTreeItem): Promise<boolean> => {
+        const namespace = element.namespace;
+        // Fetch a batch of documents for client-side pagination
+        const fetchLimit = 100;
```
Anemy: 100 seems like a lot to start with, how do you feel about 10 documents instead?

src/views/dataBrowsingController.ts
```diff
@@ -0,0 +1,329 @@
...
+  });
+};
+
+export interface DataBrowsingOptions {
+  namespace: string;
+  documents: Document[];
+  fetchDocuments?: (options?: {
+    sort?: SortOption;
```
Anemy: Can we add the sort option in another pr? This pr already has a lot in it.

src/views/dataBrowsingController.ts
```diff
@@ -0,0 +1,329 @@
...
+    panel: vscode.WebviewPanel,
+    options: DataBrowsingOptions,
+  ): Promise<void> => {
+    const abortController = this._createAbortController(panel);
+    const { signal } = abortController;
+
+    try {
+      const totalCount = options.getTotalCount
+        ? await options.getTotalCount(signal)
+        : options.initialTotalCount;
```
Anemy: I'm a bit confused here, is the handleGetDocuments returning the count? It looks like it's getting passed the documents from the webview.

src/views/dataBrowsingController.ts
```diff
@@ -0,0 +1,329 @@
...
+      });
+    }
+  };
+
+  handleRefreshDocuments = async (
+    panel: vscode.WebviewPanel,
+    options: DataBrowsingOptions,
+  ): Promise<void> => {
+    const abortController = this._createAbortController(panel);
+    const { signal } = abortController;
```
Anemy: If a users clicks to refresh twice, can we abort the previous request? That way we're not spending resources on calls that we know we want to discard. Another way to achieve this would be to disable the refresh button when we're waiting for documents, but that could also have some race conditions if the user performs another request like a pagination in the meantime. Should we abort any ongoing find/refresh to be safe?

src/views/dataBrowsingController.ts
```diff
@@ -0,0 +1,329 @@
...
+    context: vscode.ExtensionContext,
+    options: DataBrowsingOptions,
+  ): vscode.WebviewPanel {
+    log.info('Opening data browser...', options.namespace);
+    const extensionPath = context.extensionPath;
+
+    // Create and show a new data browsing webview.
+    const panel = createWebviewPanel({
+      viewType: 'mongodbDataBrowser',
+      title: `Preview: ${options.namespace}`,
```
Anemy: We have the title for the panel in the html in getWebviewHtml(), how does this title differ? Should they be the same?

src/views/data-browsing-app/document-tree-view.tsx
```diff
@@ -0,0 +1,465 @@
+import React, { useState } from 'react';
+import { css, cx, spacing } from '@mongodb-js/compass-components';
+
+// Shared typography for monospace code display
+const codeTypography = {
+  fontFamily: 'Menlo, Monaco, "Courier New", monospace',
```
Anemy: Looks like there's a `--vscode-editor-font-family` vscode css var. Should we use that here? Mentioned in here https://code.visualstudio.com/api/extension-guides/webview#theming-webview-content

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -0,0 +1,411 @@
...
+  sendSortDocuments,
+} from './vscode-api';
+import DocumentTreeView from './document-tree-view';
+
+interface PreviewDocument {
+  [key: string]: unknown;
+}
+type ViewType = 'tree' | 'json' | 'table';
+
+const ITEMS_PER_PAGE_OPTIONS = [10, 25, 50, 100];
```
Anemy: Can we add these later in another pr? This pr already has a lot in it. We likely want to save this in the extension settings as well.

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -0,0 +1,411 @@
...
+  sendGetDocuments,
+  sendRefreshDocuments,
+  sendSortDocuments,
+} from './vscode-api';
+import DocumentTreeView from './document-tree-view';
+
+interface PreviewDocument {
+  [key: string]: unknown;
+}
+type ViewType = 'tree' | 'json' | 'table';
```
Anemy: We won't have a table view, let's remove that.

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -0,0 +1,411 @@
...
+
+  const handleItemsPerPageChange = (value: string): void => {
+    const newItemsPerPage = parseInt(value, 10);
+    setItemsPerPage(newItemsPerPage);
+    setCurrentPage(1); // Reset to first page when changing items per page
+  };
+
+  const handleViewTypeChange = (value: string): void => {
+    setViewType(value as ViewType);
+    // TODO: Implement different view renderings
```
Anemy: Let's aim to not have TODOs unless they have a ticket mentioned in them, like `TODO(VSCODE-NUM): ...`

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -0,0 +1,411 @@
...
+  return (
+    <div
+      style={{
+        backgroundColor: darkMode ? '#1E1E1E' : '#FFFFFF',
+        minHeight: '100vh',
+        color: darkMode ? '#CCCCCC' : '#000000',
+      }}
+    >
+      {/* Toolbar */}
+      <div className={toolbarStyles}>
```
Anemy: `@vscode-elements/react-elements` has a toolbar container and toolbar buttons. Should we be using those?
https://vscode-elements.github.io/components/toolbar-container/

src/views/data-browsing-app/document-tree-view.tsx
Anemy, line 1: Can we not add any code formatting for now and instead do that in a follow up https://jira.mongodb.org/browse/VSCODE-727 ?

Anemy (commented):
Left a couple comments to start the discussion. There is a lot in this pr, it would likely speed things up if we broke it into multiple, by first adding just the [new webview entry](https://jira.mongodb.org/browse/VSCODE-722), and then [adding the documents list](https://jira.mongodb.org/browse/VSCODE-728), and after that the [formatting for the documents](https://jira.mongodb.org/browse/VSCODE-727).

src/test/suite/views/data-browsing-app/preview-page.test.tsx
tculig: hm, good point...the idea was to not have loads so fast that the user isn't sure if anything happened at all. We can just use sinon.useFakeTimers(); in the unit tests.

src/views/dataBrowsingController.ts
tculig: updated

src/views/dataBrowsingController.ts
tculig: yes

src/views/dataBrowsingController.ts
tculig: This is the VS Code tab title, the other one is the html document inside the webview title.

src/views/webview-app/app.tsx
tculig: it actually doesn't even work without this. It loads both pages and gets all confused...

src/views/webview-app/app.tsx
tculig: updated

src/views/dataBrowsingController.ts
tculig: I'm not 100% on this one, I'll look into it...

src/mdbExtensionController.ts
```diff
@@ -847,6 +855,34 @@ export default class MDBExtensionController implements vscode.Disposable {
...
+        const fetchDocuments = async (options?: {
+          sort?: 'default' | 'asc' | 'desc';
+          limit?: number;
+          signal?: AbortSignal;
+        }): Promise<Document[]> => element.loadPreview(options);
+
+        const getTotalCount = (signal?: AbortSignal): Promise<number> =>
+          element.getTotalCount(signal);
+
+        this._dataBrowsingController.openDataBrowser(this._context, {
```
Anemy, line 861: Can we show the view and then show the find results after? That way things are faster for the user. I think we can have things going in parallel.

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -0,0 +1,98 @@
...
+    sendGetDocuments();
+
+    return () => {
+      window.removeEventListener('message', handleMessage);
+    };
+  }, []);
+
+  return (
+    <div
+      style={{
```
Anemy: Can we follow the pattern for styles that the rest of the webview does?
`className={cssVariableDeclaredAtTheTop}`

I understand that this will change in the next pr when we start to show the toolbar and documents, however we should avoid introducing new patterns incase someone else picks it up and assumes it's the way to go.

src/views/data-browsing-app/vscode-api.ts
Anemy, line 1: There's already a vscode-api file in the webview-app, can we share some of this logic?

tculig: I was on the fence whether to keep these separate or merge in one file...will update as you suggested

src/commands/index.ts
```diff
@@ -75,6 +75,10 @@ export const ExtensionCommand = {
   mdbStopStreamProcessor: 'mdb.stopStreamProcessor',
   mdbDropStreamProcessor: 'mdb.dropStreamProcessor',
 
+  // Commands for the data browsing upgrade.
+  mdbOpenCollectionPreviewFromTreeView:
+    'mdb.internal.openCollectionPreviewFromTreeView',
```
gagik: if we want to `.internal` as a pattern, I'd expect us to do it to all the internal functions but this isn't the case right now so maybe let's stick to the old thing?

src/featureFlags.ts
```diff
@@ -1,6 +1,6 @@
 const FEATURE_FLAGS = {
-  useOldConnectionForm:
```
gagik, line 2: is this unused?

package.json
```diff
@@ -54,7 +54,7 @@
     "pretest": "pnpm run compile",
     "test": "pnpm run test-webview && pnpm run test-extension",
     "test-extension": "cross-env MDB_IS_TEST=true NODE_OPTIONS=--no-force-async-hooks-checks xvfb-maybe node ./out/test/runTest.js",
-    "test-webview": "mocha -r ts-node/register --exit --grep=\"${MOCHA_GREP}\" --file ./src/test/setup-webview.ts src/test/suite/views/webview-app/**/*.test.tsx",
+    "test-webview": "mocha -r ts-node/register --exit --grep=\"${MOCHA_GREP}\" --file ./src/test/setup-webview.ts src/test/suite/views/**/**/*.test.tsx",
```
gagik: the test seems to be no longer about testing the webview with this change

src/views/webview-app/vscode-api.ts
```diff
@@ -77,18 +85,18 @@ export const trackExtensionLinkClicked = (
...
 export const openTrustedLink = (linkTo: string): void => {
-  vscode.postMessage({
+  getVSCodeApi().postMessage({
     command: MessageType.openTrustedLink,
     linkTo,
   });
 };
 
-export default vscode;
+export default getVSCodeApi;
```
gagik: one of the reasons why I dislike default is that this rename will have no actual effect 😢 

can we turn it into a named export while we're at it?

src/test/suite/mdbExtensionController.test.ts
```diff
@@ -29,11 +30,11 @@ import { VIEW_COLLECTION_SCHEME } from '../../editors/collectionDocumentsProvide
 import type { CollectionDetailsType } from '../../explorer/collectionTreeItem';
 import { expect } from 'chai';
 import { DeepLinkTelemetryEvent } from '../../telemetry';
-import { CollectionType } from '../../explorer/documentListTreeItem';
 import {
   DEEP_LINK_ALLOWED_COMMANDS,
   DEEP_LINK_DISALLOWED_COMMANDS,
 } from '../../mdbExtensionController';
+import * as featureFlags from '../../featureFlags';
```
gagik: can we just do named exports?

tculig: Typically I would agree, but then I would need to refactor the stubbing mechanism to something much more complicated, and I don't really want to complicate things (this line: sandbox.stub(featureFlags, 'getFeatureFlag').returns(false);))

package.json
tculig: I'm with you on the named imports

src/featureFlags.ts
tculig: it seems to be unused. I can't find any references to it.

src/commands/index.ts
tculig: roger roger

src/views/data-browsing-app/vscode-api.ts
tculig: taking a second look at it, it really feels like this should be separated per folder. Its going to become a huge mess down the line if we cram it all in one file.

src/test/suite/mdbExtensionController.test.ts
addaleax: Fwiw, that kind of stubbing only works because we don't work with actual ESM yet, only transpiled code. Once we do, it's going to break

I'd check whether it's possible to avoid stubbing in tests entirely and, if necessary, export something from `featureFlags` to temporarily set feature flags in tests.

tculig: updated

src/views/data-browsing-app/preview-page.tsx
tculig: removed

src/mdbExtensionController.ts
tculig: updated

src/featureFlags.ts
```diff
@@ -12,6 +15,15 @@ export const getFeatureFlag = (flag: FeatureFlag): boolean => {
...
 };
 
+export const setFeatureFlag = (flag: FeatureFlag, value: boolean): void => {
+  FEATURE_FLAGS[flag] = value;
+};
+
+export const resetFeatureFlags = (): void => {
+  FEATURE_FLAGS.useEnhancedDataBrowsingExperience =
+    process.env.MDB_USE_ENHANCED_DATA_BROWSING_EXPERIENCE === 'true';
+};
```
addaleax, line 19: This duplicates the initialization of `FEATURE_FLAGS` – can we just call this function once and leave it at
`const FEATURE_FLAGS: Record<string, boolean> = Object.create(null);` above?

src/views/webview-app/vscode-api.ts
```diff
@@ -10,12 +10,20 @@ interface VSCodeApi {
...
 
 declare const acquireVsCodeApi: () => VSCodeApi;
-const vscode = acquireVsCodeApi();
+
+let vscode: VSCodeApi | undefined;
+
+export const getVSCodeApi = (): VSCodeApi => {
+  if (!vscode) {
+    vscode = acquireVsCodeApi();
+  }
```
addaleax:
```suggestion
  vscode ??= acquireVsCodeApi();
```

package.json
Anemy: +1 Isn't the `**/**` here redundant? I think your comment was for another comment.

src/utils/webviewHelpers.ts
```diff
@@ -0,0 +1,114 @@
...
+}: WebviewHtmlOptions): string => {
+  const nonce = getNonce();
+  const scriptUri = getWebviewUri(
+    extensionPath,
+    webview,
+    'dist',
+    'webviewApp.js',
+  );
+  // Add cache-busting query parameter
+  const scriptUriWithCacheBust = `${scriptUri.toString()}?v=${Date.now()}`;
```
Anemy: Can we leave a comment for why we do this cache burst?

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -0,0 +1,98 @@
...
+});
+
+const PreviewApp: React.FC = () => {
+  const [documents, setDocuments] = useState<PreviewDocument[]>([]);
+  const darkMode = useDarkMode();
+  const [isLoading, setIsLoading] = useState(true);
+
+  // Track when loading started for minimum loading duration
+  const loadingStartTimeRef = useRef<number>(Date.now());
+  const MIN_LOADING_DURATION_MS = 500;
```
Anemy: nit: Can we define constants outside of the react function?


```suggestion
const MIN_LOADING_DURATION_MS = 500;

const PreviewApp: React.FC = () => {
  const [documents, setDocuments] = useState<PreviewDocument[]>([]);
  const darkMode = useDarkMode();
  const [isLoading, setIsLoading] = useState(true);

  // Track when loading started for minimum loading duration
  const loadingStartTimeRef = useRef<number>(Date.now());
```

src/views/data-browsing-app/extension-app-message-constants.ts
```diff
@@ -0,0 +1,32 @@
...
+  getDocuments: 'GET_DOCUMENTS',
+
+  // Messages from extension to webview
+  loadDocuments: 'LOAD_DOCUMENTS',
+} as const;
+
+export type PreviewMessageType =
+  (typeof PreviewMessageType)[keyof typeof PreviewMessageType];
+
+export type SortOption = 'default' | 'asc' | 'desc';
```
Anemy: We aren't going to add the sorting until a much later milestone, can we remove it altogether incase we don't get to it?

src/test/suite/views/data-browsing-app/preview-page.test.tsx
Anemy: I don't think I understand your comment, where does the user come in in these tests?
Can we remove the set timeouts altogether here?

src/explorer/documentPreviewItem.ts
```diff
@@ -0,0 +1,165 @@
...
+      } = {
+        limit: options?.limit ?? this._maxDocumentsToShow,
+      };
+
+      // Add sort if specified (not 'default')
+      if (options?.sort === 'asc') {
+        findOptions.sort = { _id: 1 };
+      } else if (options?.sort === 'desc') {
+        findOptions.sort = { _id: -1 };
+      }
```
Anemy: We aren't using this yet, can we remove it?

src/explorer/documentPreviewItem.ts
```diff
@@ -0,0 +1,165 @@
...
+      this.type === CollectionType.timeseries
+    ) {
+      return 0;
+    }
+
+    try {
+      // Pass abortSignal for cancellation support via executionOptions.
+      const executionOptions = signal ? { abortSignal: signal } : undefined;
+
+      const count = await this._dataService.estimatedCount(
```
Anemy: The estimated count here is an estimation. In this view we'd like to show the actual count. Can we do a $count aggregation similar to what Compass' Documents tab does for the count? https://github.com/mongodb-js/compass/blob/ebfe6c99f46d293bcdab089615f8a46b3fd13efe/packages/compass-crud/src/utils/cancellable-queries.ts#L48 

With a `maxTimeMS` of 5000 ms as well so that we aren't waiting too long for this count.

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  useEffect(() => {
+    const handleMessage = (event: MessageEvent): void => {
+      const message: MessageFromExtensionToWebview = event.data;
+      if (message.command === PreviewMessageType.loadDocuments) {
+        const elapsed = Date.now() - loadingStartTimeRef.current;
+        const remainingTime = Math.max(0, MIN_LOADING_DURATION_MS - elapsed);
+
+        // Ensure minimum loading duration before hiding loader
```
Anemy: This is slowing things down when we don't need to slow it down. Can we remove this functionality? Showing a loader is fine while documents are still loading but we shouldn't have to wait an arbitrary amount of time.

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -0,0 +1,98 @@
...
+      }}
+    >
+      <div style={{ padding: '16px' }}>
+        {/* Inject keyframes for spinner animation */}
+        <style>{spinnerKeyframes}</style>
+
+        {isLoading ? (
+          <div className={loadingOverlayStyles}>
+            <span className={spinnerStyles}>
+              <Icon glyph="Refresh" size="large" />
```
Anemy: Like the previous comment for a loader, let's use `VscodeProgressRing` here? I know this is subject to change soon, so not a blocker.

src/utils/webviewHelpers.ts
tculig: I found I wasn't getting the fresh UI when testing in development. Perhaps this shouldn't be committed to main...removing

src/views/data-browsing-app/preview-page.tsx
tculig: updating

src/views/data-browsing-app/extension-app-message-constants.ts
tculig: oh really, I was thinking its a "must have"....removing

src/test/suite/views/data-browsing-app/preview-page.test.tsx
tculig: oh, I'm sorry, I didn't realize this is a test file!!! 🤦‍♂️

src/views/data-browsing-app/preview-page.tsx
tculig: for local development, the load is practically instantaneous, so all the screen does is flickers, and then I'm not sure if it refreshed or not. The idea was to show the loader state so that the user is certain that the reload actually happened. I will remove it, as per your request.

src/explorer/documentPreviewItem.ts
tculig: I carried this logic over from the CollectionTreeItem. 
image

src/views/data-browsing-app/preview-page.tsx
Anemy: Was that change not committed? I'm still seeing style={{...

src/views/data-browsing-app/preview-page.tsx
tculig: updated

tculig: updated

src/views/data-browsing-app/preview-page.tsx
tculig: updated

src/explorer/documentPreviewItem.ts
tculig: updated

src/explorer/documentPreviewItem.ts
tculig: removed

src/views/data-browsing-app/extension-app-message-constants.ts
tculig: removed

src/utils/webviewHelpers.ts
tculig: removed

package.json
tculig: it is, yes...

tculig: @gagik Only the webview stuff is in that folder.

src/views/data-browsing-app/preview-page.tsx
tculig: yeah, I missed this one. Updating to use only vscode components for now

src/test/suite/mdbExtensionController.test.ts
```diff
@@ -490,6 +495,7 @@ suite('MDBExtensionController Test Suite', function () {
       await testTreeItem.onDidExpand();
 
       const collectionChildren = await testTreeItem.getChildren();
+      // With enhanced data browsing disabled, Documents is at index 0
```
Anemy:
```suggestion
```

src/views/dataBrowsingController.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  handleGetDocuments = async (
+    panel: vscode.WebviewPanel,
+    options: DataBrowsingOptions,
+  ): Promise<void> => {
+    const abortController = this._createAbortController(panel);
+    const { signal } = abortController;
+
+    try {
+      const [documents, totalCount] = await Promise.all([
```
Anemy: Something we chatted about offline, in the next ticket, VSCODE-728, we'd like these two to be independent, so the user isn't waiting for both to finish when they are only interested in the results of one request.

Similarly then we separate out possible error responses for each.

src/views/dataBrowsingController.ts
```diff
@@ -0,0 +1,242 @@
...
+      return 0;
+    }
+
+    const stages = [{ $match: {} }, { $count: 'count' }];
+    const executionOptions = signal ? { abortSignal: signal } : undefined;
+
+    const result = await dataService.aggregate(
+      namespace,
+      stages,
+      {},
```
Anemy: Can we add a maxTimeMS of 5000 ms here?

src/views/dataBrowsingController.ts
```diff
@@ -0,0 +1,242 @@
...
+        command: PreviewMessageType.loadDocuments,
+        documents,
+        totalCount,
+      });
+    } catch (error) {
+      // Don't report errors for aborted requests.
+      if (signal.aborted) {
+        return;
+      }
+      log.error('Error getting documents', error);
```
Anemy, line 132: Will we surface the error from count or find to users in VSCODE-728?

tculig: yes, noted

src/test/suite/mdbExtensionController.test.ts
tculig: removed

src/views/dataBrowsingController.ts
tculig: removed the whole function until we actually need it

PR #1218, opened 2026-01-09
chore: bump node and align versions for typescript

Anemy:
Should fix draft release not recognizing .ts

https://nodejs.org/en/learn/typescript/run-natively
> If you are using v22.18.0 or later and your source code contains only [erasable TypeScript syntax](https://devblogs.microsoft.com/typescript/announcing-typescript-5-8-beta/#the---erasablesyntaxonly-option), you can execute TypeScript code without any flags.


package.json
```diff
@@ -74,7 +74,7 @@
   },
   "engines": {
     "vscode": "^1.101.1",
-    "node": "^20.19.0 || ^22.12.0 || >= 23.0.0",
+    "node": "^22.20.0 || >= 23.0.0",
```
himanshusinghs, line 77: If my memory serves right, the original change was a requirement coming out of one of the transitive dependency working strictly with the earlier mentioned values. With the updated value we're missing the guard against version below 20.19.0.

Alternative is to keep the old and update the v22 to the suggested version here.

gagik: > With the updated value we're missing the guard against version below 20.19.0.

I _think_ this allows 22.20.x or >=v23 so anything below 22.20 isn't considered compatible

gagik: would good to just check when was the last time a VSCode version had Node <22.20

Anemy: Should things be failing in the tests then? Or is it a harder to see failure? It was updated last in https://github.com/mongodb-js/vscode/pull/1115
I'm going to merge this, but will revert if something comes up.
Do you remember if it was something with the mongosh testing by chance? https://github.com/mongodb-js/mongosh/blob/991df8eaab68c96bcd30b5473be890013d8695dd/package.json#L97

himanshusinghs: Just following up on this. The semver will ignore any version from v20, including the valid versions from that range as well.

I will check when the used Node version was v20 in VSCode.

himanshusinghs: And the problematic tests were the MCP related written and manual tests.

PR #1221, opened 2026-01-14
feat: add document format, use shell by default instead of ejson VSCODE-523

Anemy:
VSCODE-523

Also fixes [VSCODE-347](https://jira.mongodb.org/browse/VSCODE-347) 

## Description

This PR adds a new setting for how the extension will display documents and playground results. The setting, `documentViewAndEditFormat`, has the values `shell` and `ejson`. By default `shell` syntax is used. This is a major change for the extension, it alters how folks interact with their data.

Shell syntax is how documents and data is displayed by the MongoDB shell, [mongosh](https://www.mongodb.com/docs/mongodb-shell/), and is easier for folks to work with than [ejson](https://www.mongodb.com/docs/mongodb-shell/reference/ejson/). 

Screenshot 2026-01-13 at 19 31 51

Screenshot 2026-01-13 at 19 15 37

## Types of changes)

### DIFF
--- a/syntaxes/mongodb-document.tmLanguage.json
+++ b/syntaxes/mongodb-document.tmLanguage.json
@@ -3,7 +3,6 @@
   "scopeName": "source.mongodb-document",
   "fileTypes": ["mongodb"],
   "patterns": [
-    { "include": "#comments" },
     { "include": "#strings" },
     { "include": "#numbers" },
     { "include": "#bson-types" },
@@ -11,39 +10,26 @@
     { "include": "#operators" }
   ],
   "repository": {
-    "comments": {
-      "patterns": [
-        {
-          "name": "comment.line.double-slash.mongodb-document",
-          "match": "//.*$"
-        },
-        {
-          "name": "comment.block.mongodb-document",
-          "begin": "/\\*",
-          "end": "\\*/"
-        }
-      ]
-    },
     "strings": {
       "patterns": [
         {
-          "name": "string.quoted.double.mongodb-document",
+          "name": "string.quoted.double.js",
           "begin": "\"",
           "end": "\"",
           "patterns": [
             {
-              "name": "constant.character.escape.mongodb-document",
+              "name": "constant.character.escape.js",
               "match": "\\\\."
             }
           ]
         },
         {
-          "name": "string.quoted.single.mongodb-document",
+          "name": "string.quoted.single.js",
           "begin": "'",
           "end": "'",
           "patterns": [
             {
-              "name": "constant.character.escape.mongodb-document",
+              "name": "constant.character.escape.js",
               "match": "\\\\."
             }
           ]
@@ -53,7 +39,7 @@
     "numbers": {
       "patterns": [
         {
-          "name": "constant.numeric.mongodb-document",
+          "name": "constant.numeric.js",
           "match": "\\b\\d+(\\.\\d+)?([eE][+-]?\\d+)?\\b"
         }
       ]
@@ -61,7 +47,7 @@
     "bson-types": {
       "patterns": [
         {
-          "name": "support.class.bson.mongodb-document",
+          "name": "support.class.js",
           "match": "\\b(ObjectId|ISODate|Double|Long|NumberLong|Int32|NumberInt|Decimal128|NumberDecimal|Binary|BinData|UUID|Code|Timestamp|MinKey|MaxKey|DBRef|BSONRegExp|RegExp)\\b"
         }
       ]
@@ -69,7 +55,7 @@
     "keywords": {
       "patterns": [
         {
-          "name": "constant.language.mongodb-document",
+          "name": "constant.language.js",
           "match": "\\b(true|false|null|undefined|NaN|Infinity)\\b"
         }
       ]
@@ -77,7 +63,7 @@
     "operators": {
       "patterns": [
         {
-          "name": "keyword.operator.mongodb-document",
+          "name": "keyword.operator.js",
           "match": "[+\\-*/%=<>!&|]"
         }
       ]


snippets/mongodb-document.json
Anemy, line 1: Let's see if we can align/share these with mongodb-constants, the ts-autocomplete, or the driver.

src/editors/editorsController.ts
```diff
@@ -355,15 +448,24 @@ export default class EditorsController {
...
+    fileUri,
+    document,
+  }: {
+    format: DocumentViewAndEditFormat;
+    fileUri: vscode.Uri;
+    document: Document;
+  }): void {
+    const content =
+      format === 'ejson'
+        ? JSON.stringify(document, null, 2)
```
tculig:
```suggestion
        ? EJSON.stringify(document, undefined, 2, { relaxed: false })
```

src/editors/collectionDocumentsProvider.ts
```diff
@@ -106,10 +113,15 @@ export default class CollectionViewProvider
...
+        format: editFormat,
         uri,
       });
 
-      return JSON.stringify(documents, null, 2);
+      if (editFormat === 'shell') {
+        return toJSString(documents, 2) ?? '';
+      }
+
+      return EJSON.stringify(documents, undefined, 2, { relaxed: false });
```
Anemy, line 127: Previously this was JSON.stringify, which was a pretty bad bug that didn't go noticed for a loooong time.

src/explorer/documentTreeItem.ts
```diff
@@ -81,13 +81,13 @@ export default class DocumentTreeItem
   async getStringifiedEJSONDocumentContents(): Promise<string> {
     const document = await this.getDocumentContents();
 
-    return EJSON.stringify(document, undefined, 2);
+    return EJSON.stringify(document, undefined, 2, { relaxed: false });
```
Anemy, line 87: This change will make us follow a strict formatter in the EJSON and shell output. Fixes issues like [VSCODE-347](https://jira.mongodb.org/browse/VSCODE-347)

src/editors/editorsController.ts
Anemy: The `document` here is already a `serialize`d EJSON javascript object:
https://github.com/mongodb/js-bson?tab=readme-ov-file#ejsonserializebson-options
From here: https://github.com/mongodb-js/vscode/pull/1221/changes#diff-60722e290d2600645706cc3faf8a9bc90b3da1d6435129cff25d22fe62f96eefR149 
I'm thinking we remove that line ^ and then we do the update you have suggested here. Then the function won't expect an already serialized EJSON js object for this case, and we can simplify things. I'll push a commit tomorrow for that. Nice

Anemy: Looking into this more, I'm thinking it would be best follow up on this one with a refactor in a future up. It already has unit tests around it and the $uuid implementation. The JSON.stringify of the serialized EJSON is what we're already doing here. How does that sound? 

We have the [serializing of the EJSON before the JSON stringify in order to reformat binary data of type 4 to $uuid](https://github.com/mongodb-js/vscode/blob/1bb9cab0e405e4c7d7afe34cd5afc35c4daccfad/src/utils/ejson.ts#L20). We might be able to use a `replacer` in the EJSON stringify to simplify that. I'll create a ticket so we don't miss it.

snippets/mongodb-document.json
```diff
@@ -0,0 +1,72 @@
...
+  "NumberInt": {
+    "prefix": ["NumberInt", "Int"],
+    "body": "NumberInt('$1')",
+    "description": "32-bit integer"
+  },
+  "NumberDecimal": {
+    "prefix": ["NumberDecimal", "Decimal"],
+    "body": "NumberDecimal('$1')",
+    "description": "128-bit decimal"
+  },
```
addaleax, line 46: This is what the server/legacy shell uses, but we've kind of standardized on the driver names in mongosh (Long, Int32, Decimal128)

snippets/mongodb-document.json
```diff
@@ -0,0 +1,72 @@
...
+  "MaxKey": {
+    "prefix": "MaxKey",
+    "body": "MaxKey()",
+    "description": "BSON MaxKey"
+  },
+  "DBRef": {
+    "prefix": "DBRef",
+    "body": "DBRef('$1', $2)",
+    "description": "BSON DBRef"
```
addaleax, line 80: DBRef has optional third and fourth arguments

snippets/mongodb-document.json
```diff
@@ -0,0 +1,72 @@
...
+  },
+  "Code": {
+    "prefix": "Code",
+    "body": "Code('$1')",
+    "description": "BSON Code"
+  },
+  "Regex": {
+    "prefix": "Regex",
+    "body": "Regex('$1', '$2')",
```
addaleax: Not how we represent this type of value in JS – it's `BSONRegExp` if we want generic representation, otherwise JS syntax frequently also works

syntaxes/mongodb-document.tmLanguage.json
```diff
@@ -0,0 +1,86 @@
...
+          "name": "constant.numeric.mongodb-document",
+          "match": "\\b\\d+(\\.\\d+)?([eE][+-]?\\d+)?\\b"
+        }
+      ]
+    },
+    "bson-types": {
+      "patterns": [
+        {
+          "name": "support.class.bson.mongodb-document",
+          "match": "\\b(ObjectId|ISODate|NumberLong|NumberInt|NumberDecimal|Binary|BinData|UUID|Code|Timestamp|MinKey|MaxKey|DBRef|Regex)\\b"
```
addaleax: ditto re: Regex, that's kinda not a thing for us

src/utils/ejson.ts
```diff
@@ -40,6 +40,6 @@ function simplifyEJSON(item: Document[] | Document): Document {
 }
 
 export function getEJSON(item: Document[] | Document): Document {
-  const ejson = EJSON.serialize(item);
+  const ejson = EJSON.serialize(item, { relaxed: false });
```
addaleax, line 43: We've definitely also had conversations about relaxed vs. canonical before, unfortunately there's no single answer re: which is better in most cases because that's going to depend on the use case (processing with another tool -> relaxed, as a format for persisting data -> canonical)

If we do switch here, there's definitely going to be people (understandably) complaining about it

src/editors/editorsController.ts
```diff
@@ -355,15 +448,24 @@ export default class EditorsController {
...
+        : (toJSString(document, 2) ?? '');
+
+    this._memoryFileSystemProvider.writeFile(fileUri, Buffer.from(content), {
```
addaleax:
```suggestion
    this._memoryFileSystemProvider.writeFile(fileUri, new TextEncoder().encode(content), {
```

no strong feelings, but I tend to prefer using APIs that are web standards and not Node.js-bound when it doesn't otherwise make a difference which one to use

snippets/mongodb-document.json
addaleax: Where is this file being used right now? :)

src/test/suite/editors/mongoDBDocumentService.test.ts
```diff
@@ -16,10 +16,16 @@ import { ExtensionContextStub } from '../stubs';
 
 const expect = chai.expect;
 
+const namespace = 'waffle.house';
+const connectionId = 'tasty_sandwhich';
```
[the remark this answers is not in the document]
Anemy: I'm going to follow the pattern in the repo on this one :)
Screenshot 2026-01-16 at 09 58 54

snippets/mongodb-document.json
Anemy: Nice catch! We could update https://github.com/mongodb-js/devtools-shared/blob/54e709137ea65fe338ec96fa6a86ceefaf8b8075/packages/query-parser/src/stringify.ts#L140 to default `BSONRegExp`. 🤔

src/utils/ejson.ts
Anemy: Thanks for mentioning, this a big switch here. I am leaning towards canonical by default even though it's more verbose. We findOneAndReplace documents in the extension when saving documents, so type information (1.0 becoming an int) is lost otherwise.

We could have this as a setting similar to the print format, that way there's a way for those folks to have the previous experience. I'll create a thread in the slack channel and see if that would make sense for us. If so I'll add it in a follow up PR. In the thread I'll also mention the change and make sure we're all onboard.

src/editors/editorsController.ts
Anemy: Updated, added a test.

snippets/mongodb-document.json
Anemy: Added in the mongosh names.

snippets/mongodb-document.json
Anemy: It's used in the snippets: https://github.com/mongodb-js/vscode/pull/1221/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R245 

So folks get it in autocomplete then:

Image

Anemy: I don't think I'll align it in this pr. Maybe pulling from [shell-bson-parser](https://github.com/mongodb-js/devtools-shared/blob/main/packages/shell-bson-parser/src/scope.ts) or [mongodb-constants](https://github.com/mongodb-js/devtools-shared/blob/main/packages/mongodb-constants/src/bson-types.ts) which we use in Compass' autocomplete would be best.

snippets/mongodb-document.json
Anemy: Do you feel we should have it in this snippet? If the optional arguments oftentimes aren't used it would be nicer to folks to not need to remove them. In Compass we only include two arguments 
Screenshot 2026-01-16 at 11 44 14

src/utils/ejson.ts
Anemy: Thread https://mongodb.slack.com/archives/GTTUKDWRH/p1768589713193009

src/editors/editDocumentCodeLensProvider.ts
```diff
@@ -91,7 +106,10 @@ export default class EditDocumentCodeLensProvider
         // to be able to save the editable document.
         if (item !== null && item._id && namespace) {
           resultCodeLensesInfo.push({
-            documentId: EJSON.deserialize(EJSON.serialize(item._id)),
+            documentId: EJSON.deserialize(
```
lerouxb, line 109: I know this is old code, but Is this to make sure that strings, numbers, actual objectids, everything get normalised to something? Actually I don't see that it would even be normalised in that case. I don't get it :)

src/editors/editorsController.ts
```diff
@@ -201,17 +244,44 @@ export default class EditorsController {
...
       );
-      const fileName = `${VIEW_DOCUMENT_SCHEME}:/${fileTitle}.json`;
+      const fileName =
+        data.format === 'ejson'
+          ? `${VIEW_DOCUMENT_SCHEME}:/${fileTitle}.json`
+          : `${VIEW_DOCUMENT_SCHEME}:/${fileTitle}`;
+
+      // We store the format in the URI query string to later
+      // reference it and ensure we follow the format on the file.
+      const formatUriQuery = `${DOCUMENT_FORMAT_URI_IDENTIFIER}=${data.format}`;
```
lerouxb, line 254: I might be misunderstanding what's going on, but is the query string the best/ideal place to store this kind of information per file? Should we store it somewhere where all these things will persist next time you open the file?

src/editors/editorsController.ts
```diff
@@ -201,17 +244,44 @@ export default class EditorsController {
...
 
       const fileUri = vscode.Uri.parse(fileName, true).with({
-        query: `?${namespaceUriQuery}&${connectionIdUriQuery}&${documentIdUriQuery}&${documentSourceUriQuery}`,
+        query: `${namespaceUriQuery}&${connectionIdUriQuery}&${documentIdUriQuery}&${documentSourceUriQuery}&${formatUriQuery}`,
```
lerouxb, line 257: I know this is old and kinda out of our hands, but it is a pity query is a string. It leaves us open to so many potential bugs in formatting and not escaping values.

src/editors/editorsController.ts
```diff
@@ -75,9 +94,11 @@ export function getViewCollectionDocumentsUri(
   );
 
   // The part of the URI after the scheme and before the query is the file name.
-  return vscode.Uri.parse(
-    `${VIEW_COLLECTION_SCHEME}:Results: ${namespaceDisplayName}.json${uriQuery}`,
-  );
+  const extension = editFormat === 'ejson' ? '.json' : '';
```
lerouxb, line 97: I don't quite follow. What happened to the .mongo file extension?

src/mcp/mcpConnectionErrorHandler.ts
```diff
@@ -15,7 +15,9 @@ export const createMCPConnectionErrorHandler = (
 
     const connectionGuidance: string[] = availableConnections.length
       ? [
-          `Following connections are available in the Sidebar for the user to connect to: ${JSON.stringify(availableConnections)}. Note to LLM: Provide user with the list of connections available and ask them to select a connection in the Sidebar.`,
+          `Following connections are available in the Sidebar for the user to connect to: ${JSON.stringify(
```
lerouxb: I just spent 3 minutes trying to spot what changed 😆 

(a trailing comma)

src/mcp/mcpConnectionErrorHandler.ts
```diff
@@ -51,7 +53,11 @@ export const createMCPConnectionErrorHandler = (
             content: [
               {
                 type: 'text',
-                text: `MCP server is having trouble connecting to ${activeConnectionName ? activeConnectionName : 'the selected connection in the MongoDB VSCode extension'}.`,
+                text: `MCP server is having trouble connecting to ${
```
lerouxb: Do we have prettier and some hooks to ensure it runs on commit set up in this repo?

(I ask because I see a lot of lines changed that didn't actually change except for formatting.)

src/participant/sampleDocuments.ts
```diff
@@ -57,7 +58,7 @@ export async function getStringifiedSampleDocuments({
     additionToPrompt = sampleDocuments[0];
   }
 
-  const stringifiedDocuments = toJSString(additionToPrompt);
+  const stringifiedDocuments = toJSString(additionToPrompt) ?? '';
```
lerouxb, line 61: Not for this PR, but we really should just fix toJSString() to always return a string like it says it does. We have this `?? ''` all over the place.

syntaxes/mongodb-document.tmLanguage.json
```diff
@@ -0,0 +1,86 @@
+{
+  "name": "MongoDB Document",
```
lerouxb, line 2: Why do we need this format? Is it not just a subset of a playground? And since it is readonly we don't have to worry about anyone sneaking in non-declarative parts of the language.

src/editors/editorsController.ts
```diff
@@ -201,17 +244,44 @@ export default class EditorsController {
...
+      });
 
       const document = await vscode.workspace.openTextDocument(fileUri);
 
       await vscode.window.showTextDocument(document, { preview: false });
+      // Setting the document language triggers a document close/reopen
+      // for the no-extension format (shell syntax).
+      await vscode.languages.setTextDocumentLanguage(
+        document,
+        data.format === 'ejson' ? 'json' : 'mongodb-document',
```
lerouxb, line 278: I mentioned this elsewhere too, but what's wrong with just using the same format we have for playgrounds? My understanding is that any shell syntax formated document would be a valid playground already.

src/editors/editorsController.ts
Anemy: It would be nicer for us to use a reference id here that then maps to something in our memory. I don't think we should try to make that change in these changes.

src/mcp/mcpConnectionErrorHandler.ts
Anemy: I think these might be from an older prettier config I had set up and somehow made its way into this pr. I had this based off a POC branch from a year+ ago that I flattened. I'll remove these to keep the diff and history.

src/editors/editorsController.ts
Anemy: We don't persist open documents . It's a bit of a bad UX at the moment, it renders them like:
Screenshot 2026-01-21 at 08 31 54
We should auto close them, or give folks a chance to connect to the cluster somehow. I'm thinking it could be worth a ticket in the epic.

The query string is the best place at the moment unless we store everything in some state that has a reference id, the reference id would be in the query string. We can't use the file name as is as a source of truth as multiple connections may have the same document id in the database and collection so the query string is where to we can pass this information. 

I think both of those can be improved, I don't think this pr is the place for that though.

src/editors/editorsController.ts
Anemy: .mongodb and .mongodb.js are for playgrounds. We're going with our own custom file scheme with no extension for shell syntax in these changes. Having them in .js could be an alternative. I think we want to avoid a new extension like .mdb or .mongo.

src/participant/sampleDocuments.ts
Anemy: Agreed. We could clean up [toJSString](https://github.com/mongodb-js/devtools-shared/blob/81c2cb856d44c84b465137409c9b9655ce452fde/packages/query-parser/src/stringify.ts#L153) a lot.
A few other things to mention:
- We should have it use BSONRegExp instead of RegExp()
- We should have it accept a replacer, like JSON.stringify does.

syntaxes/mongodb-document.tmLanguage.json
Anemy: Playgrounds have run actions associated with them. I'm thinking having a defined format gives us a bit more control. Especially down the line if we want to add a code lens like `clone` in a code lens.

src/editors/editorsController.ts
Anemy: Let's keep this in one thread: https://github.com/mongodb-js/vscode/pull/1221#discussion_r2712403618

src/editors/editDocumentCodeLensProvider.ts
Anemy: For what it's worth, these code lenses will no longer be used once we've moved to the new view. I agree it's doing some unexpected and unnecessary things. Let's handle it in another pr as it is existing code and not related to these changes.

src/participant/sampleDocuments.ts
lerouxb: we should file tickets somewhere :)

PR #1222, opened 2026-01-15
chore: add bloated-documents playground

Anemy:
**Edit: Had this branch laying around and didn't notice it could use a cleanup before opening a pr. I'll open one up for the bloated documents case eventually.

Useful for testing how the extension and other tools handle very large bloated documents.

### DIFF
--- a/playgrounds/bloated-documents.mongodb
+++ b/playgrounds/bloated-documents.mongodb
@@ -2,6 +2,8 @@
 
 use("test");
 
+const alphabet = 'abcdefghijklmnopqrstuvwxyz';
+
 function largeString(length = 1000) {
   let result = '';
   for (let i = 0; i < length; i++) {
@@ -9,8 +11,6 @@ function largeString(length = 1000) {
   }
   return result;
 }
-
-const alphabet = 'abcdefghijklmnopqrstuvwxyz';
 
 function createInnerDoc() {
   const array = [];


playgrounds/bloated-documents.mongodb
```diff
@@ -0,0 +1,56 @@
+// MongoDB Playground
+// Use Ctrl+Space inside a snippet or a string literal to trigger completions.
+
+// The current database to use.
```
Anemy:
```suggestion
// Playground for creating documents with very large strings and many fields. 

```

playgrounds/bloated-documents.mongodb
```diff
@@ -0,0 +1,54 @@
...
+
+function largeString(length = 1000) {
+  let result = '';
+  for (let i = 0; i < length; i++) {
+    result += alphabet[Math.floor(Math.random() * alphabet.length)];
+  }
+  return result;
+}
+
+const alphabet = 'abcdefghijklmnopqrstuvwxyz';
```
Anemy:
```suggestion

const alphabet = 'abcdefghijklmnopqrstuvwxyz';

function largeString(length = 1000) {
  let result = '';
  for (let i = 0; i < length; i++) {
    result += alphabet[Math.floor(Math.random() * alphabet.length)];
  }
  return result;
}
```

PR #1225, opened 2026-01-15
chore(ci): set pnpm version in publish-release

Anemy:


.github/workflows/publish-release.yaml
```diff
@@ -14,6 +14,9 @@ jobs:
     steps:
       - name: Setup pnpm
         uses: pnpm/action-setup@v4
+        with:
+          version: 10.23.0 # Aligned with "packageManager" in package.json.
```
gagik, line 18: it's weird that it's not picked up

Anemy: This action doesn't run `checkout` like the others do.

Anemy: Part of me thinks we should checkout, at least to have a consistent vsce version here and to ensure this pnpm version doesn't start being out of date.

PR #1229, opened 2026-01-20
feat: add document list and actions data browser VSCODE-728

tculig:
## Description

https://jira.mongodb.org/browse/VSCODE-728

Initial load
image

Refresh
image

Page 2
image

No documents page
image

Errors
image

More than half the lines of code in this PR are unit tests.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/test/suite/views/data-browsing-app/actions.test.tsx
+++ b/src/test/suite/views/data-browsing-app/actions.test.tsx
@@ -232,4 +232,3 @@ describe('actions test suite', function () {
     });
   });
 });
-
--- a/src/test/suite/views/data-browsing-app/messageHandler.test.tsx
+++ b/src/test/suite/views/data-browsing-app/messageHandler.test.tsx
@@ -67,9 +67,9 @@ describe('messageHandler test suite', function () {
           documents: [],
         });
 
-        expect(
-          store.getState().documentQuery.displayedDocuments,
-        ).to.deep.equal([]);
+        expect(store.getState().documentQuery.displayedDocuments).to.deep.equal(
+          [],
+        );
       });
 
       it('should handle undefined documents', function () {
@@ -78,9 +78,9 @@ describe('messageHandler test suite', function () {
           isInitialLoad: true,
         } as any);
 
-        expect(
-          store.getState().documentQuery.displayedDocuments,
-        ).to.deep.equal([]);
+        expect(store.getState().documentQuery.displayedDocuments).to.deep.equal(
+          [],
+        );
       });
 
       it('should clear getDocuments error on successful load', function () {
@@ -157,9 +157,9 @@ describe('messageHandler test suite', function () {
           totalCount: 100,
         });
 
-        expect(
-          store.getState().documentQuery.totalCountInCollection,
-        ).to.equal(100);
+        expect(store.getState().documentQuery.totalCountInCollection).to.equal(
+          100,
+        );
         expect(store.getState().documentQuery.hasReceivedCount).to.be.true;
       });
 
@@ -169,9 +169,9 @@ describe('messageHandler test suite', function () {
           totalCount: 0,
         });
 
-        expect(
-          store.getState().documentQuery.totalCountInCollection,
-        ).to.equal(0);
+        expect(store.getState().documentQuery.totalCountInCollection).to.equal(
+          0,
+        );
       });
 
       it('should clear getTotalCount error on successful count update', function () {
@@ -201,9 +201,8 @@ describe('messageHandler test suite', function () {
         });
 
         expect(store.getState().documentQuery.hasReceivedCount).to.be.true;
-        expect(
-          store.getState().documentQuery.totalCountInCollection,
-        ).to.be.null;
+        expect(store.getState().documentQuery.totalCountInCollection).to.be
+          .null;
       });
 
       it('should set getTotalCount error with provided message', function () {
@@ -254,4 +253,3 @@ describe('messageHandler test suite', function () {
     });
   });
 });
-
--- a/src/test/suite/views/dataBrowsingController.test.ts
+++ b/src/test/suite/views/dataBrowsingController.test.ts
@@ -205,8 +205,18 @@ suite('DataBrowsingController Test Suite', function () {
       });
 
       // Create controllers for both panels
-      const request1 = testController.handleGetDocuments(panel1, options, 0, 10);
-      const request2 = testController.handleGetDocuments(panel2, options, 0, 10);
+      const request1 = testController.handleGetDocuments(
+        panel1,
+        options,
+        0,
+        10,
+      );
+      const request2 = testController.handleGetDocuments(
+        panel2,
+        options,
+        0,
+        10,
+      );
 
       const controller1 =
         testController._panelAbortControllers.get(panel1)?.documents;
@@ -229,7 +239,9 @@ suite('DataBrowsingController Test Suite', function () {
     test('does not post message when request is aborted', async function () {
       // Abort the controller during the find request
       mockDataService.find = sandbox.stub().callsFake(() => {
-        testController._panelAbortControllers.get(mockPanel)?.documents?.abort();
+        testController._panelAbortControllers
+          .get(mockPanel)
+          ?.documents?.abort();
         return [{ _id: '1', name: 'test' }];
       });
       mockDataService.aggregate = sandbox.stub().resolves([{ count: 16 }]);
@@ -245,7 +257,9 @@ suite('DataBrowsingController Test Suite', function () {
     test('does not throw when error occurs on aborted request', async function () {
       // Abort the controller and throw an error during find
       mockDataService.find = sandbox.stub().callsFake(() => {
-        testController._panelAbortControllers.get(mockPanel)?.documents?.abort();
+        testController._panelAbortControllers
+          .get(mockPanel)
+          ?.documents?.abort();
         throw new Error('Connection error');
       });
       mockDataService.aggregate = sandbox.stub().resolves([{ count: 16 }]);
@@ -449,7 +463,9 @@ suite('DataBrowsingController Test Suite', function () {
 
     test('does not post message when request is aborted', async function () {
       mockDataService.find = sandbox.stub().callsFake(() => {
-        testController._panelAbortControllers.get(mockPanel)?.documents?.abort();
+        testController._panelAbortControllers
+          .get(mockPanel)
+          ?.documents?.abort();
         return [{ _id: '1', name: 'test' }];
       });
       const options = createMockOptions();
@@ -632,7 +648,9 @@ suite('DataBrowsingController Test Suite', function () {
 
       expect(postMessageStub.calledOnce).to.be.true;
       const message = postMessageStub.firstCall.args[0];
-      expect(message.command).to.equal(PreviewMessageType.updateTotalCountError);
+      expect(message.command).to.equal(
+        PreviewMessageType.updateTotalCountError,
+      );
       expect(message.error).to.equal('Aggregate failed');
     });
 
@@ -645,13 +663,17 @@ suite('DataBrowsingController Test Suite', function () {
       await testController.handleGetTotalCount(mockPanel, options);
 
       const message = postMessageStub.firstCall.args[0];
-      expect(message.command).to.equal(PreviewMessageType.updateTotalCountError);
+      expect(message.command).to.equal(
+        PreviewMessageType.updateTotalCountError,
+      );
       expect(message.error).to.equal('No active database connection');
     });
 
     test('does not post message when request is aborted', async function () {
       mockDataService.aggregate = sandbox.stub().callsFake(() => {
-        testController._panelAbortControllers.get(mockPanel)?.totalCount?.abort();
+        testController._panelAbortControllers
+          .get(mockPanel)
+          ?.totalCount?.abort();
         return [{ count: 16 }];
       });
       const options = createMockOptions();
--- a/src/views/data-browsing-app/store/messageHandler.ts
+++ b/src/views/data-browsing-app/store/messageHandler.ts
@@ -1,9 +1,5 @@
 import { store } from './index';
-import type {
-  MessageFromExtensionToWebview,
-  DocumentGetErrorMessage,
-  UpdateTotalCountErrorMessage,
-} from '../extension-app-message-constants';
+import type { MessageFromExtensionToWebview } from '../extension-app-message-constants';
 import { PreviewMessageType } from '../extension-app-message-constants';
 import {
   loadPage,
@@ -22,9 +18,7 @@ export const handleExtensionMessage = (
       store.dispatch(loadPage((message.documents as PreviewDocument[]) || []));
       break;
     case PreviewMessageType.getDocumentError: {
-      const errorMessage =
-        (message as DocumentGetErrorMessage).error ||
-        'Failed to fetch documents';
+      const errorMessage = message.error || 'Failed to fetch documents';
       store.dispatch(
         setRequestError({ type: 'getDocuments', message: errorMessage }),
       );
@@ -37,9 +31,7 @@ export const handleExtensionMessage = (
       store.dispatch(setTotalCountInCollection(message.totalCount));
       break;
     case PreviewMessageType.updateTotalCountError: {
-      const errorMessage =
-        (message as UpdateTotalCountErrorMessage).error ||
-        'Failed to fetch total count';
+      const errorMessage = message.error || 'Failed to fetch total count';
       store.dispatch(markCountReceived());
       store.dispatch(
         setRequestError({ type: 'getTotalCount', message: errorMessage }),
@@ -61,4 +53,3 @@ export const setupMessageHandler = (): (() => void) => {
     window.removeEventListener('message', handleMessage);
   };
 };
-


src/views/dataBrowsingController.ts
```diff
@@ -177,6 +300,36 @@ export default class DataBrowsingController {
...
+    ) {
+      return 0;
+    }
+
+    const dataService = this._connectionController.getActiveDataService();
+    if (!dataService) {
+      return 0;
+    }
+
+    const stages = [{ $match: {} }, { $count: 'count' }];
```
Anemy: We can remove the $match here.
```suggestion
    const stages = [{ $count: 'count' }];
```

src/views/dataBrowsingController.ts
```diff
@@ -177,6 +300,36 @@ export default class DataBrowsingController {
     );
   };
 
+  private async _getTotalCount(
+    namespace: string,
+    collectionType: string,
+    signal?: AbortSignal,
```
Anemy: Will there always be a signal? How about:
```suggestion
    signal: AbortSignal,
```

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -1,53 +1,355 @@
...
+
+const contentStyles = css({
+  padding: spacing[400],
+  flex: 1,
+});
+
+const emptyStateStyles = css({
+  textAlign: 'center',
+  padding: spacing[500],
+  color: 'var(--vscode-descriptionForeground)',
```
Anemy: Can we have the vscode styles defined as constants in a file that we can reference here?  (And in other places we're referencing the VSCode styles). I'm thinking that'll make it easier to keep track of all of the VSCode references, and in time make it quicker to use/reference them.

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -1,53 +1,355 @@
...
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'space-between',
+  padding: `${spacing[200]}px ${spacing[300]}px`,
+  borderBottom: '1px solid var(--vscode-panel-border, #444)',
+  gap: spacing[300],
+  flexWrap: 'wrap',
+  position: 'sticky',
+  top: 0,
+  backgroundColor: 'var(--vscode-editor-background, #1e1e1e)',
```
Anemy: Is the editor background ever not defined? Maybe we can remove the fallbacks.

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -1,53 +1,355 @@
...
+      fetchPageFromServer(newPage, itemsPerPage);
+    }
+  };
+
+  const handleItemsPerPageChange = (event: Event): void => {
+    const target = event.target as HTMLSelectElement;
+    const newItemsPerPage = parseInt(target.value, 10);
+    setItemsPerPage(newItemsPerPage);
+    setCurrentPage(1);
+    fetchPageFromServer(1, newItemsPerPage);
```
Anemy: Having multiple set state calls in many functions I find to be an indicator that we should use a state management system like redux. I think we're at the point where it would reduce complexity even though it'll be some boiler plate code additions. With a state object we're then making one state change in the reducer here. I find it reduces the chance of unforeseen side effects or missed function calls when things are added.

I know we chatted on it in the past, happy to chat sync again on it, I can also sketch up something if it would help. The data-modeling store in Compass is a good place to follow for patterns we've used as a team (without Compass specific parts): https://github.com/mongodb-js/compass/blob/main/packages/compass-data-modeling/src/store/index.ts

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -1,53 +1,355 @@
...
+  width: 'auto',
+  minWidth: 'unset',
+  '--vscode-settings-dropdownBorder': 'transparent',
+  '--vscode-single-select': {
+    padding: '20px',
+  },
+});
+
+const stopButtonStyles = css({
+  '--vscode-button-border': 'transparent',
```
Anemy: While these `--vscode-x` style updates bring the designs more inline with [figma](https://www.figma.com/design/QE0mDJK9hlbPK9shnQCIbG/VS-CODE--DATA-BROWSING?node-id=1675-438516&t=Jefh4vE3448yH58k-4) they are increasing our discrepancies with the native VSCode experience. On some themes they altogether hide the text as the backgrounds are important. Can we remove these overrides? 

*Light+ theme*
Image

| with | without |
| - | - |
| Image | Image |

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -1,53 +1,355 @@
...
+  sendFetchPage,
+  sendCancelRequest,
+} from './vscode-api';
 
 interface PreviewDocument {
   [key: string]: unknown;
 }
 
+const ITEMS_PER_PAGE_OPTIONS = [10, 25, 50, 100];
+export const MIN_LOADING_DURATION_MS = 500;
```
Anemy: Can we remove this minimum wait time? That way things are snappier for folks.

src/views/dataBrowsingController.ts
```diff
@@ -95,24 +108,122 @@ export default class DataBrowsingController {
...
+    panel: vscode.WebviewPanel,
+    options: DataBrowsingOptions,
+  ): Promise<void> => {
+    const abortController = this._createAbortController(panel);
+    const { signal } = abortController;
+
+    try {
+      const [documents, totalCount] = await Promise.all([
+        this._fetchDocuments(options.namespace, options.collectionType, signal),
+        this._getTotalCount(options.namespace, options.collectionType, signal),
```
Anemy: Can we uncouple these two requests? That way they're not waiting on each other or fail if one fails.

src/views/dataBrowsingController.ts
```diff
@@ -95,24 +108,122 @@ export default class DataBrowsingController {
...
+        documents,
+        totalCount,
+      });
+    } catch (error) {
+      // Don't report errors for aborted requests.
+      if (signal.aborted) {
+        return;
+      }
+      log.error('Error refreshing documents', error);
+      void panel.webview.postMessage({
```
Anemy, line 177: Can we instead surface these errors to users in the UI? Folks have complained in the past that the toast popups are both intrusive and can't display the whole error.

src/views/dataBrowsingController.ts
```diff
@@ -121,22 +232,30 @@ export default class DataBrowsingController {
...
+        limit,
       });
     } catch (error) {
       // Don't report errors for aborted requests.
       if (signal.aborted) {
         return;
       }
-      log.error('Error getting documents', error);
+      log.error('Error fetching page', error);
+      void panel.webview.postMessage({
```
Anemy, line 177: Same as above on the error.

src/views/dataBrowsingController.ts
```diff
@@ -177,6 +300,36 @@ export default class DataBrowsingController {
...
+  private async _getTotalCount(
+    namespace: string,
+    collectionType: string,
+    signal?: AbortSignal,
+  ): Promise<number> {
+    if (
+      collectionType === CollectionType.view ||
+      collectionType === CollectionType.timeseries
+    ) {
+      return 0;
```
Anemy: Something a bit buried in the ticket description, VSCODE-728, is that we should show an `N/A` with a tooltip for views and timeseries instead of a 0 which might mislead folks.

src/views/dataBrowsingController.ts
```diff
@@ -95,24 +108,122 @@ export default class DataBrowsingController {
...
     panel: vscode.WebviewPanel,
     options: DataBrowsingOptions,
   ): Promise<void> => {
     const abortController = this._createAbortController(panel);
     const { signal } = abortController;
 
+    try {
+      const [documents, totalCount] = await Promise.all([
+        this._fetchDocuments(options.namespace, options.collectionType, signal),
+        this._getTotalCount(options.namespace, options.collectionType, signal),
```
Anemy: This looks like the same request as handleRefreshDocuments. Can we de-dupe?

src/views/dataBrowsingController.ts
```diff
@@ -8,25 +9,37 @@ import type { TelemetryService } from '../telemetry';
...
 export const getDataBrowsingContent = ({
   extensionPath,
   webview,
   namespace,
+  codiconStylesheetUri,
 }: {
   extensionPath: string;
   webview: vscode.Webview;
   namespace: string;
+  codiconStylesheetUri?: string;
```
Anemy:
```suggestion
  codiconStylesheetUri: string;
```

src/views/dataBrowsingController.ts
```diff
@@ -121,22 +232,30 @@ export default class DataBrowsingController {
...
   };
 
   private async _fetchDocuments(
     namespace: string,
     collectionType: string,
     signal?: AbortSignal,
+    skip?: number,
+    limit?: number,
   ): Promise<Document[]> {
     if (collectionType === CollectionType.view) {
```
Anemy: Didn't notice this in the last pr, why are we returning an empty array for views here?

src/views/data-browsing-app/preview-page.tsx
tculig: sure, why not

src/views/data-browsing-app/preview-page.tsx
tculig: ok

src/views/data-browsing-app/preview-page.tsx
tculig: please, I beg of you, lets leave this in until we finish the project and remove it at the end. I want to be able to verify that the loading state has triggered and be able to test the abort function repeatedly. Testing locally with small data sets causes the screen to just do a quick flicker when I press reload, for example.

src/views/data-browsing-app/preview-page.tsx
tculig: ok, but there's no passing of the data between pages. Using redux would have exactly the same amount of state calls.

src/views/data-browsing-app/preview-page.tsx
tculig: easy to do

src/views/dataBrowsingController.ts
tculig: yup, that's an obvious fail on my part...

src/views/dataBrowsingController.ts
tculig: ok, but do we have a design for that?

src/views/dataBrowsingController.ts
tculig: de-duped

src/views/dataBrowsingController.ts
tculig: removing...

src/views/dataBrowsingController.ts
tculig: deduped

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -1,54 +1,339 @@
...
+      switch (message.command) {
+        case PreviewMessageType.loadDocuments:
+          dispatch(loadDocuments((message.documents as PreviewDocument[]) || []));
+          break;
+        case PreviewMessageType.loadPage:
+          dispatch(loadPage((message.documents as PreviewDocument[]) || []));
+          break;
+        case PreviewMessageType.refreshError:
+          dispatch(stopLoading());
+          // Could show an error message here if needed
```
Anemy: Shouldn't we be setting the error here?

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -1,54 +1,339 @@
...
+  const totalCountInCollection = useAppSelector(selectTotalCountInCollection);
+  const hasReceivedCount = useAppSelector(selectHasReceivedCount);
+  const totalPages = useAppSelector(selectTotalPages);
+  const startItem = useAppSelector(selectStartItem);
+  const endItem = useAppSelector(selectEndItem);
+
+  // Adjust current page if it exceeds total pages
+  useEffect(() => {
+    if (currentPage > totalPages && totalPages > 0) {
+      dispatch(setCurrentPage(totalPages));
```
Anemy: This looks like something that should be handled in the logic when we set the current page or have the total pages count, can it be in the store?

src/views/vscode-styles.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const VSCODE_EDITOR_BACKGROUND = 'var(--vscode-editor-background)';
+export const VSCODE_EDITOR_FOREGROUND = 'var(--vscode-editor-foreground)';
+export const VSCODE_PANEL_BORDER = 'var(--vscode-panel-border)';
+export const VSCODE_SIDEBAR_BACKGROUND = 'var(--vscode-sideBar-background)';
+export const VSCODE_DESCRIPTION_FOREGROUND =
+  'var(--vscode-descriptionForeground)';
+export const VSCODE_BUTTON_BORDER_TRANSPARENT = 'transparent';
+export const VSCODE_BUTTON_SECONDARY_BACKGROUND_TRANSPARENT = 'transparent';
+export const VSCODE_SETTINGS_DROPDOWN_BORDER_TRANSPARENT = 'transparent';
```
Anemy: Do we use these?
```suggestion
```

src/views/dataBrowsingController.ts
```diff
@@ -93,64 +95,125 @@ export default class DataBrowsingController {
...
+          skip,
+          limit,
+        });
+      } else {
+        void panel.webview.postMessage({
+          command: PreviewMessageType.loadDocuments,
+          documents,
+        });
+
+        void this._getTotalCount(
```
Anemy: We're still coupling the count with the fetching of documents here, can we have them both going at once?

src/views/data-browsing-app/store/documentQuerySlice.ts
```diff
@@ -0,0 +1,152 @@
...
+  displayedDocuments: [],
+  currentPage: 1,
+  itemsPerPage: DEFAULT_ITEMS_PER_PAGE,
+  isLoading: true,
+  totalCountInCollection: null,
+  hasReceivedCount: false,
+  error: null,
+};
+
+const documentQuerySlice = createSlice({
```
Anemy, line 42: For more consistency with how we do thing in Compass, I was thinking we'd follow the reducer with thunk patterns and not use the new `createSlice` syntax.
Similar to how we do it in Compass: https://github.com/mongodb-js/compass/blob/main/packages/compass-data-modeling/src/store/index.ts 
That way we can also call the asynchronous actions from the actions rather than have them in the react components. The react component is then simpler as it's only calling one action for each user action, instead of calling a few state updates and dispatches for each action.

Do you prefer the newer syntax?

src/views/data-browsing-app/preview-page.tsx
Anemy: For local testing adding a
```ts
await new Promise(resolve => setTimeout(resolve, 1000));
```
in the fetch documents request will do the same and can be an easy thing to add in/remove while developing.
That being said leaving this in for the time being is fine, thanks for letting me know the value of it, can we leave a comment around why we have it then?

src/views/dataBrowsingController.ts
Anemy: No we dont, we can ping Ighodalo in the channel if we would like some. The `ErrorSummary` from compass-components will do the trick as well: https://github.com/mongodb-js/compass/blob/main/packages/compass-components/src/components/error-warning-summary.tsx 
Unless I'm missing it, I'm not seeing any error kind of thing in the VSCode elements for it https://vscode-elements.github.io/guides/framework-integrations/react/

src/views/data-browsing-app/store/documentQuerySlice.ts
tculig: Typically, I would say lets use the latest syntax, which is createSlice. In this case, there is a second reason to not use thunks, because there is no awaiting or real async handling on the webview side. The webview fires a request to vscode, and then has a listener for the return message. Using thunks would have minimal benefit here. What I will do, is a small refactor of the current slice to centralize the message handler a little better.

src/views/vscode-styles.ts
tculig: removed

package.json
```diff
@@ -1452,8 +1452,10 @@
     "@mongosh/i18n": "^2.16.0",
     "@mongosh/service-provider-node-driver": "^3.17.1",
     "@mongosh/shell-api": "^3.23.0",
+    "@reduxjs/toolkit": "^2.11.2",
```
gagik, line 1455: ah, we are adding redux? 😢 at least it's redux toolkit...

out of curiosity, what is the motivation behind adopting it now? Or is it just so we have logic closer to Compass?

src/views/data-browsing-app/store/documentQuerySlice.ts
gagik: Redux toolkit is something I hope Compass come move to eventually, though I think Compass at this point seems have all the organizational wrapping that the toolkit is doing, so it might not be worth it.

package.json
tculig: https://github.com/mongodb-js/vscode/pull/1229#discussion_r2710147663

src/views/data-browsing-app/preview-page.tsx
gagik: regardless of redux or not, currentPage + itemsPerPage and even maybe isLoading could be all part of one state right? At least they all feel very tied to each other at the moment.

tculig: will address in an upcoming PR

PR #1188, opened 2025-11-24
test: add install tests for vs code extension VSCODE-703

tculig:
### Checklist
- [ ] New tests and/or benchmarks are included
- [ ] Documentation is changed or added
- [ ] I have signed the MongoDB Contributor License Agreement (https://www.mongodb.com/legal/contributor-agreement)

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*
- [ ] Patch (non-breaking change which fixes an issue)
- [ ] Minor (non-breaking change which adds functionality)
- [ ] Major (fix or feature that would cause existing functionality to change)

PR #1235, opened 2026-01-28
chore: add VSIX install tests VSCODE-703

gagik:
This includes some of the refactoring from https://github.com/mongodb-js/vscode/pull/1133 and adds the VSIX test script from https://github.com/mongodb-js/vscode/pull/1188

There's a few CI follows up I want to do:
- make sure PNPM is properly caching
- cache builds between these runs

### DIFF
--- a/.github/workflows/draft-release.yaml
+++ b/.github/workflows/draft-release.yaml
@@ -71,18 +71,9 @@ jobs:
           pnpm version "${NEXT_VERSION}" --no-git-tag-version
           RELEASE_TAG="v${NEXT_VERSION}"
           echo "release_tag=${RELEASE_TAG}" >> "$GITHUB_OUTPUT"
-          echo "RELEASE_TAG=${RELEASE_TAG}" >> "$GITHUB_ENV"
 
-      - name: Validate release tag
-        shell: bash
-        run: |
-          if [ -z "${{ steps.set-version.outputs.release_tag }}" ]; then
-            echo "RELEASE_TAG is not set or is empty"
-            exit 1
-          fi
-
-          if git rev-parse "${{ steps.set-version.outputs.release_tag }}" >/dev/null 2>&1; then
-            echo "Error: Tag ${{ steps.set-version.outputs.release_tag }} already existing"
+          if git rev-parse "${RELEASE_TAG}" >/dev/null 2>&1; then
+            echo "Error: Tag ${RELEASE_TAG} already existing"
             echo "If you are trying to re-create a draft release with this version, please delete the release and the tag first."
             echo "If this version has already been release consider using a different one."
             exit 1
@@ -274,9 +265,10 @@ jobs:
           pnpm run generate-vulnerability-report > /dev/null
 
       - name: Create Draft Release
+        env:
+          RELEASE_TAG: ${{ needs.prepare-version.outputs.release_tag }}
         run: |
           set -e
-          RELEASE_TAG="${{ needs.prepare-version.outputs.release_tag }}"
           echo Creating draft release for: "${RELEASE_TAG}"
           ls *.vsix
           ls *.vsix.sig


.github/workflows/actions/test-and-build/action.yaml
```diff
@@ -85,6 +85,12 @@ runs:
       run: pnpm run check-vsix-size
       shell: bash
 
+    - name: Run VSIX Install Test
```
gagik: should we run this on forks? I was thinking this has some potential security risks because of this installation being a rather odd thing.

himanshusinghs: Not sure of this thread is new or old but I guess after Slack discussion, you'd wanna add the install tests for forked PRs? Also fine if you wish to do that in a follow up PR.

himanshusinghs: Perhaps one thing to note - this action is now only used in draft-release workflow.

.github/workflows/test-and-build.yaml
```diff
@@ -31,36 +82,110 @@ jobs:
...
+
+      - name: Setup Node environment
+        uses: ./.github/workflows/actions/setup-node-environment
+
+      - name: Download .vsix
+        uses: actions/download-artifact@v4
+        with:
+          name: vsix
+
+      - name: Sign .vsix
```
himanshusinghs: Small nit: One thing I just now realised that the "Sign .vsix" is relevant only at the time of releasing (in draft-release workflow) so if you wish to further simplify, very likely we could skip the VSIX signing and uploading here.

package.json
```diff
@@ -55,6 +55,7 @@
     "test": "pnpm run test-webview && pnpm run test-extension",
     "test-extension": "cross-env MDB_IS_TEST=true NODE_OPTIONS=--no-force-async-hooks-checks xvfb-maybe node ./out/test/runTest.js",
     "test-webview": "mocha -r ts-node/register --exit --grep=\"${MOCHA_GREP}\" --file ./src/test/setup-webview.ts src/test/suite/views/**/*.test.tsx",
+    "test-install": "bash ./scripts/test-vsix-install.sh",
```
himanshusinghs, line 58: I am not sure if using bash on windows runner would work without additional setup.

gagik: [it does](https://github.com/mongodb-js/vscode/actions/runs/21448845771/job/61772408355?pr=1235)

.github/workflows/test-and-build.yaml
gagik: yeah fair

.github/workflows/draft-release.yaml
```diff
@@ -74,39 +69,214 @@ jobs:
...
-          echo "RELEASE_TAG=v${NEXT_VERSION}" >> "$GITHUB_ENV"
+          RELEASE_TAG="v${NEXT_VERSION}"
+          echo "release_tag=${RELEASE_TAG}" >> "$GITHUB_OUTPUT"
+          echo "RELEASE_TAG=${RELEASE_TAG}" >> "$GITHUB_ENV"
 
       - name: Validate release tag
         shell: bash
         run: |
-          if [ -z "${RELEASE_TAG}" ]; then
+          if [ -z "${{ steps.set-version.outputs.release_tag }}" ]; then
```
himanshusinghs: Out of curiosity - the RELEASE_TAG env should've still worked, right? As it is still getting set in the previous step?

himanshusinghs: Another small thing - I thought semgrep would flag this as well to shift the expansion to env variable and use it from there?

gagik: I think the reason why it didn't flag it was because it's in a `"` and/or if statement, but yeah I can move this there too

himanshusinghs: Ahh interesting! Thank you 💯

gagik: combined the 2 steps

PR #1236, opened 2026-01-28
chore: make snyk error when failing

gagik:
Stacked on #1235 and includes changes from https://github.com/mongodb-js/vscode/pull/1201

It was silently failing before.

This makes the test actually fail and moves snyk test into a separate workflow. We can only require at workflow-level and we require **Test and build** but we do not want Snyk to block PRs as its failures might be completely unrelated to them. Instead Snyk test should fail as a separate, non-required workflow to make sure it's on our radar generally.

### DIFF
--- a/.github/workflows/test-and-build.yaml
+++ b/.github/workflows/test-and-build.yaml
@@ -48,6 +48,44 @@ jobs:
         run: pnpm run check
         shell: bash
 
+  build-vsix:
+    name: Build VSIX
+    needs: build-and-check
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node environment
+        uses: ./.github/workflows/actions/setup-node-environment
+
+      - name: Set SEGMENT_KEY
+        env:
+          SEGMENT_KEY: ${{ secrets.SEGMENT_KEY_PROD }}
+        run: |
+          echo "SEGMENT_KEY=${SEGMENT_KEY}" >> $GITHUB_ENV
+        shell: bash
+
+      - name: Build .vsix
+        env:
+          NODE_OPTIONS: "--require ./scripts/no-npm-list-fail.js --max_old_space_size=4096"
+        run: |
+          pnpm exec vsce package --githubBranch main
+        shell: bash
+
+      - name: Check .vsix filesize
+        run: pnpm run check-vsix-size
+        shell: bash
+
+      - name: Upload .vsix
+        uses: actions/upload-artifact@v4
+        with:
+          name: vsix
+          path: '*.vsix'
+
   test:
     name: Run Tests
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -65,7 +66,7 @@
     "check-vsix-size": "node ./scripts/check-vsix-size.ts",
     "release-draft": "node ./scripts/release-draft.js",
     "reformat": "eslint . --fix && prettier --write .",
-    "snyk-test": "node scripts/snyk-test.js",
+    "snyk-test": "node scripts/snyk-test.ts",
```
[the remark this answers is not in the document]
gagik: We have erasable syntax!

scripts/snyk-test.ts
```diff
@@ -23,18 +27,19 @@ async function snykTest(cwd) {
           '--dev',
           `--json-file-output=${tmpPath}`,
         ],
-        { cwd, stdio: 'inherit' },
+        // Do not print anything to the console.
+        { cwd, stdio: 'ignore' },
       );
     } catch (err) {
-      console.warn(err);
+      error = err;
```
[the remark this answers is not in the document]
gagik: eh.... doesn't matter terribly much

.github/workflows/test-and-build.yaml
```diff
@@ -185,7 +200,6 @@ jobs:
             .sbom/snyk-test-result.json
 
       - name: Generate Vulnerability Report (Fail on >= High)
```
gagik: this is its own job now so it can fail

.github/workflows/test-and-build.yaml
```diff
@@ -185,7 +200,6 @@ jobs:
...
-        continue-on-error: ${{ github.event_name == 'pull_request' }}
         shell: bash
         run: |
-          pnpm run generate-vulnerability-report > /dev/null
```
nirinchev, line 161: Was this previously sent to /dev/null to avoid logging potentially sensitive info?

gagik: I suppose. I feel like this removes the whole point of the command... I guess the idea is that it's meant to fail above a certain level but the snyk command already fails... so really not sure what's going on there.

Out of abundance of caution I'll leave this to not print anything

.github/workflows/test-and-build.yaml
```diff
@@ -113,49 +96,26 @@ jobs:
...
         env:
-          RUNNER_OS: ${{ runner.os }}
-        run: pnpm run test-install
+          SEGMENT_KEY: ${{ secrets.SEGMENT_KEY_PROD }}
+        run: |
+          echo "SEGMENT_KEY=${SEGMENT_KEY}" >> $GITHUB_ENV
         shell: bash
 
-  finalize:
-    name: Sign and Upload
```
gagik, line 128: we don't really need to sign our vsix so this is being moved to separat task that is needed by install tests instead

PR #1237, opened 2026-02-05
feat: data-browsing monaco core implementation VSCODE-727

tculig:
## Description

Ticket: https://jira.mongodb.org/browse/VSCODE-727

Base Monaco readonly implementation
image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/views/data-browsing-app/monaco-viewer.tsx
+++ b/src/views/data-browsing-app/monaco-viewer.tsx
@@ -156,8 +156,6 @@ const MonacoViewer: React.FC<MonacoViewerProps> = ({
   }, [themeColors]);
 
   useEffect(() => {
-    console.log(colors);
-    console.log(themeKind);
     if (monaco) {
       monaco.editor.defineTheme('currentVSCodeTheme', {
         base: themeKind,


src/views/data-browsing-app/monaco-viewer.tsx
```diff
@@ -0,0 +1,274 @@
...
+} as const;
+
+const LINE_HEIGHT = 19;
+const EDITOR_PADDING = 0;
+
+/**
+ * Format JSON with unquoted keys (similar to JavaScript object notation)
+ */
+function formatJsonWithUnquotedKeys(obj: any): string {
+  return JSON.stringify(obj, null, 2).replace(/"([^"]+)":/g, '$1:');
```
lerouxb: This regex is gonna bite you. Why not just use `import { toJSString } from 'mongodb-query-parser';` like we use elsewhere? That gives you shell syntax which is what we want to display everywhere.

ie. the same format used in playgrounds, when editing a document, in the shell, in the compass querybar, etc.

src/views/data-browsing-app/document-tree-view.tsx
```diff
@@ -0,0 +1,440 @@
...
+      <div className={cardStyles}>
+        {nodes.map((node, index) =>
+          renderNode(node, index === nodes.length - 1),
+        )}
+      </div>
+    </div>
+  );
+};
+
+export default DocumentTreeView;
```
lerouxb: we're not using this file anywhere, are we?

src/views/dataBrowsingController.ts
```diff
@@ -312,6 +337,14 @@ export default class DataBrowsingController {
     );
     const codiconStylesheetUri = codiconCssUri.toString();
 
+    // Generate the Monaco Editor base URI for the webview
+    // Monaco Editor files are copied to dist/monaco-editor during webpack build
```
lerouxb, line 345: At some point we'll have to double-check that all these files get bundled with the extension. Not sure how/where. Maybe only an issue closer to when we release? But something to keep in mind.

lerouxb: Although I see you have it in the webpack config. Probably enough.

src/views/data-browsing-app/monaco-viewer.tsx
lerouxb: You can imagine shell syntax as just the declarative parts of javascript plus some special global helpers for constructing the bson types.

tculig: not gonna lie, was also looking at this myself, but it seems to work as expected. But happy that there's a better alternative.

src/views/data-browsing-app/document-tree-view.tsx
tculig: no, just didn't get to deleting it..

src/views/data-browsing-app/monaco-viewer.tsx
lerouxb: it will matter for bson types once you get beyond strings and numbers

src/utils/themeColorReader.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+
+function parseThemeFile(
+  themePath: string,
+  colors: JsonTokenColors,
+): JsonTokenColors {
+  try {
+    const content = fs.readFileSync(themePath, 'utf8');
+    // Strip single-line and block comments for JSONC support
+    const stripped = content
```
lerouxb: This breaks the file. Let's rather just use a json5 parser? I see it in some of our apps as transitive deps already anyway.

src/utils/themeColorReader.ts
```diff
@@ -0,0 +1,155 @@
...
+      .replace(/\/\/.*$/gm, '')
+      .replace(/\/\*[\s\S]*?\*\//g, '');
+    const theme: ThemeJson = JSON.parse(stripped);
+
+    if (theme.include) {
+      const parentPath = path.join(path.dirname(themePath), theme.include);
+      if (fs.existsSync(parentPath)) {
+        parseThemeFile(parentPath, colors);
```
lerouxb: This is dead code.

src/utils/themeColorReader.ts
```diff
@@ -0,0 +1,155 @@
...
+          if (scope === pattern || scope.startsWith(pattern + '.')) {
+            colors[colorKey] = foreground;
+            break;
+          }
+        }
+      }
+    }
+
+    return colors;
+  } catch {
```
lerouxb: Maybe at least log the error? We have logging infrastructure that we use in vs code already.

src/utils/themeColorReader.ts
```diff
@@ -0,0 +1,155 @@
...
+  const themeKind = vscode.window.activeColorTheme.kind;
+  const isLight =
+    themeKind === vscode.ColorThemeKind.Light ||
+    themeKind === vscode.ColorThemeKind.HighContrastLight;
+
+  const colors: JsonTokenColors = isLight
+    ? { ...DEFAULT_LIGHT_COLORS }
+    : { ...DEFAULT_DARK_COLORS };
+
+  if (!themeName) {
```
lerouxb, line 181: I think log here if we're going to use the default colours.

src/utils/themeColorReader.ts
```diff
@@ -0,0 +1,155 @@
...
+export function getThemeTokenColors(): JsonTokenColors {
+  const themeName = vscode.workspace
+    .getConfiguration('workbench')
+    .get<string>('colorTheme');
+  const themeKind = vscode.window.activeColorTheme.kind;
+  const isLight =
+    themeKind === vscode.ColorThemeKind.Light ||
+    themeKind === vscode.ColorThemeKind.HighContrastLight;
+
+  const colors: JsonTokenColors = isLight
```
lerouxb: nit, but probably better named defaultColors.

src/utils/themeColorReader.ts
```diff
@@ -0,0 +1,155 @@
...
+    ? { ...DEFAULT_LIGHT_COLORS }
+    : { ...DEFAULT_DARK_COLORS };
+
+  if (!themeName) {
+    return colors;
+  }
+
+  const themePath = findThemeFile(themeName);
+  if (!themePath) {
```
lerouxb, line 187: I think log if we can't find the theme path.

src/views/data-browsing-app/store/messageHandler.ts
```diff
@@ -17,7 +19,13 @@ export const handleExtensionMessage = (
   switch (message.command) {
     case PreviewMessageType.loadPage:
       dispatch(
-        documentsReceived((message.documents as PreviewDocument[]) || []),
+        documentsReceived(
+          message.documents
+            ? (EJSON.deserialize(message.documents, {
```
lerouxb: Does this end up in the redux store? If so then we should probably keep it as EJSON in there because once deserialized it could be binary and that's not allowed in the store.

src/views/data-browsing-app/monaco-viewer.tsx
```diff
@@ -0,0 +1,265 @@
...
+        themeColors?.punctuation ?? DEFAULT_COLORS.punctuation
+      ).replace('#', ''),
+    }),
+    [themeColors],
+  );
+
+  useEffect(() => {
+    if (monaco) {
+      monaco.editor.defineTheme('currentVSCodeTheme', {
+        base: 'vs-dark',
```
lerouxb: maybe we can pass the vscode theme here? or at least try and pick a dark vs light theme based on what we detected in the theme color conversion code.

src/utils/themeColorReader.ts
tculig: turns out it wasn't dead code. It was mutating the colors variable by reference, which is horrible and needs to be refactored.

src/views/data-browsing-app/monaco-viewer.tsx
tculig: yup, good point

src/utils/themeColorReader.ts
tculig: defaultColors is not the correct descriptor, these are the color keys that we will be pulling from the syntax highlighting of the current theme.

src/views/data-browsing-app/store/messageHandler.ts
tculig: that is a new thing that I learned today...

src/utils/themeColorReader.ts
```diff
@@ -0,0 +1,193 @@
...
+  switch (kind) {
+    case vscode.ColorThemeKind.Light:
+      return 'vs';
+    case vscode.ColorThemeKind.HighContrastLight:
+      return 'hc-light';
+    case vscode.ColorThemeKind.HighContrast:
+      return 'hc-black';
+    case vscode.ColorThemeKind.Dark:
+    default:
+      return 'vs-dark';
```
lerouxb, line 171: super nit, feel free to ignore, but I'm wondering if we should just do some kind of substring checks for things like light/dark, etc

We could then at least have a test that lists all the known themes that ship with vscode, loop through them and see which one we pick:

Image

I wonder where they store those groupings. Some metadata we have access to perhaps? Because somehow vscode itself knows to group everything under light, dark or high contrast 🤔 

Anyway. Not necessary for this PR.

tculig: its definitely a rabbit hole...I'd leave if for last if we have the time TBH.

src/views/data-browsing-app/monaco-viewer.tsx
```diff
@@ -0,0 +1,278 @@
...
+    },
+  });
+}
+
+interface MonacoViewerProps {
+  document: Record<string, unknown>;
+  themeColors?: TokenColors;
+  themeKind: MonacoBaseTheme;
+}
+const monacoWrapperStyles = css({
```
lerouxb, line 39: Are you sure this is necessary? I just removed it now and don't see the white boxes. Wasn't it something else that hid them?

I checked because this is just moving it to a specific spot, not seeing anything that actually hides it.

src/views/data-browsing-app/monaco-viewer.tsx
```diff
@@ -0,0 +1,278 @@
...
+      boolean: strip(themeColors.boolean),
+      null: strip(themeColors.null),
+      type: strip(themeColors.type),
+      comment: strip(themeColors.comment),
+      punctuation: strip(themeColors.punctuation),
+    };
+  }, [themeColors]);
+
+  useEffect(() => {
+    if (monaco) {
```
lerouxb, line 159: Tiniest style nit, but if the whole function body is contained inside one if I'd just `if (!monaco) { return; }` at the top so that the nesting is less and it is easier to read. ie. no one then has to scan to find the closing curly bracket.

src/views/data-browsing-app/monaco-viewer.tsx
tculig: for me, if I remove this, the white box flashes on first load, but then disappears...looks janky.

src/views/data-browsing-app/monaco-viewer.tsx
tculig: totally agree, will update for the next PR.

PR #1238, opened 2026-02-09
feat: add a sort dropdown to the documents view VSCODE-741

tculig:
## Description

Ticket: https://jira.mongodb.org/browse/VSCODE-741

image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/test/suite/views/data-browsing-app/actions.test.tsx
+++ b/src/test/suite/views/data-browsing-app/actions.test.tsx
@@ -44,6 +44,23 @@ const createTestState = (
   return { documentQuery: state };
 };
 
+/**
+ * Helper function to find a sort option by its sort field value.
+ * Makes tests more readable than using array indices.
+ */
+const findSortOption = (
+  sortField: '_id',
+  sortDirection: 1 | -1,
+): (typeof SORT_OPTIONS)[number] => {
+  const option = SORT_OPTIONS.find(
+    (opt) => opt.sort?.[sortField] === sortDirection,
+  );
+  if (!option) {
+    throw new Error(`Sort option not found for ${sortField}: ${sortDirection}`);
+  }
+  return option;
+};
+
 describe('actions test suite', function () {
   let postMessageStub: sinon.SinonStub;
 
@@ -222,7 +239,7 @@ describe('actions test suite', function () {
   describe('sortChanged', function () {
     it('should set sort and reset to page 1', function () {
       const store = createStore(createTestState({ currentPage: 3 }));
-      const sortOption = SORT_OPTIONS[1]; // _id: 1
+      const sortOption = findSortOption('_id', 1);
 
       store.dispatch(sortChanged(sortOption));
 
@@ -233,7 +250,7 @@ describe('actions test suite', function () {
 
     it('should send getDocuments with sort and skip=0', function () {
       const store = createStore();
-      const sortOption = SORT_OPTIONS[2]; // _id: -1
+      const sortOption = findSortOption('_id', -1);
 
       store.dispatch(sortChanged(sortOption));
 
@@ -246,7 +263,7 @@ describe('actions test suite', function () {
     });
 
     it('should send getDocuments without sort field when set to null (default)', function () {
-      const sortOption = SORT_OPTIONS[1]; // _id: 1
+      const sortOption = findSortOption('_id', 1);
       const store = createStore(createTestState({ sort: sortOption }));
       store.dispatch(sortChanged(null));
 
@@ -259,7 +276,7 @@ describe('actions test suite', function () {
     });
 
     it('should preserve sort when navigating pages', function () {
-      const sortOption = SORT_OPTIONS[1]; // _id: 1
+      const sortOption = findSortOption('_id', 1);
       const store = createStore(
         createTestState({
           sort: sortOption,
@@ -279,7 +296,7 @@ describe('actions test suite', function () {
     });
 
     it('should preserve sort when refreshing', function () {
-      const sortOption = SORT_OPTIONS[2]; // _id: -1
+      const sortOption = findSortOption('_id', -1);
       const store = createStore(createTestState({ sort: sortOption }));
 
       store.dispatch(documentsRefreshRequested());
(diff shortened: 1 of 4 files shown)


scripts/check-vsix-size.ts
```diff
@@ -15,7 +15,7 @@ const vsixFileName = path.resolve(
 );
 const size = fs.statSync(vsixFileName).size;
 
-const maxSize = 12_000_000;
+const maxSize = 15_000_000;
```
lerouxb, line 18: for posterity:
Image

src/test/suite/views/data-browsing-app/actions.test.tsx
```diff
@@ -217,6 +219,80 @@ describe('actions test suite', function () {
     });
   });
 
+  describe('sortChanged', function () {
+    it('should set sort and reset to page 1', function () {
+      const store = createStore(createTestState({ currentPage: 3 }));
+      const sortOption = SORT_OPTIONS[1]; // _id: 1
```
lerouxb: Maybe either some kind of constants so we don't have to manually look up what 1, 2, etc means when reading the tests or just `SORT_OPTIONS.find((option) => option.sort._id === 1)`? Could also be a helper function. Same for the others below.

src/test/suite/views/data-browsing-app/vscode-api.test.tsx
```diff
@@ -8,6 +8,7 @@ import {
   sendGetDocuments,
   sendGetThemeColors,
 } from '../../../../views/data-browsing-app/vscode-api';
+import { SORT_OPTIONS } from '../../../../views/data-browsing-app/store/documentQuerySlice';
```
lerouxb, line 11: These import paths _almost_ make me prefer monorepos. _almost_. 😆

src/views/data-browsing-app/store/documentQuerySlice.ts
```diff
@@ -8,12 +8,25 @@ import {
...
   [key: string]: unknown;
 }
 
+export interface SortOption {
+  label: string;
+  value: string;
+  sort: DocumentSort | null;
+}
+
+export const SORT_OPTIONS: SortOption[] = [
```
lerouxb, line 24: This is the most nitpicky thing imaginable and you can just ignore it, but value serves no real purpose, right? ie. we could have just used label as the unique key for addressing the option. label gets displayed, sort is the actual value that gets used (but isn't a string so can't be used directly as an option value) and value just exists as the intermediary thing to select an option.

src/views/data-browsing-app/store/documentQuerySlice.ts
```diff
@@ -87,13 +102,13 @@ const documentQuerySlice = createSlice({
       state.errors.getDocuments = null;
       state.errors.getTotalCount = null;
       recalculatePaginationValues(state);
-      sendGetDocuments(0, state.itemsPerPage);
+      sendGetDocuments(0, state.itemsPerPage, current(state).sort);
```
lerouxb: This is a bit of a style thing, but I'm wondering if for future readability we should rather not use positional arguments for these things but something like `{ skip, limit, sort }` rather. Just as a pattern we use elsewhere in our repos. (not 100% sure about this vscode repo).

src/views/data-browsing-app/store/documentQuerySlice.ts
```diff
@@ -125,7 +140,15 @@ const documentQuerySlice = createSlice({
...
       recalculatePaginationValues(state);
-      sendGetDocuments(0, newItemsPerPage);
+      sendGetDocuments(0, newItemsPerPage, current(state).sort);
+    },
+    sortChanged: (state, action: PayloadAction<SortOption | null>) => {
+      state.sort = action.payload;
+      state.currentPage = 1;
+      state.isLoading = true;
+      state.errors.getDocuments = null;
+      recalculatePaginationValues(state);
```
lerouxb, line 175: I know this is now old from a couple PRs ago, but isn't the redux pattern to return new state rather than mutate it?

ie. 

```typescript
return {
  ...state,
  sort: action.payload,
  currentPage: 1,
  isLoading: true,
  errors: {
    ...state.errors,
    getDocuments: null
  },
  ...getPaginationValues(something) // whatever makes sense here
};

```

lerouxb: (so that objects are new values and then react can reliably update because they are changed)

lerouxb: Also obviously for all of the redux code.

src/test/suite/views/data-browsing-app/vscode-api.test.tsx
tculig: I wonder if absolute imports work in this repo...this is just what auto-suggests fills in..

src/views/data-browsing-app/store/documentQuerySlice.ts
tculig: In general yes, but in principle, label could have some non-unicode characters, spaces, etc, and I always worry about string matching those.

src/test/suite/views/data-browsing-app/vscode-api.test.tsx
lerouxb: It is fine. I'm just leaving drive-by comments ;)

src/views/data-browsing-app/store/documentQuerySlice.ts
tculig: For "classic" redux, you would be 100% correct, but redux toolkit uses Immer under the hood, which allows this sort of syntax.

tculig: Its been like this for a few years now. Redux has moved forwards quite a lot.

gagik: Kind of a funny evolution that it went from don't mutate directly to fake mutation.

lerouxb: oof. oh well.

lerouxb (approved):

PR #1239, opened 2026-02-10
feat: add document actions to data browsing - edit, copy, clone, delete VSCODE-729

lerouxb:
VSCODE-729

This adds document actions to data browsing when hovering over the cards - edit, copy, clone, delete. They just do the same thing as the corresponding tree view actions from without the feature flag.
Screenshot 2026-02-10 at 13 00 44
Screenshot 2026-02-10 at 13 00 06


src/editors/playgroundController.ts
```diff
@@ -507,8 +507,12 @@ export default class PlaygroundController {
     if (shouldConfirmRunCopilotCode === true) {
       const name = this._connectionController.getActiveConnectionName();
       const confirmRunCopilotCode = await vscode.window.showInformationMessage(
-        `Are you sure you want to run this code generated by the MongoDB participant against ${name}? This confirmation can be disabled in the extension settings.`,
-        { modal: true },
+        `Are you sure you want to run this code generated by the MongoDB participant against ${name}?`,
+        {
+          modal: true,
+          detail:
```
lerouxb, line 513: As a drive-by I updated all these to put the extra text as detail so it is not just a narrow sea of text. I went looking for a way to make these modals wider because an object id typically does not fit in the width by 1 character. And this is all I found. Still makes these look a little bit better, though.

src/explorer/documentTreeItem.ts
```diff
@@ -57,7 +57,7 @@ export default class DocumentTreeItem
 
     const documentLabel = document._id
       ? JSON.stringify(document._id)
-      : `Document ${documentIndexInTree + 1}`;
+      : `"Document ${documentIndexInTree + 1}"`;
```
lerouxb: As a drive-by I removed the double "" around the text that's usually JSON.stringified and therefore already has "" around it, but then in this case I had to add it back.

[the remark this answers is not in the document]
lerouxb: Actually that's the plan.

tculig: huh, I'm not following...

tculig: This just seems to add outer double qoutes...

tculig:
Do we want to sync the sidebar tree view document count after deletion?
image

src/editors/playgroundController.ts
tculig: This def looks like the correct usage to me.

src/views/dataBrowsingController.ts
```diff
@@ -234,6 +269,116 @@ export default class DataBrowsingController {
...
+    documentId: any,
+  ): Promise<void> => {
+    try {
+      const shouldConfirmDeleteDocument = vscode.workspace
+        .getConfiguration('mdb')
+        .get('confirmDeleteDocument');
+
+      if (shouldConfirmDeleteDocument === true) {
+        const documentIdString = JSON.stringify(
+          EJSON.deserialize(documentId, { relaxed: false }),
```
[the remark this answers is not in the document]
tculig: Should we just deserialize in one place, in the monaco-viewer.tsx, and then pass around the deserialized values?
image

src/explorer/documentTreeItem.ts
tculig: Looks weird in the sidebar tooltip

src/views/data-browsing-app/vscode-api.ts
```diff
@@ -55,3 +46,24 @@ export const sendGetThemeColors = (): void => {
...
+  getVSCodeApi().postMessage({
+    command: PreviewMessageType.editDocument,
+    documentId,
+  });
+};
+
+export const sendCloneDocument = (document: Record<string, unknown>): void => {
+  getVSCodeApi().postMessage({
+    command: PreviewMessageType.cloneDocument,
+    document: EJSON.serialize(document, { relaxed: false }),
```
tculig: The way its setup right now, this is def already serialized at this point.

src/views/dataBrowsingController.ts
```diff
@@ -240,36 +269,137 @@ export default class DataBrowsingController {
...
+    } catch (error) {
+      log.error('Error opening document for editing', error);
+      void vscode.window.showErrorMessage(
+        `Failed to open document: ${formatError(error).message}`,
+      );
+    }
+  };
+
+  handleCloneDocument = async (
+    panel: vscode.WebviewPanel,
```
tculig: unused prop...

src/views/dataBrowsingController.ts
```diff
@@ -240,36 +269,137 @@ export default class DataBrowsingController {
...
+    options: DataBrowsingOptions,
+    document: Record<string, unknown>,
+  ): Promise<void> => {
+    try {
+      const deserialized = EJSON.deserialize(document, { relaxed: false });
+      delete deserialized._id;
+      const documentContents = toJSString(deserialized) ?? '';
+
+      const [databaseName, collectionName] = options.namespace.split(/\.(.*)/s);
```
tculig: Claude says:
Problems with the current approach:
Doesn't handle missing dots: If options.namespace has no dot (e.g., just "database"), collectionName will be undefined
Doesn't handle empty strings: Edge cases like "", ".", or ".collection" aren't handled well
Regex overhead: Using a regex for a simple split operation is unnecessary

tculig: open to discussion...

lerouxb: Hmm didn't even notice that. I'll see what we do elsewhere and I suppose copy it. Ideally we don't even have namespace in options because it just leads to this.

src/views/dataBrowsingController.ts
lerouxb: yeah I was trying to make the parameters consistent between these functions, but I suppose that's pointless.

src/views/data-browsing-app/vscode-api.ts
lerouxb: Aah I confused myself because the unit test wasn't sending it serialized in the same way.

src/explorer/documentTreeItem.ts
lerouxb: The confusion is that that's also used in the delete modal. I'll just remove it and standardise on no quotes. Although the sidebar stuff will be removed once we land the feature flag anyway.

PR #1243, opened 2026-02-11
feat: add default sort order vscode extension setting VSCODE-742

tculig:
## Description

Ticket: https://jira.mongodb.org/browse/VSCODE-742

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/views/data-browsing-app/store/documentQuerySlice.ts
+++ b/src/views/data-browsing-app/store/documentQuerySlice.ts
@@ -85,9 +85,10 @@ const recalculatePaginationValues = (state: DocumentQueryState): void => {
 export const getInitialSort = (): SortOption | null => {
   if (
     typeof window !== 'undefined' &&
-    window.MDB_VSCODE_OPTIONS?.defaultSortOrder
+    window.MDB_DATA_BROWSING_OPTIONS?.defaultSortOrder
   ) {
-    const key = window.MDB_VSCODE_OPTIONS.defaultSortOrder as SortValueKey;
+    const key = window.MDB_DATA_BROWSING_OPTIONS
+      .defaultSortOrder as SortValueKey;
     return SORT_OPTIONS.find((opt) => opt.value === key) ?? null;
   }
   return null;
--- a/src/test/suite/views/data-browsing-app/documentQuerySlice.test.tsx
+++ b/src/test/suite/views/data-browsing-app/documentQuerySlice.test.tsx
@@ -146,16 +146,16 @@ describe('documentQuerySlice', function () {
 
   describe('getInitialSort', function () {
     afterEach(function () {
-      delete window.MDB_VSCODE_OPTIONS;
+      delete window.MDB_DATA_BROWSING_OPTIONS;
     });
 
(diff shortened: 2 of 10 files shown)


src/views/data-browsing-app/store/documentQuerySlice.ts
```diff
@@ -72,11 +88,19 @@ const recalculatePaginationValues = (state: DocumentQueryState): void => {
   );
 };
 
+export const getInitialSort = (): SortOption | null => {
+  if (typeof window !== 'undefined' && window.DEFAULT_SORT_ORDER) {
```
lerouxb: Is there no better way than putting things on window to be passing default parameters to a redux reducer's slice?

lerouxb: Maybe there's a problem here that I can't spot, but surely you can do `vscode.workspace.getConfiguration('mdb').get<SortValueKey>('defaultSortOrder') ?? 'default');` and pass it to `createStore()` as `preloadedState`?

You only need it by the time you pass store to the provider here https://github.com/tculig/vscode-tiho/blob/9a609299927f7206e967296622a246828786bf03/src/views/app.tsx#L23

tculig: this is inside the webview, it shouldn't have access to vscode.workspace, but let me double check..

tculig: yeah, it doesn't, can't run vscode.workspace.getConfiguration('mdb').get<SortValueKey>('defaultSortOrder') inside any of the webview code, including the redux store.

lerouxb: Is there no standard way to pass information to a webview? How do other extensions pass do this?

tculig: As far as I can research, if you want something to be available at load time, this is the only way. Alternative is to post a massage once the webview loads, but that is gonna be prone to race conditions, flicker, etc. It IS a somewhat established pattern according to ChatGPT, Claude Code and obscure StackOverflow questions: https://stackoverflow.com/questions/58326788/how-to-pass-parameters-into-a-script-in-a-vs-code-webview

lerouxb: If there is no better way and this is the best way to do it, can we at least leave a comment where we store things on the window object explaining why. Because I imagine anyone reading the code is going to find it surprising.

As a general rule if we resort to something weird, let's write a detailed comment explaining why.

lerouxb: I suppose we could have also not rendered anything until an initial message with the config arrives (some kind of isLoading state?), but admittedly at this point that's probably more work than it is worth.

tculig: in my experience, doing complex things in onLoad, especially for webviews, is fragile AF. The onLoad flag itself is pretty non-reliable

PR #1250, opened 2026-02-18
feat: add sidebar context menu VSCODE-748

tculig:
## Description

Added the entire context menu for data browsing sidebar entry point:
image

Updated the "no documents" page to have the correct text and the "Insert document" button:
image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/mdbExtensionController.ts
+++ b/src/mdbExtensionController.ts
@@ -971,11 +971,13 @@ export default class MDBExtensionController implements vscode.Disposable {
     );
     this.registerCommand(
       ExtensionCommand.mdbDeleteAllDocuments,
-      async (element: {
+      async ({
+        databaseName,
+        collectionName,
+      }: {
         databaseName: string;
         collectionName: string;
       }): Promise<boolean> => {
-        const { databaseName, collectionName } = element;
         const namespace = `${databaseName}.${collectionName}`;
 
         const confirmationResult = await vscode.window.showInformationMessage(


src/editors/playgroundController.ts
```diff
@@ -177,27 +177,27 @@ export default class PlaygroundController {
...
-      // We need a secondary `mongodb` extension otherwise VSCode will
-      // suggest playground-1.js name when saving playground to the disk.
-      // Users can open playgrounds from the disk
-      // and we need a way to distinguish this files from regular JS files.
-      const fileName = path.join(
-        filePath,
-        `playground-${numberUntitledPlaygrounds + 1}.mongodb.js`,
+      // Find a unique playground number by checking if an untitled document
+      // with the same URI already exists. This avoids relying on isPlayground()
+      // which can fail for untitled URIs, causing the same file to be reused.
```
gagik: can a file both have an untitled URI and be in a file conflict? It's odd that that happens

tculig: Firstly: this was a pre-existing issue, I noticed that opening a playground multiple times just pastes code into the same file, which I don't think is the intended result.  "We compare full URI strings rather than using isPlayground(), which relies on fsPath and can behave incorrectly for untitled-scheme URIs."

gagik: got it, could we update `isPlayground` to not use fsPath? If we can't trust isPlayground then we shouldn't keep it around as is IMO. 

It then might be reason about it with
```ts
const playgroundFiles = vscode.workspace.textDocuments.filter((uri) => isPlayground(uri)).sort();

const lastPlayground = parseInt(playgroundFiles[playgroundFiles.length - 1].replace(filePath....));
```

tculig: reverting this, LeRoux has a solution in his PR

src/views/dataBrowsingController.ts
```diff
@@ -364,57 +364,13 @@ export default class DataBrowsingController {
   };
 
   handleDeleteAllDocuments = async (
-    panel: vscode.WebviewPanel,
+    _panel: vscode.WebviewPanel,
```
gagik: what's the reason for this name change?

src/views/dataBrowsingController.ts
```diff
@@ -364,57 +364,13 @@ export default class DataBrowsingController {
...
-      void panel.webview.postMessage({
-        command: PreviewMessageType.documentDeleted,
-      });
-    } catch (error) {
-      log.error('Error deleting all documents', error);
-      void vscode.window.showErrorMessage(
-        `Failed to delete all documents: ${formatError(error).message}`,
-      );
-    }
+    await vscode.commands.executeCommand<boolean>(
```
gagik: some of the other `handle...` functions that delegate it to a command use a try/catch here, could it be useful?

If a lot of these end up being delegated to commands maybe even inling it in the switch statement could be worth it.

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -367,7 +371,19 @@ const PreviewApp: React.FC = () => {
               />
             ))}
             {displayedDocuments.length === 0 && !getDocumentsError && (
-              <div className={emptyStateStyles}>No documents to display</div>
+              <div className={emptyStateStyles}>
+                <span>This collection has no documents</span>
```
lerouxb, line 375: We'll have to remember to handle this for https://github.com/mongodb-js/vscode/pull/1255

Leaving it at `No documents to display` would have meant we wouldn't have anything to change :)

src/views/dataBrowsingController.ts
lerouxb: Looks like the parameter doesn't get used anymore.

src/views/dataBrowsingController.ts
lerouxb: mdbDeleteAllDocumentsFromTreeView below does seem to handle its errors already. Which is admittedly inconsistent, although that in itself might be OK.

src/mdbExtensionController.ts
```diff
@@ -966,6 +969,61 @@ export default class MDBExtensionController implements vscode.Disposable {
         );
       },
     );
+    this.registerCommand(
+      ExtensionCommand.mdbDeleteAllDocumentsFromTreeView,
+      async (
+        element: ShowPreviewTreeItem | CollectionTreeItem,
```
lerouxb: I don't like reusing element here for two separate things. You're also reusing mdbDeleteAllDocumentsFromTreeView (ie FROM TREE VIEW) from the data browsing controller.

Maybe a better way to reuse the code would be to have two commands, but then they call the same function? 

Or have one neutrally named command that just takes `{ databaseName: string, collectionName: string}` with no mention of element, ShowPreviewTreeItem, CollectionTreeItem or FromTreeView or FromAnything, really.

src/views/dataBrowsingController.ts
tculig: worth being consistent IMO...

src/views/dataBrowsingController.ts
tculig: @lerouxb is right, I'm removing the parameter completely..

src/mdbExtensionController.ts
tculig: updating..

src/mdbExtensionController.ts
```diff
@@ -966,6 +969,62 @@ export default class MDBExtensionController implements vscode.Disposable {
         );
       },
     );
+    this.registerCommand(
+      ExtensionCommand.mdbDeleteAllDocuments,
+      async (element: {
```
lerouxb: nit, but you can also just:

```
{ databaseName, collectionName }: { databaseName: string, collectionName: string }
```

tculig: claude is really dumb sometimes...

PR #1246, opened 2026-02-13
fix: Update node driver to latest

kmruiz:
### Checklist
- [ ] New tests and/or benchmarks are included
- [ ] Documentation is changed or added
- [ ] I have signed the MongoDB Contributor License Agreement (https://www.mongodb.com/legal/contributor-agreement)

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*
- [ ] Patch (non-breaking change which fixes an issue)
- [ ] Minor (non-breaking change which adds functionality)
- [ ] Major (fix or feature that would cause existing functionality to change)

PR #1255, opened 2026-02-23
feat: open cursor results in the data browsing view VSCODE-736

lerouxb:
VSCODE-736

----

**Feature-flag:**

To enable this feature, remember to set the `MDB_USE_ENHANCED_DATA_BROWSING_EXPERIENCE=true` env var:

```
~/mongo/vscode % cat .env
MDB_USE_ENHANCED_DATA_BROWSING_EXPERIENCE=true
```

**Overview:**

If a playground results in a `Cursor` or `AggregationCursor`, we now pass the `constructionOptions` field that was added on to the result [in mongosh](https://github.com/mongodb-js/mongosh/pull/2664) to the Data Browser webview which can then use it to rebuild the query that was run in the playground and ultimately use that to paginate through the results just like how we can now browse through all documents in a collection.

Screenshot 2026-02-24 at 18 47 14

**Tab Title:**

This view has the title "Playground Results: databaseName.collectionName" to differentiate it from other "Playground Results" views and from the view you get when you navigate a collection's documents. (Which is just titled databaseName.collectionName.)

**Position:**

Like regular playground results (and unlike the collection data browser) it also opens in a split panel next to the playground.

**Fallback to existing functionality:**

For results that are not a find or aggregate cursor we fall back to the existing functionality where we just open the results in a "file". It should be easy o make this work for `aggregateDb` and `runCursorCommand` too, but I left those out for now because those different from what we have in that they don't have a `collectionName`. Maybe a follow-up?

**Removed toolbar items:**

The toolbar on this view has the Bulk Delete action as well as the Sort dropdown hidden. We could probably add support for bulk delete relatively easily, but sort is tricky because the user could have sorted on anything and the dropdown just has default, _id: 1 and _id: -1.

**Disabled count query:**

The total matching document count query is disabled for now, so the total is always null. (rendered as N/A). I'll file a follow-up to try an make that work, at least for most cases.

**Context actions:**

Right now I left Edit, Copy, Clone and Delete unchanged. There was talk of disabling clone & delete if there's a projection because clone would insert a partial document and delete could be deleting something different to what you're looking at. Copy is obviously safe to keep regardless and Edit is safe because it opens the whole document by id, so you should either see the whole document or get an error.

---

**Testing:**

* For find we extract and strip limit and skip from the findOptions and/or chains. We then take it into account, when we use skip and limit in order to paginate. ie. you can't go past the limit that was in the original query and the documents the query skipped remain skipped. (So this should behave the same as Compass' Documents tab.) Check that all that actually works.

* We leave aggregations alone and rely on the fact that you can have multiple $limit and/or $skip stages. Check that this assumption works.

* There is a lot of room for "off by one" errors in pagination code in general. Double-check that. (That's why I have a numbers collection that just has 0 to 100 for 101 numbers total as documents in a collection.)

* Check that next works even with a null total. (which should always be the case for now).

* Check that prev works.

* Check that refresh works.

* Check that you can insert a doc. (do we want to keep this? Or do we want to also drop this if there's a projection or something going on?)

* Check the N/A text in the toolbar's tooltip.

* Check context actions (edit, copy, clone, delete) for all sorts of queries.

* Check that all chainable methods on cursors work as expected. See mongosh' code:

Screenshot 2026-02-24 at 19 07 41

---

**TODO:**

* [x] tests
* [ ] factor out some code from dataBrowsingController.ts. It is getting a bit unwieldy to work on.
* [x] make sure this is feature flagged for now 😬 
* [x] file follow-up for getting count to work (it is null for now) (https://jira.mongodb.org/browse/VSCODE-754)
* [x] file follow-up for aggregateDb and runCursorCommand (https://jira.mongodb.org/browse/VSCODE-756)
* [x] file follow-up for possibly disabling clone and delete if there is a projection. (and insert?) (https://jira.mongodb.org/browse/VSCODE-755)
* [ ] check that all chainable methods work on cursors as expected. We might have to check for some of them and just fall back to the old playground results in that case. (map? pretty?)

### DIFF
--- a/src/views/dataBrowsingController.ts
+++ b/src/views/dataBrowsingController.ts
@@ -653,7 +653,10 @@ export default class DataBrowsingController {
     if (!connectionOptions) {
       throw new Error('No connection options found');
     }
-    // TODO: connect only once and reuse the connection for subsequent requests instead of reconnecting every time
+    // TODO(VSCODE-757, VSCODE-758): connect only once and reuse the connection
+    // for subsequent requests instead of reconnecting every time. Or switch to
+    // only using node service provider in the whole extension, then use the
+    // same connection everywhere.
     const serviceProvider = await NodeDriverServiceProvider.connect(
       connectionOptions.url,
       connectionOptions.options,
@@ -738,7 +741,10 @@ export default class DataBrowsingController {
     if (!connectionOptions) {
       throw new Error('No connection options found');
     }
-    // TODO: connect only once and reuse the connection for subsequent requests instead of reconnecting every time
+    // TODO(VSCODE-757, VSCODE-758): connect only once and reuse the connection
+    // for subsequent requests instead of reconnecting every time. Or switch to
+    // only using node service provider in the whole extension, then use the
+    // same connection everywhere.
     const serviceProvider = await NodeDriverServiceProvider.connect(
       connectionOptions.url,
       connectionOptions.options,


src/mcp/mcpController.ts
```diff
@@ -408,11 +410,11 @@ ${jsonConfig}`,
 
       const connectParams: MCPConnectParams | undefined =
         connectionId && mongoClientOptions
-          ? {
+          ? ({
               connectionId: connectionId,
               connectionString: mongoClientOptions.url,
               connectOptions: mongoClientOptions.options,
-            }
+            } as MCPConnectParams)
```
lerouxb: from https://github.com/mongodb-js/vscode/pull/1246, will go away once that's merged.

src/test/suite/views/data-browsing-app/actions.test.tsx
```diff
@@ -25,8 +25,8 @@ const createTestState = (
 
   // Recalculate computed pagination values based on overrides
   state.totalDocuments =
-    state.totalCountInCollection !== null
-      ? state.totalCountInCollection
+    state.totalCountForQuery !== null
```
lerouxb, line 28: I just renamed totalCountInCollection to totalCountForQuery because the relevant count is not going to always be for a collection going forward.

src/views/data-browsing-app/store/documentQuerySlice.ts
```diff
@@ -44,42 +45,108 @@ export interface ErrorsState {
...
+  totalCountForQuery: number | null;
+
   itemsPerPage: number;
+
+  // for now we will only display controls like bulk delete and the total count
+  // for basic queries (ie. query === null)
   sort: SortOption | null;
+  query: ServiceProviderQuery | null;
+
+  // TODO: these are derived values, why are we storing them?
```
lerouxb: Feels like in regular redux this would be a big no-no. I don't know about redux toolkit.

src/views/data-browsing-app/store/documentQuerySlice.ts
```diff
@@ -44,42 +45,108 @@ export interface ErrorsState {
...
+};
+
+export const isCursorQuery = (
+  state: any,
+): state is DocumentQueryState & { query: ServiceProviderQuery } => {
+  return state.query !== null;
+};
+
+const recalculatePaginationValues = (
+  state: Draft<DocumentQueryState>,
```
lerouxb, line 112: Added Draft<> because redux toolkit wasn't happy. Will have to confirm if that's really necessary and if it is something I'm doing wrong.

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -83,7 +85,7 @@ const paginationInfoStyles = css({
   fontSize: '13px',
   whiteSpace: 'nowrap',
   display: 'flex',
-  gap: spacing[200],
+  gap: spacing[100],
```
lerouxb, line 87: This was just twice as much as it should have been and we never noticed.

src/views/dataBrowsingController.ts
```diff
@@ -46,16 +53,144 @@ function getDefaultDocumentSort(): DocumentSort | undefined {
...
+  query: CursorConstructionOptionsWithChains & { options: { method: 'find' } },
+): ParsedQuery {
+  let limit: number | null = null;
+  let skip: number | null = null;
+
+  const [, , filter, findOptions, dbOptions] = query.options.args;
+  if (findOptions) {
+    if (findOptions.limit) {
+      limit = findOptions.limit;
+      delete findOptions.limit;
```
tculig: This feels a bit dodgy...breaks the immutability principles.

src/views/dataBrowsingController.ts
```diff
@@ -487,44 +671,164 @@ export default class DataBrowsingController {
...
+    }
     if (skip !== undefined && skip > 0) {
-      findOptions.skip = skip;
+      // we add the skip from the pagination onto the parsed query skip if it is exists
+      findOptions.skip = (findOptions.skip ?? 0) + skip;
+    }
+
+    if (parsedQuery.limit) {
+      // if the parsed query has a limit we want to honor that so that we don't go past it
+      const remaining = parsedQuery.limit - (findOptions.skip ?? 0);
```
tculig, line 686: "findOptions.skip here is the combined value of parsedQuery.skip + pagination skip. But parsedQuery.limit is the user's original limit which is relative to where the cursor starts (after parsedQuery.skip). The  remaining calculation should subtract only the pagination skip (not the full skip including the query's own skip). "

tculig: ```const remaining = parsedQuery.limit - skip; ```

src/views/data-browsing-app/store/documentQuerySlice.ts
```diff
@@ -44,42 +45,101 @@ export interface ErrorsState {
...
+        // database, pipeline, aggregateOptions, dbOptions
+        args: Parameters<ServiceProvider['aggregateDb']>;
+      }
+    | {
+        method: 'runCursorCommand';
+        // database, spec, runCursorCommandOptions, dbOptions
+        args: Parameters<ServiceProvider['runCursorCommand']>;
+      };
+  chains: {
+    method: 'string';
```
tculig:
```suggestion
    method: string;
```

src/views/dataBrowsingController.ts
lerouxb: Yeah I should probably operate off a copy. Will change.

src/views/dataBrowsingController.ts
lerouxb: Hmm. I better add tests for this specifically. That should make these clear one way or another :)

src/views/data-browsing-app/store/documentQuerySlice.ts
lerouxb: oof 🤦

src/views/data-browsing-app/preview-page.tsx
```diff
@@ -308,7 +322,7 @@ const PreviewApp: React.FC = () => {
               onClick={(): void => {
                 dispatch(previousPageRequested());
               }}
-              disabled={currentPage <= 1 || isLoading}
+              disabled={(totalPages !== null && currentPage <= 1) || isLoading}
```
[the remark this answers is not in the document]
lerouxb: no.

src/views/dataBrowsingController.ts
```diff
@@ -487,44 +675,168 @@ export default class DataBrowsingController {
...
+    const connectionOptions =
+      this._connectionController.getMongoClientConnectionOptions();
+    if (!connectionOptions) {
+      throw new Error('No connection options found');
     }
+    // TODO: connect only once and reuse the connection for subsequent requests instead of reconnecting every time
+    const serviceProvider = await NodeDriverServiceProvider.connect(
+      connectionOptions.url,
+      connectionOptions.options,
+    );
```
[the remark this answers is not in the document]
lerouxb: Yes. It says so right there.

PR #1257, opened 2026-02-26
chore: switch vscode to use ai SDK VSCODE-753

gagik:
`mongodb-rag-core` is deprecated so instead we adopt the Docs API with AI SDK directly. We should consider having a `devtools-shared` tool library for this in the future to avoid compass - vscode - etc. code duplication for the transport.

**Note: Will be good to wait for a review from someone from the Docs AI team before merging.**

### DIFF
--- a/src/participant/docsChatbotAIService.ts
+++ b/src/participant/docsChatbotAIService.ts
@@ -60,7 +60,7 @@ export class DocsChatbotAIService {
       },
       providerOptions: {
         openai: {
-          store: false,
+          store: true,
         },
       },
     }) as unknown as DocsStreamResult;


src/participant/chatMetadata.ts
```diff
@@ -4,7 +4,6 @@ import { v4 as uuidv4 } from 'uuid';
 export type ChatMetadata = {
   databaseName?: string;
   collectionName?: string;
-  docsChatbotConversationId?: string;
```
gagik, line 7: wondering if concept of conversation ID might still be possible, will see.

src/test/suite/participant/participant.test.ts
```diff
@@ -1743,21 +1748,21 @@ Schema:
           // Expect the error to be reported through the telemetry service
           expect(
             telemetryTrackStub.getCalls(),
-          ).to.have.length.greaterThanOrEqual(2);
+          ).to.have.length.greaterThanOrEqual(4);
```
gagik: 1. First PromptSubmitted: Tracks the attempt to use AI SDK
2. ParticipantResponseFailed
3. Second PromptSubmitted: Tracks the fallback attempt to use VSCode Copilot
4. ParticipantResponseGenerated

gagik: Not 100% sure the original test was correct but this result makes sense...

src/test/suite/participant/participant.test.ts
```diff
@@ -1743,21 +1748,21 @@ Schema:
...
 
-          const firstTelemetryEvent = telemetryTrackStub.firstCall
+          const failedTelemetryEvent = telemetryTrackStub.getCalls()[1]
             .args[0] as ParticipantResponseFailedTelemetryEvent;
-          expect(firstTelemetryEvent.type).to.equal(
+          expect(failedTelemetryEvent.type).to.equal(
             'Participant Response Failed',
```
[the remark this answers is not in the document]
gagik: true but also this can be a good thing? if more events are appearing we should consider whether these flows should expect them.

src/participant/docsChatbotAIService.ts
```diff
@@ -51,168 +29,40 @@ export class DocsChatbotAIService {
...
-    signal?: AbortSignal;
-    body?: string;
-    headers?: { [key: string]: string };
-  }): Promise<Response> {
-    return fetch(uri, {
+  private _createModel(): ReturnType<
+    ReturnType<typeof createOpenAI>['responses']
+  > {
+    const openai = createOpenAI({
+      baseURL: `${this._serverBaseUri}/api/${MONGODB_DOCS_CHATBOT_API_VERSION}`,
```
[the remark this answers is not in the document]
gagik: not wrong but realistically no user is overriding this and if they are, they will probably derive the format.

src/participant/docsChatbotAIService.ts
```diff
@@ -51,168 +29,40 @@ export class DocsChatbotAIService {
...
+    return streamText({
+      model: this._createModel(),
+      messages,
+      abortSignal: signal,
+      headers: {
+        'X-Request-Origin': `vscode-mongodb-copilot-v${version as string}/docs`,
+      },
+      providerOptions: {
+        openai: {
+          store: true,
```
gagik, line 63: (matches previous behavior)

PR #1258, opened 2026-03-02
feat: remove the data browsing feature flag VSCODE-730

lerouxb:
VSCODE-730

This also removes the now-dead code.

### DIFF
--- a/src/explorer/collectionTreeItem.ts
+++ b/src/explorer/collectionTreeItem.ts
@@ -10,8 +10,6 @@ import type TreeItemParent from './treeItemParentInterface';
 import SchemaTreeItem from './schemaTreeItem';
 import { CollectionType } from './documentUtils';
 
-export const MAX_DOCUMENTS_VISIBLE = 10;
-
 function getIconPath(
   type: string,
   isExpanded: boolean,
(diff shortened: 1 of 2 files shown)


src/explorer/collectionTreeItem.ts
```diff
@@ -64,7 +58,7 @@ function isChildCacheOutOfSync(
     ? collapsibleState !== vscode.TreeItemCollapsibleState.Expanded
     : collapsibleState !== vscode.TreeItemCollapsibleState.Collapsed;
 }
-export type DocumentsTreeItem = ShowPreviewTreeItem | DocumentListTreeItem;
+export type DocumentsTreeItem = ShowPreviewTreeItem;
```
tculig: we can probably do without this rename

src/explorer/schemaTreeItem.ts
```diff
@@ -7,9 +7,10 @@ import FieldTreeItem from './fieldTreeItem';
 import formatError from '../utils/formatError';
 import { getImagesPath } from '../extensionConstants';
 import type TreeItemParent from './treeItemParentInterface';
-import { MAX_DOCUMENTS_VISIBLE } from './documentListTreeItem';
 import type { DataService } from 'mongodb-data-service';
 
+export const MAX_DOCUMENTS_VISIBLE = 10;
```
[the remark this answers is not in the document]
tculig: I feel like this was an incorrect deduction from Copilot..

src/explorer/collectionTreeItem.ts
lerouxb: good point. Didn't notice that.

PR #1259, opened 2026-03-02
feat: Add data browsing telemetry events VSCODE-738

tculig:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/src/test/suite/views/dataBrowsingController.test.ts
+++ b/src/test/suite/views/dataBrowsingController.test.ts
@@ -14,6 +14,12 @@ import { CollectionType } from '../../../explorer/documentUtils';
 import { EJSON } from 'bson';
 import ExtensionCommand from '../../../commands';
 import { NodeDriverServiceProvider } from '@mongosh/service-provider-node-driver';
+import {
+  DataBrowserDocumentsFetchedTelemetryEvent,
+  DataBrowserDocumentEditedTelemetryEvent,
+  DataBrowserDocumentClonedTelemetryEvent,
+  DataBrowserDocumentInsertedTelemetryEvent,
+} from '../../../telemetry';
 
 suite('DataBrowsingController Test Suite', function () {
   const sandbox: SinonSandbox = sinon.createSandbox();
@@ -39,6 +45,7 @@ suite('DataBrowsingController Test Suite', function () {
     refresh: SinonStub;
     refreshCollection: SinonStub;
   };
+  let trackStub: SinonStub;
 
   function createMockPanel(): vscode.WebviewPanel {
     postMessageStub = sandbox.stub().resolves(true);
@@ -90,9 +97,10 @@ suite('DataBrowsingController Test Suite', function () {
       refresh: sandbox.stub().returns(true),
       refreshCollection: sandbox.stub().returns(true),
     };
+    trackStub = sandbox.stub();
     testController = new DataBrowsingController({
       connectionController: mockConnectionController as any,
-      telemetryService: { track: sandbox.stub() } as any,
+      telemetryService: { track: trackStub } as any,
     });
     mockPanel = createMockPanel();
   });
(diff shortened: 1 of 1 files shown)


src/telemetry/telemetryEvents.ts
```diff
@@ -695,6 +695,74 @@ export class DeepLinkTelemetryEvent implements TelemetryEventBase {
...
+
+/** Reported when a new document is inserted from the data browser */
+export class DataBrowserDocumentInsertedTelemetryEvent
+  implements TelemetryEventBase
+{
+  type = 'Data Browser Document Inserted';
+  properties = {};
+}
+
+/** Reported when one or more documents are deleted from the data browser */
```
lerouxb: We should split the delete all documents action from the action where we delete one document from the document card into separate telemetry events.

src/telemetry/telemetryEvents.ts
```diff
@@ -695,6 +695,74 @@ export class DeepLinkTelemetryEvent implements TelemetryEventBase {
...
+{
+  type = 'Data Browser Document Cloned';
+  properties = {};
+}
+
+/** Reported when a new document is inserted from the data browser */
+export class DataBrowserDocumentInsertedTelemetryEvent
+  implements TelemetryEventBase
+{
+  type = 'Data Browser Document Inserted';
```
lerouxb, line 779: I think anything you can do from the tree view and also from the data browser we want to track which one it came from so we know which ones get used more. We can probably do that with an extra property? Like `view: 'tree' | 'data-browser'` or something like that.

So that would be refresh, delete all documents and insert document.

src/telemetry/telemetryEvents.ts
```diff
@@ -695,6 +695,74 @@ export class DeepLinkTelemetryEvent implements TelemetryEventBase {
   }
 }
 
+/** Reported when the data browser webview is opened */
+export class DataBrowserOpenedTelemetryEvent implements TelemetryEventBase {
+  type = 'Data Browser Opened';
+  properties: {
+    /** The type of the collection being browsed - e.g. 'collection', 'view', 'timeseries' */
+    collection_type: string;
```
lerouxb, line 706: Oh yeah we also have 'unknown' for query results. Which I was never sure about. We could try and look up what the collection type was, but then we probably have to detect at which point you're doing a more complicated query where that's not really true anymore. Or just live with 'unknown' as a possible value for now.

lerouxb: Oh and we probably want to differentiate between the "basic" case where you're navigating a collection and the "playground cursor results" case (not sure what to name them, really) so that we can track which ones are more popular and if people even use one vs the other and from which one users are more likely to follow with some other action.

I'm thinking that should probably extend to all the actions?

src/telemetry/telemetryEvents.ts
```diff
@@ -720,4 +845,11 @@ export type TelemetryEvent =
   | PresetConnectionEditedTelemetryEvent
   | SidePanelOpenedTelemetryEvent
   | TreeItemExpandedTelemetryEvent
-  | DeepLinkTelemetryEvent;
+  | DeepLinkTelemetryEvent
+  | DataBrowserOpenedTelemetryEvent
```
lerouxb, line 862: Can we add a DataBrowserClosedTelemetryEvent if possible? If it turns out to require code surgery let's just file a follow-up.

src/telemetry/telemetryEvents.ts
tculig: implementing

src/telemetry/telemetryEvents.ts
tculig: implemented

PR #1260, opened 2026-03-03
chore: fix snyk, move security treshold to high, bump dependencies VSCODE-753

gagik:
We are most likely unaffected by any of these dependencies but it's still worth bumping them.

### DIFF
--- a/scripts/snyk-test.ts
+++ b/scripts/snyk-test.ts
@@ -6,6 +6,7 @@ const { promisify } = require('util');
 const execFile = promisify(childProcess.execFile);
 
 async function snykTest(cwd): Promise<{
+  error?: Error;
   results: Record<string, unknown>;
 }> {
   const tmpPath = path.join(os.tmpdir(), 'tempfile-' + Date.now());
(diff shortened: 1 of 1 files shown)


scripts/snyk-test.ts
```diff
@@ -15,29 +14,25 @@ async function snykTest(cwd): Promise<{
...
         // Do not print anything to the console.
         { cwd, stdio: 'ignore' },
       );
     } catch (err) {
-      error = err;
+      // Snyk exits with code 1 when vulnerabilities are found — that's
+      // expected and not a script-level failure. Any other exit code
+      // (e.g. 2 for a Snyk runtime error) should still be surfaced.
+      if (err.code !== 1) {
+        throw err;
```
lerouxb: If the task doesn't fail, will we still be notified of vulnerabilities somehow?

gagik: generate-vulnerability-report is meant to fail here but it doesn't work so I'll revert that last commit which made this change

PR #1267, opened 2026-03-04
use shell-api cursors, not service provider cursors

lerouxb:
.skip() and .limit() on find() works because we strip those out. .projection(), .pretty(), etc. doesn't work on find/aggregate because those are our shell-api cursor methods and somehow I forgot to wrap the service provider cursors.

### DIFF
--- a/src/views/dataBrowsingController.ts
+++ b/src/views/dataBrowsingController.ts
@@ -713,71 +713,71 @@ export default class DataBrowsingController {
       connectionOptions.options,
     );
 
-    const findOptions: {
-      limit: number;
-      skip?: number;
-      sort?: DocumentSort;
-      promoteValues: false;
-    } = {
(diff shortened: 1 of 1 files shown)


src/views/dataBrowsingController.ts
```diff
@@ -797,28 +810,35 @@ export default class DataBrowsingController {
...
+    );
+    const mongo = new Mongo(instanceState);
 
-      for (const chain of parsedQuery.chains) {
-        cursor = cursor[chain.method](...chain.args);
-      }
+    const pipeline = [
+      ...parsedQuery.aggregate.pipeline,
+      { $skip: skip },
+      { $limit: limit },
```
himanshusinghs: I suppose we will deal with the $out and $merge in follow up PR?

lerouxb: yup. separate issue. also .map() doesn't work.

PR #1268, opened 2026-03-04
only allow 'safe' cursors to open in the data browser

lerouxb:


src/language/worker.ts
```diff
@@ -152,19 +153,53 @@ export const execute = async ({
...
+  }: HandleMessageDeps = {},
+): Promise<void> => {
   if (name === ServerCommand.executeCodeFromPlayground) {
-    parentPort?.postMessage({
+    const payload = await executeFn(data);
+    // .map() cannot be cloned by  so just strip the constructionOptions and
+    // then it won't be opened in the data browser and this result gets the
+    // usual fallback experience
+    const safePayload =
+      payload.data?.result && isSafeQueryResultFn(payload.data.result)
```
lerouxb, line 194: This file changed a bit to make handleMessageFromParentPort testable, but the point is this line.

lerouxb: (we didn't have tests for this before)

PR #1277, opened 2026-03-09
chore: add an E2E test VSCODE-767

lerouxb:
This repo doesn't have any e2e tests. There are barely integration tests. Almost everything stubs / mocks lots of things.

I had a bug with an aggregation that used BSON values where those BSON values silently got turned into strings in the round trip through all the IPC layers to mongosh and back. It would be nice to be able to catch that sort of thing.

Also releasing is very manual and involves a manual testing matrix. We should be able to cover most of it with tests pretty easily.

So this is the copilot + claude opus 4.6 + some hand-holding and iteration first draft at adding e2e tests.

As a drive-by I fixed the eslint setup to work properly in vscode after the pnpm switch so that it finds all the same things that gets found when running eslint in CI. And there are now no errors for the ESLint task in the OUTPUT panel.

---

Suggested follow-up: Run the e2e tests against a built package as part of release.

---

**pro tip**: use `pnpm run test-e2e-only` to skip the pre-test-e2e hook and therefore not pay the cost of `pnpm run compile` every time if you don't have to.

### DIFF
--- a/src/test/e2e/helpers.ts
+++ b/src/test/e2e/helpers.ts
@@ -239,6 +239,10 @@ export async function launchVSCode(): Promise<ElectronApplication> {
       '--disable-extensions',
       '--disable-gpu',
       '--no-sandbox',
+      // On CI (Linux), SecretStorage uses the system keyring
+      // (gnome-keyring/kwallet/etc.). If no keyring daemon is available, this
+      // call hangs indefinitely, blocking the entire flow.
+      '--password-store=basic',
       `--user-data-dir=${userDataDir}`,
       `--extensions-dir=${extensionsDir}`,
       '--skip-release-notes',


package.json
```diff
@@ -55,6 +55,7 @@
     "test": "pnpm run test-webview && pnpm run test-extension",
     "test-extension": "cross-env MDB_IS_TEST=true NODE_OPTIONS=--no-force-async-hooks-checks xvfb-maybe node ./out/test/runTest.js",
     "test-webview": "mocha -r ts-node/register --exit --grep=\"${MOCHA_GREP}\" --file ./src/test/setup-webview.ts src/test/suite/views/**/*.test.tsx",
+    "test-e2e": "pnpm run compile && playwright test --config src/test/e2e/playwright.config.ts",
```
lerouxb: This isn't being called in CI yet.

package.json
```diff
@@ -52,9 +52,12 @@
     "watch:extension": "tsc -p ./ --watch",
     "watch:extension-bundles": "webpack --mode development --watch",
     "pretest": "pnpm run compile",
-    "test": "pnpm run test-webview && pnpm run test-extension",
+    "test": "pnpm run test-e2e && pnpm run test-webview && pnpm run test-extension",
     "test-extension": "cross-env MDB_IS_TEST=true NODE_OPTIONS=--no-force-async-hooks-checks xvfb-maybe node ./out/test/runTest.js",
     "test-webview": "mocha -r ts-node/register --exit --grep=\"${MOCHA_GREP}\" --file ./src/test/setup-webview.ts src/test/suite/views/**/*.test.tsx",
+    "pretest-e2e": "pnpm run compile",
+    "test-e2e": "pnpm run test-e2e-only",
+    "test-e2e-only": "xvfb-maybe playwright test --config src/test/e2e/playwright.config.ts",
```
gagik, line 60: why not just have test-e2e?

lerouxb: from the description:

pro tip: use pnpm run test-e2e-only to skip the pre-test-e2e hook and therefore not pay the cost of pnpm run compile every time if you don't have to.

lerouxb: I just structured it that way to make it easy to run just the tests without waiting many minutes for it to compile when you know you just changed the test and not the code under test. Feels worth it especially since it is lightning fast. ie. much faster than unit tests.

gagik: got it, sgtm

PR #1278, opened 2026-03-12
feat: add skills VSCODE-766

nirinchev:
Adds the agent-skills submodule and exposes them through the vscode extension.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -19,9 +19,6 @@ playgrounds/**
 resources
 .sbom/**
 CHANGELOG.md
-ext/**
-!ext/agent-skills/skills/**
-ext/agent-skills/skills/mongodb-mcp-setup/**
 .claude/**
 .github/**
 .husky/**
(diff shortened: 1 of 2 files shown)


.vscodeignore
```diff
@@ -19,3 +19,16 @@ playgrounds/**
 resources
 .sbom/**
 CHANGELOG.md
+ext/**
+!ext/agent-skills/skills/**
+ext/agent-skills/skills/mongodb-mcp-setup/**
```
tculig: Is this a temporary override, or a pattern?

gagik:
should we have this as a submodule or is there a world where we would use some tooling like [npx skills](https://skills.sh/) to keep in the loop with the published content?

.vscodeignore
nirinchev: Sorry, forgot to respond to this - we're replacing the mcp-setup skill (which is general-purpose one) with one specifically for VSCode as the way they are configured is different - env variables vs settings.

PR #1306, opened 2026-04-13
fix: add the missing globals we use to the playground templates

lerouxb:
I noticed that this started happening where some newly created playgrounds stated off red:

Screenshot 2026-04-13 at 11 41 13
Screenshot 2026-04-13 at 11 55 25

So far it seems to only happen if vscode's active workspace is vscode itself. The error is coming from eslint and I guess at some point something changed in how it resolves the eslint config or in vscode's config or it picks up vscode's config..

Probably not really an issue for most real users, but I did notice that all our playground templates except three have the `/* global use, db */` (or whatever globals get used in the template) pattern at the top and that fixes it.

Might as well align those?

---

Four options:
* it is effectively the user's eslint config breaking it here, so we can decide it is their responsibility
* we can add all the globals they could ever use, add eslint config to ignore unused globals or add eslint config to just turn off validation for using undefined globals, etc. (some variation or combination of those)
* we could not make playgrounds .js files but rather make a new format that just happens to be using js parsers under the hood and then eslint or similar _cannot_ accidentally match it
* we could just live with it ;)

### DIFF
--- a/src/templates/playgroundFromCollectionTreeItemTemplate.ts
+++ b/src/templates/playgroundFromCollectionTreeItemTemplate.ts
@@ -1,7 +1,8 @@
 import { createTemplate } from './templateHelpers';
 
 export const playgroundFromCollectionTreeItemTemplate = createTemplate(
-  (databaseName, collectionName) => `// MongoDB Playground
+  (databaseName, collectionName) => `/* global use, db */
+// MongoDB Playground
 // Use Ctrl+Space inside a snippet or a string literal to trigger completions.
 
 // The current database to use.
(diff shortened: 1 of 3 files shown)


gagik (approved):

himanshusinghs (approved):

nirinchev (approved):

src/templates/playgroundFromCollectionTreeItemTemplate.ts
```diff
@@ -1,7 +1,8 @@
 import { createTemplate } from './templateHelpers';
 
 export const playgroundFromCollectionTreeItemTemplate = createTemplate(
-  (databaseName, collectionName) => `// MongoDB Playground
+  (databaseName, collectionName) => `/* global use, db */
+// MongoDB Playground
```
Anemy, line 5: Since this is the same string in every playground, would it make sense for us to put it into a helper function? So all of them then call that templateHeader. Not a blocker, fine as is.

Anemy (approved):

PR #1307, opened 2026-04-14
fix: undefined query should work like null query VSCODE-774

lerouxb:
VSCODE-774

### DIFF
--- a/src/views/data-browsing-app/store/documentQuerySlice.ts
+++ b/src/views/data-browsing-app/store/documentQuerySlice.ts
@@ -106,7 +106,7 @@ const DEFAULT_ITEMS_PER_PAGE = 10;
 export const isBasicQuery = (
   state: any,
 ): state is DocumentQueryState & { query: null } => {
-  return state.query === null || state.query === undefined;
+  return !!state.query;
 };
 
 const recalculatePaginationValues = (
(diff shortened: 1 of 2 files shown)


Anemy (approved):
I might rename `isBasicQuery` to something like `isEmptyFindResult` or something, that's a bit of a nit and unrelated, fine as is. The copilot comment is something we likely want to fix, the type gaurd isn't correct anymore.

lerouxb:
It isn't empty finds, though. That's still a cursor. By basic I mean is not a cursor. Could have checked the inverse too, though.

src/views/data-browsing-app/store/documentQuerySlice.ts
```diff
@@ -106,7 +106,7 @@ const DEFAULT_ITEMS_PER_PAGE = 10;
 export const isBasicQuery = (
   state: any,
 ): state is DocumentQueryState & { query: null } => {
-  return state.query === null;
+  return state.query === null || state.query === undefined;
```
mabaasit: [nit] I recently learnt you can also do `state.query == null` to cover both.

mabaasit (approved):

src/views/data-browsing-app/store/documentQuerySlice.ts
lerouxb: In this case it is either {} or should be null so I'm just gonna go with !state.query

src/views/data-browsing-app/store/documentQuerySlice.ts
```diff
@@ -168,7 +168,7 @@ export const getInitialQuery = (): ServiceProviderQuery | null => {
     window.MDB_DATA_BROWSING_OPTIONS?.query
   ) {
     const queryString = window.MDB_DATA_BROWSING_OPTIONS.query;
-    return deserializeBSON(queryString) as ServiceProviderQuery;
+    return (deserializeBSON(queryString) as ServiceProviderQuery) || null;
```
lerouxb, line 171: The real bug is that some things deserialize to `undefined` and it should be either a ServiceProviderQuery or null

PR #1310, opened 2026-05-21
feat: guard remote code execution VSCODE-759

alenakhineika:
VSCODE-759

## Description

- The playground creation templates (`mdbRunAllPlaygroundBlocks`, `mdbRunSelectedPlaygroundBlocks`, `mdbRunAllOrSelectedPlaygroundBlocs`) were converted to use the `createTemplate` helper, which applies `JSON.stringify` to all arguments before inserting.
- The playground run commands (`mdbRunAllPlaygroundBlocks`, `mdbRunSelectedPlaygroundBlocks`, `mdbRunAllOrSelectedPlaygroundBlocs`) have been moved to `DEEP_LINK_DISALLOWED_COMMANDS`.
- The tree-item-only commands (`refreshConnection`, `refreshDatabase`, `refreshCollection`, `refreshDocumentList`, `refreshSchema`) have been moved to `DEEP_LINK_DISALLOWED_COMMANDS`.
- An explicit `DEEP_LINK_PARAM_SCHEMA` was introduced that maps each allowed command to either `null` (no parameters accepted) or a typed parameter definition (`mdb.connectWithURI`, `mdb.searchForDocuments`, `mdb.exportToLanguage`, `mdb.changeDriverSyntaxForExportToLanguage`, `mdb.codeLens.showMoreDocumentsClicked`).
- The `showWarningMessage` confirmation dialog was added when the `connectionString` is pre-supplied from outside of VSCode calls. The dialog displays only the `hosts` extracted via `ConnectionString.hosts`, never the full URI (which may contain credentials). The existing `showInputBox` flow for manual entry is unchanged.
- The application scope was added to sensitive settings (`mdb.confirmRunAll`, `mdb.confirmRunCopilotCode`, `mdb.confirmDeleteDocument`, `mdb.persistOIDCTokens`, `mdb.showOIDCDeviceAuthFlow`, `mdb.mcp.readOnly`). Users will need to move those to global settings.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


package.json
```diff
@@ -1227,11 +1227,13 @@
         "mdb.confirmRunAll": {
           "type": "boolean",
           "default": true,
+          "scope": "application",
```
lerouxb, line 1230: Nice. I wasn't even aware this was a thing.

src/mdbExtensionController.ts
```diff
@@ -157,6 +159,91 @@ export const DEEP_LINK_DISALLOWED_COMMANDS = [
...
+    if (!(key in schema)) {
+      return `Unknown parameter '${key}' for command '${command}'.`;
+    }
+  }
+
+  for (const [key, def] of Object.entries(schema)) {
+    const value = params[key];
+    if (value === undefined || value === null) continue;
+
+    if (Array.isArray(value)) {
```
lerouxb: Why only arrays? Is it possible to get here with an array?

I see below you check string, boolean and number, shouldn't the final fallback error rather than just allow everything else?

src/mdbExtensionController.ts
```diff
@@ -157,6 +159,91 @@ export const DEEP_LINK_DISALLOWED_COMMANDS = [
...
+    includeDriverSyntax: { type: 'boolean' },
+  },
+  [ExtensionCommand.mdbCodelensShowMoreDocuments]: {
+    operationId: { type: 'string' },
+    connectionId: { type: 'string' },
+    namespace: { type: 'string' },
+  },
+};
+
+function validateDeepLinkParams(
```
lerouxb, line 195: This is probably fine, but since we've been using `zod` everywhere recently (would be surprised if it isn't already a transitive dep of this extension) I'm wondering if we shouldn't just use that here too.

src/connectionController.ts
```diff
@@ -271,33 +271,49 @@ export default class ConnectionController {
...
-
-            try {
-              // eslint-disable-next-line no-new
-              new ConnectionString(uri);
-            } catch (error) {
-              return formatError(error).message;
-            }
-
-            return null;
+      if (connectionString) {
```
lerouxb, line 274: I'm actually surprised we didn't already do this.. 😬

src/templates/playgroundCloneDocumentTemplate.ts
```diff
@@ -1,12 +1,16 @@
...
+import { createTemplate } from './templateHelpers';
+
+const playgroundCloneDocumentTemplate = createTemplate(
+  (databaseName, collectionName) => `/* global use, db */
 // MongoDB Playground
 // Use Ctrl+Space inside a snippet or a string literal to trigger completions.
 
 // The current database to use.
-use('CURRENT_DATABASE');
+use(${databaseName});
```
lerouxb, line 9: Cool. Much nicer and safer pattern than find/replace.

src/mdbExtensionController.ts
alenakhineika: Nice catch! The array check is a left over, i will remove this. Was trying to catch cases like `vscode://mongodb.mongodb-vscode/searchForDocuments?databaseName=a&databaseName=b`, but those are covered with types validations below. And will add a fallback, thx!

src/mdbExtensionController.ts
alenakhineika: You are actually right, we do already have zod as a transitive dep, using it here makes a lot of sense. Let me update.

lerouxb (approved):

lerouxb (approved):

lerouxb (approved):

PR #1313, opened 2026-06-09
chore(telemetry): track view format VSCODE-783

Anemy:
VSCODE-783

It will be nice to see who is using this. I'm thinking likely nobody. Maybe we remove the setting altogether.


src/test/suite/telemetry/telemetryService.test.ts
```diff
@@ -230,6 +235,7 @@ suite('Telemetry Controller Test Suite', function () {
           properties: {
             source: 'treeview',
             success: true,
+            view_format: getDocumentViewAndEditFormat(),
```
alenakhineika: This test would be always false positive, because we test against the same value in the setting. We should test against the exact string we expect here, e.g. `'shell'`.

src/telemetry/telemetryEvents.ts
```diff
@@ -220,10 +221,17 @@ export class DocumentUpdatedTelemetryEvent implements TelemetryEventBase {
 
     /** Whether the operation was successful */
     success: boolean;
+
+    /** Whether the user edited the document in shell format or ejson */
```
alenakhineika: The `edited` here is a bit misleading, because the event fires on save.

src/editors/mongoDBDocumentService.ts
```diff
@@ -100,10 +107,14 @@ export default class MongoDBDocumentService {
         },
       );
       this._telemetryService.track(
-        new DocumentUpdatedTelemetryEvent(source, true),
+        new DocumentUpdatedTelemetryEvent(
+          source,
+          true,
+          getDocumentViewAndEditFormat(),
```
alenakhineika:
```suggestion
          format,
```
We should bypass the `format` here and in `_saveDocumentFailed`, so if a user opens a doc in `shell`, flips the setting to `ejson`, then saves, the extension parses their text as `shell`, but telemetry will report `view_format: 'ejson'`.

alenakhineika (changes requested):

src/editors/mongoDBDocumentService.ts
Anemy: done in https://github.com/mongodb-js/vscode/pull/1313/commits/a35f1bc09aec3ffd46037f9c195eef736e9b7328

alenakhineika (approved):

PR #1311, opened 2026-06-02
feat(data-browser): add setting to view documents in native VSCode editors VSCODE-776

Anemy:
VSCODE-776 
Issue https://github.com/mongodb-js/vscode/issues/1296 

Adds a new setting for bypassing the webview for viewing documents.

PR #1330, opened 2026-08-05
feat(data-browsing): add option to use classic editor-based view VSCODE-776

### LINKED ISSUE #1296: Feature Request: Allow disabling new data browser
## Detailed Description

It would be nice if there was a control allowing users to enable or disable the new data browser UI to view the data as plain JSON like before. Currently the only option to bypass the new UI is to use `.toArray()`.

## Context

I personally prefer it over the new data browser UI, since I don't have to collapse every single property manually to view the full document content. It's also easier to copy things using only keyboard movements controls (it's also way faster, but I guess this part will get better over time).

## Possible Implementation

Something like this would do the trick I guess. I tried commenting out all code but the `_openInResultPane` call, and I get the result I want:
``` typescript
async _openResult(result: PlaygroundRunResult): Promise<void> {
   const shouldAlwaysUseResultPane = vscode.workspace
      .getConfiguration('mdb')
      .get('alwaysUseResultPane');

    if (result.constructionOptions && !shouldAlwaysUseResultPane) {
      const { method } = result.constructionOptions.options;
      if (method === 'find' || method === 'aggregate') {
        // open find or aggregate cursor results in the data browser
        // Note: We leave out aggregateDb and runCursorCommand for now because
        // those don't have an associated collectionName and that's a bit
        // removed from how data browser currently works.
        await vscode.commands.executeCommand(
          ExtensionCommand.mdbOpenDataBrowserFromPlayground,
          { result },
        );
        return;
      }
    }

    // as a fallback, show results that aren't find or aggregate cursors in the result pane
    await this._openInResultPane(result);
  }
```

anionDev:
Add a new opt-in setting, `mdb.useClassicDataBrowsingExperience` (default: false), that lets users switch back to the classic, editor-based experience for viewing collection documents and playground find/aggregate results, instead of the webview-based Data Browser.

- `mdb.viewCollectionDocuments` now routes to the existing editor-based collection view (`EditorsController.onViewCollectionDocuments`) when the setting is enabled, instead of opening the Data Browser webview.
- `PlaygroundController._openResult` keeps find/aggregate cursor results in the classic result pane when the setting is enabled, instead of routing them to the Data Browser.
- As desired in the Jira-ticket: Track a telemetry event (`DataBrowsingExperienceSettingTelemetryEvent`) both on activation and on setting change, so you can measure adoption and decide whether the default should change.

Closes #1296
VSCODE-776

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


alenakhineika:
Hi @anionDev, thank you for taking the time to put this together! Unfortunately this overlaps with #1311, which our team already has open for the same change, therefore I'm going to close this PR as a duplicate.

That said, feel free to browse the open issues, and thanks again for your interest in contributing!

